PR #302, opened 2019-05-15
translation

edentw:
Hi,
This is a traditional Chinese Translation, I hope you can put it in.
for zh_TW


csagan5:
This is not the correct way to submit translation patches, see https://www.chromium.org/developers/design-documents/ui-localization

PR #366, opened 2019-09-01
Remove irrelevant patches

Madis0:
This PR removes patches that are irrelevant to mobile Chromium.

- Remove-EV-certificates.patch - mobile Chromium does not treat EV differently in the UI, [desktop will not soon, either](https://groups.google.com/a/chromium.org/forum/#!msg/security-dev/h1bTcoTpfeI/jUTk1z7VAAAJ)
- extensions-always-show-component-extensions-in-the-ext-list.patch - mobile Chromium does not implement extension support at the moment, therefore `chrome://extensions` is invalid anyway


csagan5:
The extensions patch can be dropped since there is no extensions support on Android; the EV verification however has other consequences than the visual ones: `IsCaBrowserForumEvOid()` is used to apply some certificate validation exceptions, which I do not wish to follow upstream about (see https://bugs.chromium.org/p/chromium/issues/detail?id=705285).

Madis0:
From what I understand, the bug is not about whether such certificate is valid, but rather whether it is valid to consider as an EV certificate. If the browser treats OV/DV/EV certs the same (as in, checks their overall validity properly), what is the issue here?

csagan5:
The browser does not treat the certificates the same; check the usage of the function I mentioned in the codebase.

csagan5:
The `Delete extensions-always-show-component-extensions-in-the-ext-list.patch` will be removed in next release.

Issue #388, opened 2019-10-19
Chromium crashes Clicking on setting .


Techguyprivate:
**Bromite version**
No, vanilla chrmium 
Version: 77.3865.104
Arch: `arm64
Android version: 9 pie
Device model: two devices. samsung galaxy m30s, realme 2 pro

**Is this bug about the SystemWebView?**
No

No support for SystemWebView installations is provided.

**Is the bug reproducible with latest version?** yes

Do not report bugs which are not reproducible with latest version

**Can the bug be reproduced with corresponding Chromium version?** yes

Please pick the same version of Chromium as Bromite from here: https://github.com/bromite/chromium/releases
If the bug is reproducible then it might be a configuration issue or an upstream bug. Upstream bugs can be reported on the [Chromium issue tracker](https://bugs.chromium.org/p/chromium/issues/list) and do not forget to read [Chromium project bug reporting guidelines](https://www.chromium.org/for-testers/bug-reporting-guidelines) first.

**Is the bug a crash?** yes.
![Uploading Record_2019-10-19-15-33-31_e2990213a583746c04dbe25f92ec7f6e.gif…]()

If yes then individuate and post the logcat dump (remove privacy sensitive information, if any), otherwise remove this section.

**Describe the bug**
Vanilla chromium crashed when clicking setting.
Write here a clear and concise description of the bug.
Install the latest version of vanilla chromium from bromite repo. 

Click on setting.
**To Reproduce**

**Expected behavior**

**Screenshots**

If applicable, add screenshots to help explain your problem. Otherwise remove this section.

Techguyprivate:
![Record_2019-10-19-15-33-31_e2990213a583746c04dbe25f92ec7f6e](https://user-images.githubusercontent.com/43615747/67143192-6aa13500-f286-11e9-99b3-e2bdee583f64.gif)

kowith337:
I want to report the new Bromite update are have the same issue, however, only when you want to change homepage.

csagan5:
@Techguyprivate please test with latest version of Chromium (`77.3865.104`), which is now available

csagan5:
> I want to report the same issue on the Bromite that likely have the same issue, however, only when you want to change homepage.

@kowith337 that is a separate issue

Techguyprivate:
I just want to clarify what I did else .


1. Clearing browser cache, data, and all app data.

 2. Exiting the browser completely, from recent apps and forcibly stopping it from app management 
      in android setting .

[...]

Issue #387, opened 2019-10-18
Unable to enable page predictions


mthomos:
**Bromite version**

Version: '78.0.3904.63`
Arch: `arm64`
Android version: `9.0`
Device model: 'Xperia XZ1'

**Is this bug about the SystemWebView?**
No

**Is the bug reproducible with latest version?**

Yes

**Can the bug be reproduced with corresponding Chromium version?**
Not provided yet

**Is the bug a crash?**
bug

**Describe the bug**

It is impossible to enable page prediction in privacy settings

**To Reproduce**

1. Go to Privacy settings
2. Tap to enable Use page predictions
3. Tap to enable Preload pages for faster browsing
4. Press back button to return Settings Menu
5. Go to Privacy settings again

**Expected behavior**

Use page predictions and Preload pages for faster browsing are disabled and not enabled

mthomos:
I think that the changes about PREF_NETWORK_PREDICTIONS in this patch should be reverted https://github.com/bromite/bromite/blob/master/build/patches/Disable-safe-browsing.patch

csagan5:
@mthomos yes, it's a mistake - it will be fixed in next release.

nikolowry:
Can confirm that https://github.com/bromite/bromite/commit/28e950d9beaafe30f683959d1789cf9915ac8d38 resolves this issue

Issue #375, opened 2019-09-24
Cannot enable search suggestions/tap to search


mthomos:
**Bromite version**

Version: `77.0.3865.95`
Arch: `arm64`
Android version: `9.0`
Device model: 'Sony Xperia XZ1'

**Is this bug about the SystemWebView?**
No

**Is the bug reproducible with latest version?** 'Yes'

**Can the bug be reproduced with corresponding Chromium version?**
No

**Is the bug a crash?**

No

**Describe the bug**

I don't know if it's intented to work or not. Using latest version and all search engines, suggestions don't work.

**To Reproduce**

1. Tap to omnibox
2. Type text

**Expected behavior**

See suggestions for search

csagan5:
The problem here is that the option is disabled by default and the option is no more visible in settings; upstream moved it.

I will have to look into this to possibly restore the menu options also for non-logged-in users.

csagan5:
The relevant menu entries have been removed in upstream commit https://github.com/chromium/chromium/commit/c672a0683e89eba3554e2562156b3c8532136f21

csagan5:
In `77.0.3865.104` is now possible to enable search suggestions, however "Tap to search" seems to have no effect.

nikolowry:
@csagan5 can confirm same results on `78.0.3904.63` -- search suggestions work and enabling "Tap to search" has no effect

csagan5:
I have identified them missing code bit; the contextual search toggle will be fixed in next release.

csagan5:
@mthomos is this problem still present in latest version? I am afraid that even with the code changes the preference is not working properly.

icejr:
It seems like this issues is also fixed (at least for me) on `78.0.3904.119`

mthomos:
Tap to search doesn't work.

csagan5:
@icejr can you please check that it is not in the tap menu when it's disabled in settings, and viceversa?

icejr:
@csagan5 I'm having a strange issue... no matter what option I set (enable/disable) in settings for tap to search, I'm still able to see a 'Web Search' option when selecting a text... but web search/tap to search is working just fine

nikolowry:
@csagan5 as of the last two versions, these options have disappeared from my settings' Privacy menu (clean builds). Assumed they were dropped, but now after seeing these latest comments I'd thought it might be worth mentioning

csagan5:
[...]

PR #390, opened 2019-10-20
[Build] [Patches] [Disable Safe Browsing] Hotfix for failing patch

nikolowry:
Applying `Disable-safe-browsing.patch` was failing due to `chrome/android/java/src/org/chromium/chrome/browser/preferences/privacy/PrivacyPreferences.java:36` and `chrome/android/java/strings/android_chrome_strings.grd:421`.

NOTE: When `Disable-safe-browsing.patch` fails, these additional patches fail as well(dependent patches):
- `Add-custom-tab-intents-privacy-option.patch`
- `Add-option-to-not-persist-tabs-across-sessions.patch`
- `Add-a-proxy-configuration-page.patch`

### DIFF
--- a/build/patches/Disable-safe-browsing.patch
+++ b/build/patches/Disable-safe-browsing.patch
@@ -75,7 +75,7 @@ diff --git a/chrome/android/java/res/xml/privacy_preferences.xml b/chrome/androi
 +++ b/chrome/android/java/res/xml/privacy_preferences.xml
 @@ -7,24 +7,11 @@
      xmlns:app="http://schemas.android.com/apk/res-auto">
- 
+
      <org.chromium.chrome.browser.preferences.ChromeBaseCheckBoxPreference
 -        android:key="navigation_error"
 -        android:title="@string/navigation_error_title"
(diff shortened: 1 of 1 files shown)


csagan5:
@nikolowry thanks for the PR, I wonder why the patch failed though - perhaps needs to be applied with some fuzz factor?

I had already updated the patches because of #387 and #388 but I had not yet released them because I was planning to do that on next release.

I will now push the fixed patches that I already have; please see if there is anything from this PR that could still be submitted?

nikolowry:
#387 is able to be replicated with the patch in this pull request (I never enable it, so it would have gone unnoticed by me) and #388 never effected me (but I only built up to 77.0.3865.95 during the last major releases).

I can build with your patch fixes today and report back on #387

csagan5:
> #387 is able to be replicated with the patch in this pull request

Alright, thanks, I will give a look for next build - it is perhaps another patch mishap.

> I wonder why the patch failed though

This mystery remains though.

csagan5:
@nikolowry you mean that #387 (and not #388) is still present? In this case can you test if #375 is really solved?

nikolowry:
#387 was present with the patch provided by me in this pull request -- just finished building with https://github.com/bromite/bromite/commit/28e950d9beaafe30f683959d1789cf9915ac8d38 and can confirm it is resolved with that commit

nikolowry:
> I wonder why the patch failed though

I'm no diff/patch algorithm expert, but I'm assuming the changes in `PrivacyPreferences.java` wasn't able to be applied because of `Revert-Remove-pre-unified-consent-code-in-sync-and-privacy-directory.patch`. If it happens again, it might be worth the effort to try and consolidate those two patches.

Issue #546, opened 2020-05-05
I18n support for new menu


zhmars:
### Is your feature request related to privacy?

No.

### Is there a patch available for this feature somewhere?

No.

### Describe the solution you would like

Hi, I notice some menu introduced by patches is not localized. Could you please consider adding i18n support for that? I would like to translate bromite into my language.

Please feel free to close this since the feature is not related to privacy.

Screenshot: https://imgur.com/a/PVi5bd2

### Describe alternatives you have considered

Sorry, I dont know.

Edited: remove some screenshots and fix typo.

csagan5:
I understand the problem but I would need somebody to submit a patch for this as I am not able to correctly add the translations.

csagan5:
A partial fix has been released with `83.0.4103.46`.

csagan5:
@zhmars please check if your latest patch looks good on this pre-release: https://github.com/bromite/bromite/releases/tag/83.0.4103.53

Otherwise feel free to open a new PR.

zhmars:
Thanks. Everything looks good except the `About Bromite` menu in setting page. Do you know anything about this?

Screenshot: https://imgur.com/a/ko0hE3y

csagan5:
Fixed in `83.0.4103.53`.

csagan5:
@zhmars that part is automatically replaced through a script, for now I will keep it like this

zhmars:
Ok, thanks a lot for your explanation.

PR #665, opened 2020-08-01
Update zh_CN translations

zhmars:
- Add translation for `View source` introduced by v84.0.4147.90.
- Remove `About Bromite` translation from `chromium_strings_zh-CN.xtb` since it never works.


csagan5:
Any idea why that translation does not work?

csagan5:
Thanks for the update!

zhmars:
Thanks. For why that translation not work, see https://github.com/bromite/bromite/issues/546#issuecomment-631271211.

zhmars:
Hi, @csagan5.

Is the patch actually applied in latest release 84.0.4147.119? I didn't see it after fresh installation.

And I also noticed the `Update-i18n-zh_CN-support.patch` in repo have been restored.

csagan5:
@zhmars must have happened by mistake, will correct it on next release; thanks

Issue #142, opened 2018-09-26
Cookies not being cleared on quit


tcz305:
**Bromite version**

Version: 69.0.3497.109
Arch: arm64
Android version: Oreo

**Is the bug reproducible with latest version?**
Yes

**Can the bug be reproduced with corresponding [Chromium](https://github.com/bromite/chromium/releases) version?**
No

**Is the bug a crash?**
No

**Describe the bug**
With the setting "keep cookies only until you quit your browser" checked, and logging into a site eg amazon etc, and later exiting the browser by clicking exit, login remains through restarts **unless site storage is cleared**.
All sites also remember all info between sessions like store location entered, site browsing history, etc.
_I'm not sure if this is by design that all info is being stored in site storage but I thought clearing cookies would clear all the info_.

**To Reproduce**
1. Go to any site
2. Click on login or enter any other info
3. Exit browser
4. On restart appears all cookies (seem to) remain.

**Expected behavior**
Expected all cookies to be cleared including logins and entered info and site browsing history and location etc.

Techguyprivate:
Where is the setting for session cookies? Can you post a screenshot?

csagan5:
@tcz305 a session is closed when the app exits. You need to forcefully stop it to be sure the session ended.

csagan5:
`69.0.3497.91` also affected.

csagan5:
Is this problem still reproducible?

Adam-Stambaugh:
I have the same issue on my end. Cookies aren't clearing when I quit out of the browser with the option enabled as shown below.
![Screenshot_20190920-133357_Bromite](https://user-images.githubusercontent.com/49969486/65350660-d15e0080-dbab-11e9-97c2-7d2eeb72cfaa.png)
I'm running LineageOS 16.0, WebView (AOSP) 76.0.3809.111, and Bromite 76.0.3865.81.

sviluppoSoftware:
i confirm this bug it's still present in 78.0.3904.130 version too

kef71:
I can confirm this bug it's still present in v79.0.3945.123

uazo:
It can be done at startup, after `ChromeActivity.finishNativeInitialization()` adding for example

```
BrowsingDataBridge.getInstance().clearBrowsingData(
        new BrowsingDataBridge.OnClearBrowsingDataListener() {
	   @Override
            public void onBrowsingDataCleared() {
		Context context = ContextUtils.getApplicationContext();
[...]

PR #671, opened 2020-08-05
session only cookies patch - rev 01

uazo:
as https://github.com/bromite/bromite/issues/142#issuecomment-669246681


csagan5:
Thanks for your work! See my reply on #142

build/patches/Session-only-cookies-v2.patch
```diff
@@ -0,0 +1,775 @@
...
+diff --git a/components/browser_ui/site_settings/android/java/src/org/chromium/components/browser_ui/site_settings/FourStateCookieSettingsPreference.java b/components/browser_ui/site_settings/android/java/src/org/chromium/components/browser_ui/site_settings/FourStateCookieSettingsPreference.java
+old mode 100644
+new mode 100755
+index f2353a6ba2..5eb5f7ef58
+--- a/components/browser_ui/site_settings/android/java/src/org/chromium/components/browser_ui/site_settings/FourStateCookieSettingsPreference.java
++++ b/components/browser_ui/site_settings/android/java/src/org/chromium/components/browser_ui/site_settings/FourStateCookieSettingsPreference.java
+@@ -31,6 +31,7 @@ public class FourStateCookieSettingsPreference
+         BLOCK_THIRD_PARTY_INCOGNITO,
+         BLOCK_THIRD_PARTY,
+         BLOCK
```
csagan5, line 196:
```suggestion
         BLOCK,
```

build/patches/Session-only-cookies-v2.patch
```diff
@@ -0,0 +1,775 @@
...
++        ,SESSION_ONLY
```
csagan5, line 197:
```suggestion
+        SESSION_ONLY
```

build/patches/Session-only-cookies-v2.patch
```diff
@@ -0,0 +1,775 @@
...
+     }
+ 
+     /**
+@@ -48,6 +49,9 @@ public class FourStateCookieSettingsPreference
+         public boolean cookiesContentSettingEnforced;
+         //  Whether third-party blocking is enforced.
+         public boolean thirdPartyBlockingEnforced;
++
```
csagan5, line 205:
```suggestion
```

csagan5:
@uazo this is the build with this patch: https://fdroid.bromite.org/fdroid/bromite/brm_sess_only.apk

I have also updated your branch with the adjusted version I built.

I do not use any particular tool to manage the patches, just `git am`, `git cherry-pick`, `git rebase`; the best would be to learn and use [quilt](https://wiki.debian.org/UsingQuilt).

csagan5:
@uazo does session-only also block third-party cookies? Do you think we could make it work like that? If yes I would also update the description and move it before `Block all cookies`

uazo:
> does session-only also block third-party cookies? 

Yes, chromium code block third-party cookies if a cookies is session-only.
You can see here https://github.com/bromite/bromite/pull/671/files#diff-2fad0c6caf1f9396df1464bde53f7a30R278

>  If yes I would also update the description and move it before Block all cookies

please tell me what exactly to write, my english is pretty poor.
then I edit them in the code and run grit (id for the translations)

csagan5:
@uazo I have updated the description in this patch and moved it accordingly, thanks for confirming.

PR #675, opened 2020-08-07
Add history on incognito tabs

uazo:


csagan5:
I am building with this patch to see the effect.

csagan5:
This is the build with this patch: https://fdroid.bromite.org/fdroid/bromite/brm_sess_only_inc_hist.apk

csagan5:
This seems to work correctly and can be released on next version.

Issue #592, opened 2020-05-23
Why not on F droid?


rogerxed:
I couldn't find this app on F droid , why this app is not available on there??

PR #683, opened 2020-08-12
Added link to the issue on the F-Droid repository

anewuser:
People reading the FAQ to know why it isn't on app stores will be interested in following this issue.

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -43,9 +43,9 @@ No, and this is unlikely to change. Many limitations apply for submissions there
 Bromite favors user freedom in software choice: the device is yours so you get to choose which software to run on it, end of the story.
 
 ## Is Bromite on F-Droid?
-It is not on the official F-Droid repository and there are no (more) plans to submit it.
+It is not on the official F-Droid repository yet due to the resources needed to compile Chromium builds (see https://gitlab.com/fdroid/rfp/-/issues/378).
 
-You can use F-Droid client to install and receive updates via [the official Bromite F-Droid repository](https://www.bromite.org/fdroid).
+You can use the F-Droid client to install and receive updates via [the official Bromite F-Droid repository](https://www.bromite.org/fdroid).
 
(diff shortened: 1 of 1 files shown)


csagan5:
The issue to begin reading about this is #4, not that one, and what you mentioned is not the only reason why Chromium forks are not on F-Droid.

anewuser:
@csagan5 Please correct the text then to prevent people from creating new issues about this.

csagan5:
@anewuser before creating a new issue you are kindly requested to search the issue tracker; this is valid for any issue, not just about F-Droid.

anewuser:
@csagan5  I had already found it but I was trying to save people's time, since the FAQ answer isn't clear, and several issues mentioning your custom F-Droid server come up when people look it up. Also, the most recent issue about this has  been closed without any comment (#592).

csagan5:
@anewuser that issue is missing the issue template; the issue tracker cannot be used to ask questions, just for bug reports or feature requests.

The FAQ answer is clear in stating that Bromite is not on the F-Droid official repository, it does not need to justify it. That would be another question: "Why is Bromite not on the F-Droid official repository?" But that's not a question to put on FAQs, forums/discussion aggregators are better suited for that. It spawns a long, complex discussion, it has some technical aspects and we already did it on multiple issue trackers.

csagan5:
I could add a link to issue #4, but the discussion has happened also on GitLab and other places, so it would not be comprehensive. I am not adding the link because it would be incomplete without adding the other links, and then after adding all the links that give a good context it would not look like a FAQ answer anymore.

anewuser:
Thank you for the in-depth response. I personally was just looking for an issue I could subscribe to and be notified about when/if the browser is added to the main repository, but now I understand you're trying to avoid technical debates and misunderstandings.

csagan5:
I am also looking for that: an issue about adding Chromium to F-Droid. Once that happens Bromite can be easily added as well.

PR #687, opened 2020-08-14
add build tools

uazo:
therefore, I was able to create scripts to help modify a fix.

`build-from-repos.sh`
generate the repos and start the build

`export-all-patch.sh`
traverses all commits regenerating patches

so, you can edit a fix with the commands
```
$ git commit --fixup [COMMIT_SHA]
$ git rebase --autosquash -i [COMMIT_SHA]^
```
where [COMMIT_SHA] is the commit for the patch to be modified.
I personally use gitui (https://extrawurst.itch.io/gitui) to retrieve SHA and staging files.

I would like to add docker later, so that I already have a precompiled environment ready for editing


build/build-from-repos.sh
```diff
@@ -0,0 +1,70 @@
...
+gclient sync -D
+
+# remove origin for chromium
+git remote remove origin
+
+echo -e ${RED} ------- apply patchs ${NC}
+for file in $(cat ~/bromite/build/patches_list.txt) ; do 
+	echo -e ${RED} " -> Apply $file" ${NC}
+
+	REPL="0,/^---/s//FILE:"$file"\n---/"
```
uazo, line 54: @csagan5 
i tried with tags and notes before, but sadly git-rebase doesn't keep those values.
so I preferred to add a description in the commit, in the form

`FILE: <patch name>`

which is then deleted with export-all-patch.sh (https://github.com/bromite/bromite/pull/687/files#diff-0a85d7dc8b850b3e4ac402a503566661R17)

build/export-all-patch.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+VERSION=$(cat ~/bromite/build/RELEASE)
+CURRENT_RELEASE=$(git -C ~/chromium/src/ rev-parse --verify refs/tags/$VERSION)
+
+ALLPATCHS=$(git -C ~/chromium/src/ rev-list HEAD...$CURRENT_RELEASE)
+
+for patch in $ALLPATCHS; do
+
+	PATCH_FILE=$(git -C ~/chromium/src/ show -s $patch | grep FILE: | sed 's/FILE://g' | sed 's/^[ \t]*//;s/[ \t]*$//')
```
uazo, line 10: @csagan5 
here I take the file name of the patch inserted above and trim spaces

csagan5:
I have not added build tools on purpose, because I do not wish to maintain them. Back then @nikolowry started this separate project: https://github.com/nikolowry/bromite-builder

build/build-from-repos.sh
csagan5: Yeah it looks like solid work; I only released once some basic scripts: https://github.com/bromite/bromite/blob/for-fdroid/fdroid/build.sh

They are very outdated by now; then others picked up from there and did their own scripts (see other comment). I also have some scripts but I do not release them because then I would have to maintain them. Building Chromium is cumbersome but once you can build that then adding the patches should be easy.

uazo:
Perfect ok I understand, it seems very complete, I am available for explanations if you want to take some ideas @nikolowry

PR #709, opened 2020-08-23
fix bromite bugs 377 and 679

### LINKED ISSUE #377: Set proxy to "None" will not fallback to system proxy
**Bromite version**
> Version: `77.0.3865.95 (1568938391)`
> Arch: `arm64 (ChromeModernPublic)`
> Android version: `9.0`
> Device model: `XT1970-3 (kane)`

**Is this bug about the SystemWebView?**
> No

**Is the bug reproducible with latest version?**
> I'm using current release version

**Can the bug be reproduced with corresponding Chromium version?**
> No, because vanilla chromium doesn't have `Proxy` feature.

**Is the bug a crash?**
> No, this bug doesn't occur any crash attempts.

**Describe the bug**
> Bromite set proxy to `None` by default, If you didn't touch or config anything in `chrome://proxy` and have setup to use proxy at APN or Wi-Fi, Bromite will use system proxy.
> 
> However, if you've changed and revert back to `None`, it will use direct connection and it's impossible to revert, unless you've reset app data. (that's mean cleanup everything!)

**To Reproduce**
> 1. Set proxy at Wi-Fi or APN
> 2. Open Bromite, then do test to known your IP address (e.g. search `IP address` on DuckDuckGo)
> 3. Go to `chrome://proxy`, change to another value, apply change, then reset to **None**, then apply change.
> 4. Do step 2 again.

**Expected behavior**
> Bromite should revert back to system proxy, not direct connections.

**Screenshots**
> image
> image

**Notes**
> Make sure you didn't have any VPN running, either main or work profile.
>

### LINKED ISSUE #679: Could not add  proxy as list - proxy configuration
### Bromite version

Version: `84.0.41447.113`
Arch:  `arm64` 
Android version: 9  (example: `9.0`)
Device model: Onclite (example: `SM-G960UZKABST`)

Yes/No
no

Yes

I dont know

### Is the bug a crash?
No

### Describe the bug
Proxy Config: It says a "list" , so I added  a list with ";"  but it ends up with the first one

1. Open proxy configuration
2. take a list of proxys: 
   Eg: 
   PROXY 183.23.3.23:8080;
   PROXY 123.32.35.6:80
3. clicked ON "USE SINGLE PROXY LIST FOR ALL SCHEMES"
   I believe the PAC format is correct in my case on step 2 when reads with step3
4. click on apply
5. Only one proxy is saved instead of a list

Steps to reproduce the bug:
2. Click on '...'
3. Scroll down to '...'

### Expected behavior
It should be a list when add a list but its just one proxy

### Screenshots
Nil

uazo:
fixes #377 
fixes #679

### DIFF
--- a/build/patches/Add-a-proxy-configuration-page-fix-377-679.patch
+++ b/build/patches/Add-a-proxy-configuration-page-fix-377-679.patch
@@ -1,0 +1,236 @@
+From: root <root@ubuntu.cabponte>
+Date: Sun, 23 Aug 2020 20:00:18 +0000
+Subject: fix 377 and 679
+
+---
+ chrome/browser/resources/proxy_config.html |  5 ++-
+ chrome/browser/resources/proxy_config.js   | 24 +++++++----
+ chrome/browser/ui/webui/proxy_config_ui.cc | 46 +++++++++++++++-------
+ net/proxy_resolution/proxy_config.cc       | 16 +++++---
+ 4 files changed, 63 insertions(+), 28 deletions(-)
+ mode change 100644 => 100755 chrome/browser/resources/proxy_config.html
+ mode change 100644 => 100755 chrome/browser/resources/proxy_config.js
+ mode change 100644 => 100755 chrome/browser/ui/webui/proxy_config_ui.cc
+ mode change 100644 => 100755 net/proxy_resolution/proxy_config.cc
+
+diff --git a/chrome/browser/resources/proxy_config.html b/chrome/browser/resources/proxy_config.html
+old mode 100644
+new mode 100755
+index 898716564728630258b4d4c442d21da9d326f1a1..1db228c3bbfe52c35adea95a36fe8190a07d6cbb
+--- a/chrome/browser/resources/proxy_config.html
++++ b/chrome/browser/resources/proxy_config.html
+@@ -44,14 +44,15 @@
+         Reset will update the displayed configuration to match the one currently in use.
+       </div>
+       <div class="section-container">
+-          <input type="radio" id="empty" name="mode" value="empty" checked><label for="empty">None</label><br/>
++          <input type="radio" id="empty" name="mode" value="empty"><label for="empty">System Default</label><br/>
++          <input type="radio" id="direct" name="mode" value="direct"><label for="direct">Direct</label><br/>
+           <input type="radio" id="auto-detect" name="mode" value="auto-detect"><label for="auto-detect">Auto-detect (WPAD DHCP/DNS)</label><br/>
+           <input type="radio" id="use-pac-url" name="mode" value="use-pac-url"><label for="use-pac-url">Use PAC URL: <input id='pac-url' value="" size="40" /></label>
+           <p><input type="checkbox" id="pac-mandatory" name="pac-mandatory"><label for="pac-mandatory">Do not allow fallback to direct connection in case PAC script fails</label></p>
+           <input type="radio" id="use-single-list" name="mode" value="use-single-list"><label for="use-single-list">Use a single proxy list for all schemes (<a target="_blank" href="https://developer.mozilla.org/en-US/docs/Web/HTTP/Proxy_servers_and_tunneling/Proxy_Auto-Configuration_(PAC)_file#Description">PAC format</a>):
+               <textarea id="single-proxies"></textarea>
+             </label><br/>
+-          <input type="radio" id="use-list-per-scheme" name="mode" value="use-list-per-scheme"><label for="use-list-per-scheme">Use a proxy list per scheme:</label><br/>
++          <input type="radio" id="use-list-per-scheme" name="mode" value="use-list-per-scheme"><label for="use-list-per-scheme">Use a proxy list per scheme (use ; as separator):</label><br/>
+               <label for="use-list-per-scheme">HTTP (<a target="_blank" href="https://developer.mozilla.org/en-US/docs/Web/HTTP/Proxy_servers_and_tunneling/Proxy_Auto-Configuration_(PAC)_file#Description">PAC format</a>):<br/>
+               <textarea id="http-proxies"></textarea></label><br/>
+               <label for="use-list-per-scheme">HTTPS (<a target="_blank" href="https://developer.mozilla.org/en-US/docs/Web/HTTP/Proxy_servers_and_tunneling/Proxy_Auto-Configuration_(PAC)_file#Description">PAC format</a>):<br/>
+diff --git a/chrome/browser/resources/proxy_config.js b/chrome/browser/resources/proxy_config.js
+old mode 100644
+new mode 100755
+index 804580cd1bbe0ce7c5c2f70ef59e91d28f1d2e09..a55bd9c1ae666659705d658eb37a93b48e6f56d7
+--- a/chrome/browser/resources/proxy_config.js
++++ b/chrome/browser/resources/proxy_config.js
+@@ -40,6 +40,7 @@ var ProxyConfigView = (function() {
+   var kIdClearButton = 'clear';
+ 
+   var kIdModeEmpty = 'empty';
++  var kIdModeDirect = 'direct';
+   var kIdModeAutoDetect = 'auto-detect';
+   var kIdModeUsePacURL = 'use-pac-url';
+ 
+@@ -125,7 +126,7 @@ var ProxyConfigView = (function() {
+           "rules": {
+             "bypass_rules": "",
+             "reverse_bypass": false,
+-            "type": "empty"
++            "type": "none"
+           }
+         };
+     },
+@@ -138,7 +139,14 @@ var ProxyConfigView = (function() {
+         return {
+           "auto_detect": false,
+           "rules": {
+-            "type": "empty"
++            "type": "none"
++          }
++        };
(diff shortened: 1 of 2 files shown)


uazo:
for #679 maybe i've found a upstream bug here https://github.com/bromite/bromite/pull/709/files#diff-1256fb39a688a226d836478e39ba4ab3R1169

csagan5:
There are too many whitespace changes here, can you redo the patch without rewriting all the files? It makes it near to impossible to see what has changed.

Also: the mode of files was changed, but that is not necessary.

build/patches/Add-a-proxy-configuration-page-fix-377-679.patch
```diff
@@ -0,0 +1,236 @@
...
+         type = Type::PROXY_LIST;
+-        return;
++        // bromite #679
++        continue; // was: return; maybe an upstream bug
+       }
+ 
+       // Trim whitespace off the url scheme.
+@@ -148,11 +149,16 @@ std::string ProxyConfig::ProxyRules::ToString() const {
+ 
+   // special case: a single proxy server specified
```
csagan5, line 218: It should say "a single proxy server list specified"

build/patches/Add-a-proxy-configuration-page-fix-377-679.patch
```diff
@@ -0,0 +1,236 @@
...
+           <input type="radio" id="use-pac-url" name="mode" value="use-pac-url"><label for="use-pac-url">Use PAC URL: <input id='pac-url' value="" size="40" /></label>
+           <p><input type="checkbox" id="pac-mandatory" name="pac-mandatory"><label for="pac-mandatory">Do not allow fallback to direct connection in case PAC script fails</label></p>
+           <input type="radio" id="use-single-list" name="mode" value="use-single-list"><label for="use-single-list">Use a single proxy list for all schemes (<a target="_blank" href="https://developer.mozilla.org/en-US/docs/Web/HTTP/Proxy_servers_and_tunneling/Proxy_Auto-Configuration_(PAC)_file#Description">PAC format</a>):
+               <textarea id="single-proxies"></textarea>
+             </label><br/>
+-          <input type="radio" id="use-list-per-scheme" name="mode" value="use-list-per-scheme"><label for="use-list-per-scheme">Use a proxy list per scheme:</label><br/>
++          <input type="radio" id="use-list-per-scheme" name="mode" value="use-list-per-scheme"><label for="use-list-per-scheme">Use a proxy list per scheme (use ; as separator):</label><br/>
```
csagan5, line 36: Also newlines work? I would omit this, otherwise it would be needed on every box.

build/patches/Add-a-proxy-configuration-page-fix-377-679.patch
```diff
@@ -0,0 +1,236 @@
...
+old mode 100644
+new mode 100755
+index 133ff6c46cfcb53938671c6ec4a22b663d00bd29..82cda2455b15a550f805d5620613185b075c5f1d
+--- a/net/proxy_resolution/proxy_config.cc
++++ b/net/proxy_resolution/proxy_config.cc
+@@ -110,7 +110,8 @@ void ProxyConfig::ProxyRules::ParseFromString(const std::string& proxy_rules) {
+                                    &single_proxies,
+                                    ProxyServer::SCHEME_HTTP);
+         type = Type::PROXY_LIST;
+-        return;
```
csagan5, line 210: I think this was fine, the bug was fixed below in the other change for the single proxy list.

build/patches/Add-a-proxy-configuration-page-fix-377-679.patch
```diff
@@ -0,0 +1,236 @@
...
++        continue; // was: return; maybe an upstream bug
+       }
+ 
+       // Trim whitespace off the url scheme.
+@@ -148,11 +149,16 @@ std::string ProxyConfig::ProxyRules::ToString() const {
+ 
+   // special case: a single proxy server specified
+   if (type == Type::PROXY_LIST) {
+-    if (single_proxies.size() == 1) {
+-      return single_proxies.Get().ToURI();
```
csagan5, line 221: This is where #679 was.

csagan5 (changes requested):

csagan5:
Nice work! It's much clearer now, thanks.
I left a review.

build/patches/Add-a-proxy-configuration-page-fix-377-679.patch
uazo: sorry I do not understand

build/patches/Add-a-proxy-configuration-page-fix-377-679.patch
uazo: yes, this value is what is passed to the prefs.
above is where it parses from prefs string

build/patches/Add-a-proxy-configuration-page-fix-377-679.patch
uazo: I think it's needed. Is one of the things I would ask myself right away, what do I delimit it from? what kinda of programmers couldn't write it? now I have to go to the site to find out if it's a bug or what! :)

build/patches/Add-a-proxy-configuration-page-fix-377-679.patch
csagan5: The bug is already fixed with the other change. I think there's no upstream bug, see the tests here: https://source.chromium.org/chromium/chromium/src/+/master:net/proxy_resolution/proxy_config_unittest.cc

build/patches/Add-a-proxy-configuration-page-fix-377-679.patch
csagan5: You still need to visit the documentation to see how the PAC format works.


If you want to add it, shouldn't the label for the other boxes also mention it?

uazo: "without directions, always straight."

> If you want to add it, shouldn't the label for the other boxes also mention it?
as you like, no important

build/patches/Add-a-proxy-configuration-page-fix-377-679.patch
uazo: I don't see "myproxy:80 **; (semicolon)** https://myotherproxy"

so we save and upstream demands it

csagan5: I will merge it and test myself.

csagan5 (approved):

csagan5:
@uazo the second patch file conflicts with the first and seems a duplicate.

uazo:
`build/patches/Add-a-proxy-configuration-page.patch` have fix, is cumulative

`build/patches/Add-a-proxy-configuration-page-fix-377-679.patch` need master `build/patches/Add-a-proxy-configuration-page.patch`

which do you prefer? check three times bromite_patch_list

csagan5:
@uazo what do you mean with "check three times bromite_patch_list"?

`Add-a-proxy-configuration-page-fix-377-679.patch` does not apply after your version of `Add-a-proxy-configuration-page.patch`, it seems to be a reverse of your `Add-a-proxy-configuration-page.patch`.

Are they supposed to be identical?

uazo:
if you use `Add-a-proxy-configuration-page-fix-377-679.patch` you need to add to https://github.com/bromite/bromite/blob/master/build/bromite_patches_list.txt

if you use 'Add-a-proxy-configuration-page.patch' you don't need any changes to bromite_patches_list.txt, it's a diff from upstream diff, replaces the current one

csagan5:
Ok, but in future please make the PR so that after merged it is consistent and has only 1 version of the patch; I will fix it in `master`.

build/patches/Add-a-proxy-configuration-page-fix-377-679.patch
uazo: you can see with

`cat /data/user/0/xxxxxxx/app_chrome/Default/Preferences | grep proxy`

where xxxxxxx is your local app, for my build are is org.chromium.chrome.stable

build/patches/Add-a-proxy-configuration-page-fix-377-679.patch
csagan5: I don't understand the quote; I removed it from the version that will be used.

build/patches/Add-a-proxy-configuration-page-fix-377-679.patch
csagan5: I verified that it is necessary; you could send a patch upstream about this.

build/patches/Add-a-proxy-configuration-page-fix-377-679.patch
uazo: > I removed it from the version that will be used.

ok

build/patches/Add-a-proxy-configuration-page-fix-377-679.patch
uazo: do you?

csagan5: Done.

Issue #668, opened 2020-08-04
[Feature Request] Security Enhancements


BL4CKH47H4CK3R:
### Is your feature request related to privacy?

Yes

### Is there a patch available for this feature somewhere?

Yes. An extension already has some great privacy features. Check it out
https://github.com/jake-cryptic/AbsoluteDoubleTrace

### Describe the solution you would like

 - Current Timezone Spoofing (GMT +0) (Otherwise 
   there is no meaning of using Proxy/VPN)
 - WebGL & Device Model Fingerprinting Protection 
   (Device Model, Vendor remains unmasked)
   https://browserleaks.com/webgl
 - JS Crypto Currency Mining Domain Blocking
 - WebRTC Device Enumeration Protection
 - Screen Resolution Spoofing
 - Battery API Spoofing
 - Network Information API Spoofing
 - Browser Plugin Fingerprinting Protection
 - Hardware Fingerprinting Protection
 - Beacon/'Ping' Request Blocking
 - Blocks Malicious Top Level Domains
 - Hyperlink Auditing Prevention
 - HTTP Referrer Headers Controls
 - Google Header Tracking Controls
 - E-Tag Tracking Mitigation
 - Removal of URL Tracking Parameters

### Describe alternatives you have considered

As extension support is not available so I think these features can be added to enhance privacy

csagan5:
Feel free to submit patches.

Issue #131, opened 2018-09-17
WebGL fingerprinting deception


Eloston:
**Is your feature request related to privacy?**
Only privacy-related feature requests are considered.

Yes

**Is there a patch available for this feature somewhere?**

Not that I'm aware of.

**Describe the solution you'd like**

Do you know if it's practical to implement a WebGL fingerprinting deception mechanism similar to the one for Canvas image data?

Specifically, for WebGL image hashes as implemented in https://browserleaks.com/webgl

**Describe alternatives you've considered**

N/A

csagan5:
@Eloston I have thought of this in the past, but I quickly dismissed the possibility because of the possible vectorial nature of WebGL renderings. You can have pretty advanced color shifts (and certainly not limited to what we see in that image hash there) and texture usage, so a single pixel change might cause major visual glitches.

However this is just speculation, I have not actually tried and right now by looking at the approach used there perhaps it is possible. If we look at https://browserleaks.com/js/webgl.js towards the end there is:

```javascript

try{
  var j=new Uint8Array(131072);
  if (a.readPixels(0,0,256,128,a.RGBA,a.UNSIGNED_BYTE,j),i=JSON.stringify(j).replace(/,?"[0-9]+":/g,""),""==i.replace(/^{[0]+}$/g,""))
```

And `a` is created as:
```javascript

a=b.getContext("webgl2",{preserveDrawingBuffer:!0})||
b.getContext("experimental-webgl2",{preserveDrawingBuffer:!0})||
b.getContext("webgl",{preserveDrawingBuffer:!0})||
b.getContext("experimental-webgl",{preserveDrawingBuffer:!0})||
b.getContext("moz-webgl",{preserveDrawingBuffer:!0})}
```

So if we get a hold of `readPixels` on all of the above (those available in Chromium), then yes we can just re-use the same subchannel color data randomisation approach on the extracted raster. The only boring/lengthy part would be to implement it for all of the modes supported (not just RGBA) as I did for the `SkImage` case; likely it is using a `SkImage` internally as well, so the code might be reused.

I will give a look to this and report back here, thanks for opening this.

Eloston:
Interesting, thanks for the preliminary research.

Here are my thoughts based on what you found:

> can we intercept the result?

[...]

PR #729, opened 2020-09-05
WebGL & Device Model Fingerprinting Protection

uazo:
as you wrote https://github.com/bromite/bromite/issues/668, I propose a possible patch to hide vendor info in WebGLDebugRendererInfo.

tell me what you think and if you prefer a switch

### DIFF
--- a/build/bromite_patches_list.txt
+++ b/build/bromite_patches_list.txt
@@ -150,3 +150,4 @@ Disable-the-DIAL-repeating-discovery.patch
 Timezone-customization.patch
 Block-all-connection-requests-with-qjz9zk-in-the-domain-name-or-with-a-trk-scheme.patch
 Automated-domain-substitution.patch
+Hack-Webgl-WebGLDebugShaders.patch
--- a/build/patches/Hack-Webgl-WebGLDebugShaders.patch
+++ b/build/patches/Hack-Webgl-WebGLDebugShaders.patch
@@ -1,4 +1,4 @@
-From: root <root@ubuntu.cabponte>
+From: uazo <uazo@users.noreply.github.com>
 Date: Sat, 5 Sep 2020 06:58:52 +0000
 Subject: hack WebGLDebugShaders extension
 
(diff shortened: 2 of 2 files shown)


csagan5:
@uazo first we should discuss what changes to make; in #668 I did not intend that any patch submitted will be merged.

There is already a patch that adds a flag to disable WebGL.

uazo:
I know it's okay if you don't merge!
let's talk about the patch.

> There is already a patch that adds a flag to disable WebGL.

why disable?

csagan5:
> why disable?

If you do not disable it then one can still detect the device based on the available features/methods; there used to be [a patch like this one from ungoogled-chromium](https://github.com/Eloston/ungoogled-chromium/blob/7ba01e6/patches/extra/ungoogled-chromium/disable-webgl-renderer-info.patch) in Bromite.

This change here you want to make would have the same problem.

uazo:
> If you do not disable it then one can still detect the device based on the available features/methods

do you know something more?

> be a patch like this one from ungoogled-chromium in Bromite.

thanks, I will check next time

csagan5:
> > If you do not disable it then one can still detect the device based on the available features/methods
> 
> do you know something more?

I would need to find the GitHub issues where this was discussed; obviously for Ungoogled-Chromium they think it's still worth disabling the text returned in these cases.

See also #131, for which I have a half-developed patch somewhere.

csagan5:
See https://github.com/Eloston/ungoogled-chromium/issues/429#issuecomment-422129242
And the following comments.

This is a Bromite patch which was removing the returned information (before a flag to disable WebGL was added): https://github.com/bromite/bromite/blob/f21bb8a1d4a865d35db82389cec27b945b925cf3/patches/BRM052_Canvas-fingerprinting-mitigations-for-image-data-and-webGL.patch

csagan5:
You can even submit this patch to ungoogled-chromium, it would help the complete the patch they still have.

uazo:
ok, I get it, more complex than I thought.

PR #731, opened 2020-09-06
Remove AQS and some other parameters from search URLs

### LINKED ISSUE #726: Consider removing AQS and other parameters in google url searches

yes

I don't know

remove "The optional assisted query stats, aka AQS, used for logging purposes."
I found in code

```
    // The optional assisted query stats, aka AQS, used for logging purposes.
    // This string contains impressions of all autocomplete matches shown
    // at the query submission time.  For privacy reasons, we require the
    // search provider to support HTTPS protocol in order to receive the AQS
    // param.
    // For more details, see http://goto.google.com/binary-clients-logging .
    std::string assisted_query_stats;
```

now the search url is composed like this (looking for"fq")
https://www.google.com/search?q=fq&oq=fg&aqs=chrome..69i57.724j0j4&sourceid=chrome-mobile&ie=UTF-8

**QUERY**
q=fq

**GOOGLE_ORIGINAL_QUERY_FOR_SUGGESTION**
oq=fg 

**GOOGLE_ASSISTED_QUERY_STATS**
aqs=chrome..69i57.724j0j4 

with this contents:
- The type of page that is currently behind displayed and how it is displayed (e.g., with search term replacement or without). It's not the webpage but the page of android app (https://source.chromium.org/chromium/chromium/src/+/master:out/android-Debug/gen/third_party/metrics_proto/omnibox_event.pb.h;l=125)
- the list of experiment stats corresponding to the latest |results_| to be logged to SearchboxStats as part of a GWS experiment, if any. Experiments are fully disabled in bromite

sourceid=chrome-mobile
ie=UTF-8

1) remove that codes from components/search_engines/prepopulated_engines.json

2) check the use and remove all specific google code from
https://source.chromium.org/chromium/chromium/src/+/master:components/search_engines/template_url.cc;l=650
and
https://source.chromium.org/chromium/chromium/src/+/master:components/search_engines/template_url.cc;l=931

uazo:
fixes https://github.com/bromite/bromite/issues/726

I left the necessary parameters for ~~autocomplete~~ suggestions and image search

### DIFF
--- a/build/patches/Remove-some-params-from-search-url.patch
+++ b/build/patches/Remove-some-params-from-search-url.patch
@@ -1,0 +1,76 @@
+From: uazo <uazo@users.noreply.github.com>
+Date: Sun, 6 Sep 2020 11:53:15 +0000
+Subject: remove aqs and other parameters from search urls
+
+---
+ components/search_engines/template_url.cc | 17 -----------------
+ 1 file changed, 17 deletions(-)
+
+diff --git a/components/search_engines/template_url.cc b/components/search_engines/template_url.cc
+--- a/components/search_engines/template_url.cc
++++ b/components/search_engines/template_url.cc
+@@ -643,13 +643,11 @@ bool TemplateURLRef::ParseParameter(size_t start,
+     if (!optional)
+       url->insert(start, kDefaultCount);
+   } else if (parameter == "google:assistedQueryStats") {
+-    replacements->push_back(Replacement(GOOGLE_ASSISTED_QUERY_STATS, start));
+   } else if (parameter == "google:baseURL") {
+     replacements->push_back(Replacement(GOOGLE_BASE_URL, start));
+   } else if (parameter == "google:baseSuggestURL") {
+     replacements->push_back(Replacement(GOOGLE_BASE_SUGGEST_URL, start));
+   } else if (parameter == "google:currentPageUrl") {
+-    replacements->push_back(Replacement(GOOGLE_CURRENT_PAGE_URL, start));
+   } else if (parameter == "google:cursorPosition") {
+     replacements->push_back(Replacement(GOOGLE_CURSOR_POSITION, start));
+   } else if (parameter == "google:imageOriginalHeight") {
+@@ -671,13 +669,8 @@ bool TemplateURLRef::ParseParameter(size_t start,
+     replacements->push_back(Replacement(TemplateURLRef::GOOGLE_IMAGE_URL,
+                                         start));
+   } else if (parameter == "google:inputType") {
+-    replacements->push_back(Replacement(TemplateURLRef::GOOGLE_INPUT_TYPE,
+-                                        start));
+   } else if (parameter == "google:omniboxFocusType") {
+-    replacements->push_back(
+-        Replacement(TemplateURLRef::GOOGLE_OMNIBOX_FOCUS_TYPE, start));
+   } else if (parameter == "google:iOSSearchLanguage") {
+-    replacements->push_back(Replacement(GOOGLE_IOS_SEARCH_LANGUAGE, start));
+   } else if (parameter == "google:contextualSearchVersion") {
+     replacements->push_back(
+         Replacement(GOOGLE_CONTEXTUAL_SEARCH_VERSION, start));
+@@ -688,26 +681,18 @@ bool TemplateURLRef::ParseParameter(size_t start,
+     replacements->push_back(Replacement(GOOGLE_ORIGINAL_QUERY_FOR_SUGGESTION,
+                                         start));
+   } else if (parameter == "google:pageClassification") {
+-    replacements->push_back(Replacement(GOOGLE_PAGE_CLASSIFICATION, start));
+   } else if (parameter == "google:pathWildcard") {
+     // Do nothing, we just want the path wildcard removed from the URL.
+   } else if (parameter == "google:prefetchQuery") {
+     replacements->push_back(Replacement(GOOGLE_PREFETCH_QUERY, start));
+   } else if (parameter == "google:RLZ") {
+-    replacements->push_back(Replacement(GOOGLE_RLZ, start));
+   } else if (parameter == "google:searchClient") {
+-    replacements->push_back(Replacement(GOOGLE_SEARCH_CLIENT, start));
+   } else if (parameter == "google:searchFieldtrialParameter") {
+-    replacements->push_back(Replacement(GOOGLE_SEARCH_FIELDTRIAL_GROUP, start));
+   } else if (parameter == "google:searchVersion") {
+-    replacements->push_back(Replacement(GOOGLE_SEARCH_VERSION, start));
+   } else if (parameter == "google:sessionToken") {
+-    replacements->push_back(Replacement(GOOGLE_SESSION_TOKEN, start));
+   } else if (parameter == "google:sourceId") {
+ #if defined(OS_ANDROID) || defined(OS_IOS)
+-    url->insert(start, "sourceid=chrome-mobile&");
+ #else
+-    url->insert(start, "sourceid=chrome&");
+ #endif
+   } else if (parameter == "google:suggestAPIKeyParameter") {
+     url->insert(start,
+@@ -719,9 +704,7 @@ bool TemplateURLRef::ParseParameter(size_t start,
+   } else if (parameter == kGoogleUnescapedSearchTermsParameter) {
+     replacements->push_back(Replacement(GOOGLE_UNESCAPED_SEARCH_TERMS, start));
+   } else if (parameter == "yandex:referralID") {
+-    replacements->push_back(Replacement(YANDEX_REFERRAL_ID, start));
+   } else if (parameter == "mailru:referralID") {
+-    replacements->push_back(Replacement(MAIL_RU_REFERRAL_ID, start));
+   } else if (parameter == "yandex:searchPath") {
+     url->insert(start, YandexSearchPathFromDeviceFormFactor());
+   } else if (parameter == "inputEncoding") {
--- a/build/patches/Remove-some-params-from-search-url.patch
+++ b/build/patches/Remove-some-params-from-search-url.patch
@@ -1,10 +1,10 @@
-From: uazo <uazo@users.noreply.github.com>
-Date: Sun, 6 Sep 2020 11:53:15 +0000
+From: root <root@ubuntu.cabponte>
+Date: Sun, 6 Sep 2020 14:50:37 +0000
 Subject: remove aqs and other parameters from search urls
 
 ---
- components/search_engines/template_url.cc | 17 -----------------
- 1 file changed, 17 deletions(-)
+ components/search_engines/template_url.cc | 19 -------------------
+ 1 file changed, 19 deletions(-)
 
 diff --git a/components/search_engines/template_url.cc b/components/search_engines/template_url.cc
 --- a/components/search_engines/template_url.cc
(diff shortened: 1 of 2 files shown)


build/patches/Remove-some-params-from-search-url.patch
```diff
@@ -0,0 +1,76 @@
...
+ 1 file changed, 17 deletions(-)
+
+diff --git a/components/search_engines/template_url.cc b/components/search_engines/template_url.cc
+--- a/components/search_engines/template_url.cc
++++ b/components/search_engines/template_url.cc
+@@ -643,13 +643,11 @@ bool TemplateURLRef::ParseParameter(size_t start,
+     if (!optional)
+       url->insert(start, kDefaultCount);
+   } else if (parameter == "google:assistedQueryStats") {
+-    replacements->push_back(Replacement(GOOGLE_ASSISTED_QUERY_STATS, start));
```
csagan5, line 16: Will this make sure that `{google:assistedQueryStats}` disappears from the URL?

build/patches/Remove-some-params-from-search-url.patch
```diff
@@ -0,0 +1,76 @@
...
+   } else if (parameter == "google:baseURL") {
+     replacements->push_back(Replacement(GOOGLE_BASE_URL, start));
+   } else if (parameter == "google:baseSuggestURL") {
+     replacements->push_back(Replacement(GOOGLE_BASE_SUGGEST_URL, start));
+   } else if (parameter == "google:currentPageUrl") {
+-    replacements->push_back(Replacement(GOOGLE_CURRENT_PAGE_URL, start));
+   } else if (parameter == "google:cursorPosition") {
+     replacements->push_back(Replacement(GOOGLE_CURSOR_POSITION, start));
```
csagan5, line 24: This should be removed as well.

build/patches/Remove-some-params-from-search-url.patch
```diff
@@ -0,0 +1,76 @@
...
+-    replacements->push_back(Replacement(GOOGLE_SEARCH_FIELDTRIAL_GROUP, start));
+   } else if (parameter == "google:searchVersion") {
+-    replacements->push_back(Replacement(GOOGLE_SEARCH_VERSION, start));
+   } else if (parameter == "google:sessionToken") {
+-    replacements->push_back(Replacement(GOOGLE_SESSION_TOKEN, start));
+   } else if (parameter == "google:sourceId") {
+ #if defined(OS_ANDROID) || defined(OS_IOS)
+-    url->insert(start, "sourceid=chrome-mobile&");
+ #else
+-    url->insert(start, "sourceid=chrome&");
```
csagan5, line 63: I would keep this one e.g. same as Desktop search.

build/patches/Remove-some-params-from-search-url.patch
uazo: it's related to omnibox suggestions.
just uncheck the settings

uazo:
please don't merge. I have got a crash writing "chrome:/" 

Abort message: '[FATAL:autocomplete_match.cc(1333)] Check failed: i->offset > last_offset (7 vs. 16) Classification for "chrome://version/" with offset of 7 is unsorted in relation to last offset of 16. Provider: HistoryURL.'

uazo:
Some additional info.
I found the same behavior in bromite:

image

that value is strange.
activating the flag "Prevent line autocomplete":
image

it seems more correct to me.
I'm trying to figure out where the bug is

csagan5:
> Some additional info.
> I found the same behavior in bromite:

What page is that?

uazo:
> What page is that?

chrome://omnibox/

uazo:
I found the cause is "Disable omission of URL elements in Omnibox" patch.
https://github.com/bromite/bromite/blob/master/build/patches/Disable-omission-of-URL-elements-in-Omnibox.patch
it would probably be better not to change those flags, but change the caller.
remember where it is used?

csagan5:
> it would not be better to edit directly that?

If you look for each of those constants (`kFormatUrlOmitHTTP` etc), they are used in many places; editing that file will not solve this.

> I found the cause is "Disable omission of URL elements in Omnibox" patch.

In which way? Is it because of a specific flag?

uazo:
> In which way? Is it because of a specific flag?

Sorry, I didn't have time to look at it.

however i tried that by putting the flags back and applying this

```
diff --git a/components/url_formatter/android/java/src/org/chromium/components/url_formatter/UrlFormatter.java b/components/url_formatter/android/java/src/org/chromium/components/url_formatter/UrlFormatter.java
index 3ad37094fc..37ba76c43a 100644
--- a/components/url_formatter/android/java/src/org/chromium/components/url_formatter/UrlFormatter.java
+++ b/components/url_formatter/android/java/src/org/chromium/components/url_formatter/UrlFormatter.java
@@ -50,7 +50,7 @@ public final class UrlFormatter {
      * @return Formatted URL.
      */
     public static String formatUrlForDisplayOmitScheme(String uri) {
-        return UrlFormatterJni.get().formatUrlForDisplayOmitScheme(uri);
+        return UrlFormatterJni.get().formatUrlForCopy(uri);
     }

     /**
@@ -70,7 +70,7 @@ public final class UrlFormatter {
      * @return Formatted URL.
      */
     public static String formatUrlForDisplayOmitHTTPScheme(String uri) {
-        return UrlFormatterJni.get().formatUrlForDisplayOmitHTTPScheme(uri);
+        return UrlFormatterJni.get().formatUrlForCopy(uri);
     }

     /**
@@ -90,7 +90,7 @@ public final class UrlFormatter {
      * @return Formatted URL.
      */
     public static String formatUrlForDisplayOmitSchemeOmitTrivialSubdomains(String uri) {
-        return UrlFormatterJni.get().formatUrlForDisplayOmitSchemeOmitTrivialSubdomains(uri);
+        return UrlFormatterJni.get().formatUrlForCopy(uri);
     }
     /**
      * Builds a String representation of <code>uri</code> suitable for display to the user,
@@ -108,7 +108,7 @@ public final class UrlFormatter {
      * @return Formatted URL.
      */
     public static String formatUrlForDisplayOmitUsernamePassword(String uri) {
-        return UrlFormatterJni.get().formatUrlForDisplayOmitUsernamePassword(uri);
+        return UrlFormatterJni.get().formatUrlForCopy(uri);
     }

     /**
@@ -131,7 +131,7 @@ public final class UrlFormatter {
      *         it fails to parse it.
      */
     public static String formatUrlForSecurityDisplay(String uri) {
-        return UrlFormatterJni.get().formatStringUrlForSecurityDisplay(uri, SchemeDisplay.SHOW);
+        return UrlFormatterJni.get().formatUrlForCopy(uri);
     }

     /**
@@ -143,7 +143,7 @@ public final class UrlFormatter {
      */
     public static String formatUrlForSecurityDisplay(GURL url, @SchemeDisplay int schemeDisplay) {
         if (url == null) return "";
-        return UrlFormatterJni.get().formatUrlForSecurityDisplay(url, schemeDisplay);
+        return UrlFormatterJni.get().formatUrlForCopy(url.getSpec());
     }

     /**
@@ -153,7 +153,7 @@ public final class UrlFormatter {
      */
     @Deprecated
     public static String formatUrlForSecurityDisplay(String uri, @SchemeDisplay int schemeDisplay) {
-        return UrlFormatterJni.get().formatStringUrlForSecurityDisplay(uri, schemeDisplay);
+        return UrlFormatterJni.get().formatUrlForCopy(uri);
     }

     @VisibleForTesting
```

everything works perfectly.

csagan5:
> If you look for each of those constants (`kFormatUrlOmitHTTP` etc), they are used in many places; editing that file will not solve this.

See my previous comment.

build/patches/Remove-some-params-from-search-url.patch
csagan5: Which settings would these be?

uazo:
>> If you look for each of those constants (kFormatUrlOmitHTTP etc), they are used in many places; editing that file will not solve this.

> See my previous comment.

what do you think if we close this pr? what you ask me is impossible... the flag is used everywhere in the code, understanding what changes is a big problem.
maybe we may think of giving the possibility to edit by the user?

csagan5:
> what do you think if we close this pr? what you ask me is impossible... the flag is used everywhere in the code, understanding what changes is a big problem.
> maybe we may think of giving the possibility to edit by the user?

What I am asking is that merging the PR does not make URL elisions re-appear somewhere in the browser. If it is possible then we can change the current patch, otherwise not because we would loose that feature.

uazo:
> If you look for each of those constants (kFormatUrlOmitHTTP etc), they are used in many places

I'm sorry to come back to this, but I think you should rethink about your position.
for example, this code expects to have only that value in `domain`, while in bromite, being `kFormatUrlOmitTrivialSubdomains`  zero, the whole url would return, bringing bugs not easily fixed or worse security problems.

```
std::vector<std::string> GetMatchingDomains(
    const std::vector<password_manager::MatchingReusedCredential>&
        matching_reused_credentials) {
  std::vector<std::string> matching_domains;
  matching_domains.reserve(matching_reused_credentials.size());
  for (const auto& credential : matching_reused_credentials) {
    // This only works for Web credentials. For Android credentials, there needs
    // to be special handing and should use affiliation information instead of
    // the signon_realm.
    std::string domain = base::UTF16ToUTF8(url_formatter::FormatUrl(
        GURL(credential.signon_realm),
        url_formatter::kFormatUrlOmitDefaults |
            url_formatter::kFormatUrlOmitHTTPS |
            url_formatter::kFormatUrlOmitTrivialSubdomains |
            url_formatter::kFormatUrlTrimAfterHost,
        net::UnescapeRule::SPACES, nullptr, nullptr, nullptr));
    matching_domains.push_back(std::move(domain));
  }
  return base::flat_set<std::string>(std::move(matching_domains)).extract();
}
```

note, this is just an example, but code that relies on those flags for things that have nothing to do with display there is all over the place.

> What I am asking is that merging the PR does not make URL elisions re-appear somewhere in the browser

Please allow me to edit the `Disable-omission-of-URL-elements-in-Omnibox.patch`, if, subsequently, elided urls emerge we will fix them.

csagan5:
> bringing bugs not easily fixed or worse security problems.

This patch has been used for long in Bromite and I do not recall of bugs related to it.

> if, subsequently, elided urls emerge we will fix them.

This is an approach of the type "let the users suffer the bugs and report them", and I do not agree with it.

The simplified domain display has been abandoned: https://bugs.chromium.org/p/chromium/issues/detail?id=1090393#c75

So the patch could be updated accordingly, but I am afraid it is covering omissions that were introduced also before that experiment.

uazo:
> This patch has been used for long in Bromite and I do not recall of bugs related to it.

bugs maybe not (although in my experience, many users never complain) but `dchecks` failed a bit.

> This is an approach of the type "let the users suffer the bugs and report them", and I do not agree with it.

and neither do I in principle.
however I think that modification is too invasive and incorrect. I'd rather have a small UI bug than a big hidden one

> So the patch could be updated accordingly, but I am afraid it is covering omissions that were introduced also before that experiment.

before writing to you I checked in the history of cr.

if you look in https://chromium.googlesource.com/chromium/src/+blame/refs/heads/main/chrome/browser/flag_descriptions.cc for `kOmniboxUIHideSteadyStateUrlTrivialSubdomainsName` or `kOmniboxUIHideSteadyStateUrlSchemeName` you will see that, at that time, it was easy to go back to the code that disables the elide, but now it is no longer so, and in my opinion simply because the developers used, for something else, those flags that previously had been done only for displaying.

it's a correct principle, I make my team follow it too, if it's possible (and make sense) not to duplicate code.

PR #740, opened 2020-09-13
[Brave link Update]

Nitin1818:


csagan5:
They deleted that repository; the actual Chromium patches are now in https://github.com/brave/brave-core.

Fixed in d2d29d3.

PR #757, opened 2020-09-30
update timezone override patch for v86

uazo:
patch for v86.

### DIFF
--- a/build/bromite_patches_list.txt
+++ b/build/bromite_patches_list.txt
@@ -148,3 +148,4 @@ Block-all-connection-requests-with-qjz9zk-in-the-domain-name-or-with-a-trk-schem
 Hardening-against-incognito-mode-detection.patch
 Automated-domain-substitution.patch
 Remove-weblayer-dependency-on-Play-Services.patch
+Timezone-customization.patch
--- a/build/patches/Timezone-customization.patch
+++ b/build/patches/Timezone-customization.patch
@@ -1,0 +1,1365 @@
+From: uazo <uazo@users.noreply.github.com>
+Date: Wed, 30 Sep 2020 07:40:01 +0000
+Subject: Timezone customization
+
+Allow specifying a custom timezone, or using a random one.
+
+See also: https://github.com/bromite/bromite/wiki/TimezoneOverride
+---
+ .../ChromeSiteSettingsHelpClient.java         |  18 ++
+ .../browser_ui/site_settings/android/BUILD.gn |   3 +
+ .../res/layout/time_zone_select_dialog.xml    |  36 ++++
+ ...ezoneoverride_site_settings_preference.xml |  68 ++++++
+ .../res/xml/single_website_preferences.xml    |   2 +
+ .../res/xml/site_settings_preferences.xml     |   4 +
+ .../java/res/xml/website_preferences.xml      |   9 +
+ .../ContentSettingsResources.java             |  29 +++
+ .../site_settings/SingleCategorySettings.java |  73 ++++++-
+ .../site_settings/SingleWebsiteSettings.java  |  20 ++
+ .../site_settings/SiteSettings.java           |   2 +-
+ .../site_settings/SiteSettingsCategory.java   |  11 +-
+ .../site_settings/SiteSettingsHelpClient.java |   7 +
+ ...imezoneOverrideSiteSettingsPreference.java | 193 ++++++++++++++++++
+ .../browser_ui/site_settings/Website.java     |  10 +
+ .../WebsitePermissionsFetcher.java            |   4 +
+ .../WebsitePreferenceBridge.java              |  12 ++
+ .../android/website_preference_bridge.cc      |  16 ++
+ .../strings/android/site_settings.grdp        |  35 ++++
+ .../browser/content_settings_pref_provider.cc |  16 ++
+ .../browser/content_settings_pref_provider.h  |   5 +
+ .../core/browser/content_settings_registry.cc |  12 ++
+ .../core/browser/content_settings_utils.cc    |   7 +
+ .../core/browser/host_content_settings_map.cc |   8 +
+ .../core/browser/host_content_settings_map.h  |   3 +
+ .../core/common/content_settings.cc           |   4 +-
+ .../core/common/content_settings.h            |   2 +
+ .../core/common/content_settings.mojom        |   2 +
+ .../common/content_settings_mojom_traits.cc   |   4 +-
+ .../common/content_settings_mojom_traits.h    |  10 +
+ .../core/common/content_settings_types.h      |   3 +
+ .../core/common/pref_names.cc                 |   3 +
+ .../content_settings/core/common/pref_names.h |   2 +
+ .../renderer/content_settings_agent_impl.cc   |  88 ++++++++
+ .../renderer/content_settings_agent_impl.h    |   4 +
+ .../WebLayerSiteSettingsClient.java           |   3 +
+ 36 files changed, 717 insertions(+), 11 deletions(-)
+ create mode 100755 components/browser_ui/site_settings/android/java/res/layout/time_zone_select_dialog.xml
+ create mode 100755 components/browser_ui/site_settings/android/java/res/layout/timezoneoverride_site_settings_preference.xml
+ create mode 100755 components/browser_ui/site_settings/android/java/src/org/chromium/components/browser_ui/site_settings/TimezoneOverrideSiteSettingsPreference.java
+
+diff --git a/chrome/android/java/src/org/chromium/chrome/browser/site_settings/ChromeSiteSettingsHelpClient.java b/chrome/android/java/src/org/chromium/chrome/browser/site_settings/ChromeSiteSettingsHelpClient.java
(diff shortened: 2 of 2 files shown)


csagan5:
Thanks!

csagan5:
@uazo I will commit the updated patches for v87 soon so that you can update this patch as well; my patches might not compile without errors but at least will allow you to update the patch meanwhile.

csagan5:
@uazo done

uazo:
great work!
I'll start tomorrow.

I was trying too but I'm stuck on `Disable-safe-browsing.patch`! but I swear that I will make it!
A silly question: but why not just leave it active with the offline database?
Another thing about version: last v87 version I found is 87.0.4280.62, you have used .57, I can ask you why?

csagan5:
> I was trying too but I'm stuck on `Disable-safe-browsing.patch`! but I swear that I will make it!

It's double work; there is already the folks at ungoogled-chromium and Vanadium doing similar patch updates work. You can use already the patches I updated, right?

> A silly question: but why not just leave it active with the offline database?

You are probably referring to this: https://github.com/GrapheneOS/Vanadium/issues/35
I am not sure if "offline" really means offline in all cases; safe browsing used to be possible to disable completely during build.

> Another thing about version: last v87 version I found is 87.0.4280.62, you have used .57, I can ask you why?

Because when I started updating the patches it was .57

uazo:
> It's double work; there is already the folks at ungoogled-chromium and Vanadium doing similar patch updates work. You can use already the patches I updated, right?

yes, yes, of course, I just want to try while I was waiting for you..
I have seen that some patch are easily applied with wiggle, others (mine!) require more work, I would like to understand how to optimize them for less maintenance

> I am not sure if "offline" really means offline in all cases;

from what I've seen it would seem so. would have to be investigated

> safe browsing used to be possible to disable completely during build

so, why that patch?

> Because when I started updating the patches it was .57

Ah, ok, sure!

another question (sorry if I disturb you), about the patch order, how do you decide it?

csagan5:
> I would like to understand how to optimize them for less maintenance

There are some tricks, but in some cases nothing can be done.





> so, why that patch?

*Used*, it is not possible anymore to disable it. Without the patch it cannot be disabled.

> about the patch order, how do you decide it?

They are always by date of creation, unless they touch similar files; in those cases it is better to cluster them together.

csagan5:
@uazo there is currently a problem with the patch for autoplay:
```
gen/components/browser_ui/site_settings/android/site_settings_jni_headers/WebsitePreferenceBridge_jni.h:178:13: error: function 'JNI_WebsitePreferenceBridge_GetAutoplayOrigins' has internal linkage but is not defined [-Werror,-Wundefined-internal]
static void JNI_WebsitePreferenceBridge_GetAutoplayOrigins(JNIEnv* env, const
            ^
gen/components/browser_ui/site_settings/android/site_settings_jni_headers/WebsitePreferenceBridge_jni.h:187:10: note: used here
  return JNI_WebsitePreferenceBridge_GetAutoplayOrigins(env,
         ^
```

I have updated the patches again; I can leave this issue to you. Let me know if you find a fix.

uazo:
What's about this?

```
./../components/safe_browsing/core/file_type_policies.cc:53:40: error: use of undeclared identifier 'IDR_DOWNLOAD_FILE_TYPES_PB'
  return bundle.LoadDataResourceString(IDR_DOWNLOAD_FILE_TYPES_PB);
```

do you have same bug with yours?
Do I give up or do you want me to fix this too?

uazo:
>> there is currently a problem with the patch for autoplay

> has internal linkage but is not defined

I'll take care of it, usually that error is due to the wrong number of arguments between `@native` in java and the counterpart c++

csagan5:
> do you have same bug with yours?

No, but it could be because I did not start from scratch with the build.

csagan5:
The `IDR_DOWNLOAD_FILE_TYPES_PB` is reproducible also for me when skipping the autoplay patch, which is currently broken.

I will try to fix this one about `IDR_DOWNLOAD_FILE_TYPES_PB`; if you can fix the autoplay and the timezone patch we can include them in the release for v87.

uazo:
I have already fixed several compilation problems, I still don't know if chromium works. now I am waiting for the compilation to finish, later I will send them to you and then I do a pr with the fixup! (if you need it and if you want to compare it)

> if you can fix the autoplay and the timezone patch we can include them in the release for v87.

yes, as soon as i can compile

csagan5:
Did you already fix the `IDR_DOWNLOAD_FILE_TYPES_PB` problem?

uazo:
for now I've changed with

```
std::string FileTypePolicies::ReadResourceBundle() {
  //ui::ResourceBundle& bundle = ui::ResourceBundle::GetSharedInstance();
  return nullptr; // bundle.LoadDataResourceString(IDR_DOWNLOAD_FILE_TYPES_PB);
}
```

then I'll find if it's ok.
that file is the compiled version of `components/safe_browsing/core/resources/download_file_types.asciipb`, I think it's only for safe_browsing

uazo:
look there is a rather difficult bug to find caused by `chrome/android/java/res/values/values.xml` (at least, it took me a while).

if you don't delete the strings

```
allow_custom_tab_intents_title
allow_custom_tab_intents_summary
close_tabs_on_exit_title
close_tabs_on_exit_summary
```

can't compile with indecipherable error (for me)

csagan5:
> then I'll find if it's ok.

I was also thinking to remove those methods; I will try to compile with them removed.

> look there is a rather difficult bug to find caused by `chrome/android/java/res/values/values.xml`

What is the solution? They need to be moved elsewhere?

uazo:
> What is the solution? They need to be moved elsewhere?

in my opinion they are not really needed, maybe they are the typo of some old version. at least when I had added the strings in the patches it was enough for me to insert in the `.grd`.
if It run I can confirm it.

csagan5:
Yes, they should be in the `.grd`; please remove them from `values.xml`.

uazo:
another cryptic error

```
FAILED: libchrome__combined.so libchrome__combined.so.TOC lib.unstripped/libchrome__combined.so
python "../../build/toolchain/gcc_solink_wrapper.py" --readelf="../../third_party/android_ndk/toolchains/llvm/prebuilt/linux-x86_64/bin/aarch64-linux-android-readelf" --nm="../../third_party/android_ndk/toolchains/llvm/prebuilt/linux-x86_64/bin/aarch64-linux-android-nm" --strip=../../buildtools/third_party/eu-strip/bin/eu-strip  --sofile="./lib.unstripped/libchrome__combined.so" --tocfile="./libchrome__combined.so.TOC" --output="./libchrome__combined.so" -- ../../third_party/llvm-build/Release+Asserts/bin/clang++ -shared -Wl,-soname="libchrome__combined.so" -Wl,--version-script=gen/chrome/android/libchrome_linker_script.txt -Wl,--hash-style=gnu -Wl,-soname,libchrome.so --link-only -Wl,--fatal-warnings -Wl,--build-id -fPIC -Wl,-z,noexecstack -Wl,-z,relro -Wl,-z,now -Wl,-z,defs -Wl,--as-needed -fuse-ld=lld -Wl,-z,max-page-size=4096 -Wl,--color-diagnostics -Wl,--no-rosegment -Wl,--no-call-graph-profile-sort -Wl,--exclude-libs=libgcc.a -Wl,--exclude-libs=libvpx_assembly_arm.a --target=aarch64-linux-android21 -Wl,-mllvm,-enable-machine-outliner=never -Werror -Wl,--gc-sections -nostdlib++ --sysroot=../../third_party/android_ndk/toolchains/llvm/prebuilt/linux-x86_64/sysroot -B../../third_party/android_ndk/toolchains/llvm/prebuilt/linux-x86_64 -Wl,--warn-shared-textrel -Wl,--pack-dyn-relocs=android -Wl,-wrap,calloc -Wl,-wrap,free -Wl,-wrap,malloc -Wl,-wrap,memalign -Wl,-wrap,posix_memalign -Wl,-wrap,pvalloc -Wl,-wrap,realloc -Wl,-wrap,valloc -Wl,-wrap,strdup -Wl,-wrap,strndup -Wl,--dynamic-linker,/system/bin/linker64 -o "./lib.unstripped/libchrome__combined.so" @"./libchrome__combined.so.rsp"
ld.lld: error: undefined symbol: ProfileWriter::ProfileWriter(Profile*)
>>> referenced by bookmark_bridge.cc:714 (../../chrome/browser/android/bookmarks/bookmark_bridge.cc:714)
>>>               browser/bookmark_bridge.o:(BookmarkBridge::FileSelected(base::FilePath const&, int, void*)) in archive obj/chrome/browser/libbrowser.a

```

I try to solve. any suggestions?

csagan5:
I would guess that `profile_writer.cc` is not in the corresponding `BUILD.gn`, or the constructor signature is invalid. I can fix this easily.

uazo:
yes, I did that in `chrome/browser/BUILD.gn`

uazo:
so, it's running! now I'll prepare the list of changes I made

uazo:
https://gist.github.com/uazo/89608436196ac921a0ac25d0deff7d38

how do we divide the work?

csagan5:
@uazo create a PR when you are done with the changes; I do not have other changes to submit and will wait for yours.

Issue #772, opened 2020-10-08
permanent desktop view mode


furkanunluturkk:
we tablet users use our devices mostly as a laptop replacement and when we browse in internet we'd like to see desktop versions of websites. i don't have google play on my device hence number of available browsers are limited. Bromite is a great browser, fastest browser I've ever used. however seeing the mobile versions of websites draws me back from using Bromite.
it'd be great to have that feature.

PR #787, opened 2020-10-24
Add flag to always view the desktop site for all websites

### LINKED ISSUE #780: feature request: "always view the desktop site for all websites"

No.

I know a couple of browsers have that feature but don't know anything about a related patch.

An option in settings/preferences which make the browser view the desktop version for all the websites user visits. That is a killing feature for tablet users.

Support for "Chrome Extensions" so that user can install a "user agent switcher" extension and that extension handles the job.

uazo:
fix #780 
we have to choose where to put the flag:

in site setting: +4 files
In homepage settings: +2 files

in my opinion last is better because that flag is linked to the tab.
that is, if I change the flag I have to close the tab, otherwise it does not take the change.

### DIFF
--- a/build/patches/Add-a-flag-to-always-view-desktop-site-for-all-websites.patch
+++ b/build/patches/Add-a-flag-to-always-view-desktop-site-for-all-websites.patch
@@ -1,157 +1,348 @@
 From: uazo <uazo@users.noreply.github.com>
-Date: Sat, 24 Oct 2020 19:40:04 +0000
-Subject: Add a flag to always view the desktop site for all websites
+Date: Mon, 26 Oct 2020 11:55:40 +0000
+Subject: Add flag to always view the desktop site for all websites
 
 ---
- .../java/res/xml/homepage_preferences.xml      |  5 +++++
- .../chrome/browser/WebContentsFactory.java     | 18 ++++++++++++++++--
- .../homepage/settings/HomepageSettings.java    | 16 ++++++++++++++++
- .../ChromeSiteSettingsClient.java              | 14 ++++++++++++++
- .../preferences/ChromePreferenceKeys.java      |  5 ++++-
- .../java/res/xml/site_settings_preferences.xml |  4 ++++
- .../browser_ui/site_settings/SiteSettings.java | 14 ++++++++++++++
- .../site_settings/SiteSettingsClient.java      |  3 +++
- .../strings/android/browser_ui_strings.grd     |  4 ++++
- .../navigation_controller_android.cc           |  7 +++++++
- .../frame_host/navigation_controller_android.h |  4 ++++
- .../frame_host/navigation_controller_impl.cc   |  6 ++++++
- .../frame_host/navigation_controller_impl.h    |  4 ++++
- .../framehost/NavigationControllerImpl.java    | 11 +++++++++++
- .../browser/NavigationController.java          |  2 ++
- 15 files changed, 114 insertions(+), 3 deletions(-)
+ chrome/android/chrome_java_resources.gni      |  1 +
+ chrome/android/chrome_java_sources.gni        |  1 +
+ .../android/java/res/xml/main_preferences.xml |  5 ++
+ .../java/res/xml/useragent_preferences.xml    | 30 ++++++++++
+ .../chrome/browser/app/ChromeActivity.java    | 15 ++++-
+ .../settings/UserAgentPreferences.java        | 56 +++++++++++++++++++
+ .../chromium/chrome/browser/tab/TabImpl.java  | 47 +++++++++++++++-
+ .../browser/tabmodel/TabWindowManager.java    | 20 +++++++
+ chrome/browser/android/tab_android.cc         |  5 +-
+ chrome/browser/android/tab_android.h          |  3 +-
+ .../preferences/ChromePreferenceKeys.java     |  5 +-
+ .../org/chromium/chrome/browser/tab/Tab.java  |  2 +
+ .../strings/android_chrome_strings.grd        | 14 +++++
+ .../navigation_controller_android.cc          |  6 +-
+ .../navigation_controller_android.h           |  3 +-
+ .../framehost/NavigationControllerImpl.java   |  6 +-
+ 16 files changed, 207 insertions(+), 12 deletions(-)
+ create mode 100644 chrome/android/java/res/xml/useragent_preferences.xml
+ create mode 100644 chrome/android/java/src/org/chromium/chrome/browser/settings/UserAgentPreferences.java
 
-diff --git a/chrome/android/java/res/xml/homepage_preferences.xml b/chrome/android/java/res/xml/homepage_preferences.xml
---- a/chrome/android/java/res/xml/homepage_preferences.xml
-+++ b/chrome/android/java/res/xml/homepage_preferences.xml
-@@ -12,6 +12,11 @@
-         android:summaryOn="@string/text_on"
-         android:summaryOff="@string/text_off" />
- 
-+    <org.chromium.components.browser_ui.settings.ChromeSwitchPreference
-+        android:key="always_desktop_mode_switch"
-+        android:summaryOn="@string/text_always_desktop_mode"
-+        android:summaryOff="@string/text_always_desktop_mode" />
-+
+diff --git a/chrome/android/chrome_java_resources.gni b/chrome/android/chrome_java_resources.gni
+--- a/chrome/android/chrome_java_resources.gni
++++ b/chrome/android/chrome_java_resources.gni
+@@ -1130,4 +1130,5 @@ chrome_java_resources = [
+   "java/res/xml/sync_and_services_preferences.xml",
+   "java/res/xml/theme_preferences.xml",
+   "java/res/xml/tracing_preferences.xml",
++  "java/res/xml/useragent_preferences.xml",
+ ]
+diff --git a/chrome/android/chrome_java_sources.gni b/chrome/android/chrome_java_sources.gni
+--- a/chrome/android/chrome_java_sources.gni
++++ b/chrome/android/chrome_java_sources.gni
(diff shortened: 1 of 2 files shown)


csagan5:
If the site settings approach is too cumbersome then I would suggest to create its own menu section called `User Agent`; in future other options might be there as well.

uazo:
uhm .. is there already a patch that does that?

csagan5:
No, I am proposing to create a new menu entry; you can see how a new menu entry was added for the `AdBlock settings`, it is not very complicated.

As for the feature itself: the toggle is already available from the burger menu, this new setting should not conflict with that. Is the new setting about "Toggle desktop mode for all sites" or "Always desktop mode"? I think it is more about the first, since the toggling still happens from the burger menu.

uazo:
> I think it is more about the first, since the toggling still happens from the burger menu.

yes, it's more of a default.
the flag on the burger can be turned off and the page would appear not in desktop mode. do i have to take it off?

csagan5:
This is how I think the feature should work (ideally):

* toggling happens only from the burger menu, as it is in current Chromium/Bromite
* if one wants the setting to "stick" and work for all sites, then goes to Site settings and enables this new option; toggling the new option alone does not affect the user agent, which is still governed by the burger menu toggle, but only whether the UA switches apply to the current tab or all tabs

If we might have User Agent customization in future, then having this option on its own sub-menu from the Main settings is instead better.

uazo:
> but only whether the UA switches apply to the current tab or all tabs

can you explain it to me better?

> If we might have User Agent customization in future, then having this option on its own sub-menu from the Main settings is instead better.

why change the UA? what's the point of changing the UA? I would already know where to act, I saw where code build the new UA

csagan5:
> can you explain it to me better?

Current behaviour: the toggle in the burger menu is relative to the current tab
New behaviour: the toggle in the burger menu is relative to the current tab OR to all tabs depending on the new setting that we introduce here

> why change the UA? what's the point of changing the UA? I would already know where to act, I saw where code build the new UA

Cannot be in this PR; some users would like for example to present themselves as another browser, or as a spider, and some websites will server rather different content based on that. You can search online about "user agent switching" to understand what I mean.

uazo:
> New behaviour: the toggle in the burger menu is relative to the current tab OR to all tabs depending on the new setting that we introduce here

So, with A=new flag, B=burger flag

1) A=false then actual behaviour
    new tab -> default B=false
    decide B 

2) A=true then
    new tab -> default B=true
    decide B
    if user change B=false --> new tab --> default B=false until browser is closed?


> some users would like for example to present themselves as another browser, or as a spider, and some websites will server rather different content based on that

i know the principle, i use it on pc for work. but on the mobile for what purpose?

csagan5:
No, the new flag must not toggle desktop/non-desktop, just whether it is applying to *all* tabs ("sticking" ON or OFF when switching tabs).

Try it yourself: open 2 tabs. Switch to desktop mode on the 1st tab. Now open the 2nd tab: desktop mode is not enabled. This behaviour has to be preserved when the new option is **not** enabled.

When the new options is enabled: opening the 2nd tab would **also** have desktop mode.

uazo:
> When the new options is enabled: opening the 2nd tab would also have desktop mode.

Ok, it's so. My patch does that.

csagan5:
It seems to apply the behaviour only to new tabs, so if there are pre-existing tabs with desktop or non-desktop mode this would become inconsistent from user perspective, no?

Also, it says "always-desktop mode", but what we want is "Toggle desktop mode for all tabs"

uazo:
> Toggle desktop mode for all tabs

where it is written?

csagan5:
It is the only sensible option without removing the toggle from the burger menu.

> It seems to apply the behaviour only to new tabs, so if there are pre-existing tabs with desktop or non-desktop mode this would become inconsistent from user perspective, no?

What about this question?

uazo:
>> It seems to apply the behaviour only to new tabs, so if there are pre-existing tabs with desktop or non-desktop mode this would become inconsistent from user perspective, no?

> What about this question?

Yes, but just close that tab or adjust it via burger menu or open a new tab or restart the browser.
User say "An option in settings/preferences which make the browser view the desktop version for all the websites user visits" and #772 I have understood, from then on.

csagan5:
@uazo as we discussed in the past before starting the implementation we should discuss how the feature is going to look like to avoid double work; the next step after user feedback is not the PR but discussing the feature itself. The feature is not defined by the user submission alone; in this case (and any other case) we cannot introduce inconsistencies otherwise they will just come back as bug reports (as it has already happened).

This is yet another case where touching non-privacy-related features leads to lots of work which are outside the project scope.

uazo:
> otherwise they will just come back as bug reports (as it has already happened).

ok, i finally understand your point of view and i think you are right. I'll try to make the changes you asked for.

> we should discuss how the feature is going to look like to avoid double work

you're right about this too

> This is yet another case where touching non-privacy-related features leads to lots of work which are outside the project scope.

not a problem for me, really.
I want bromite to get better and better.

csagan5:
thanks @uazo, I really appreciate your work and it might be sufficient for me and another 99% of the users, but once there is a bug report about the inconsistency we must accept it and also fix it, that is why I insist on avoiding introducing inconsistencies

uazo:
thanks to your persistence I found another way that I like more.
among other things, I'm curious to see how those of brave did it (later I look at it).
tell me if there is something wrong

obviously I leave the options in English to you!

uazo:
> toggling the new option alone does not affect the user agent

ah ops, I forgot this.

build/patches/Add-a-flag-to-always-view-desktop-site-for-all-websites.patch
```diff
@@ -0,0 +1,471 @@
...
+ import java.util.ArrayList;
+ import java.util.List;
+@@ -1987,11 +1990,19 @@ public abstract class ChromeActivity<C extends ChromeActivityComponent>
+         } else if (id == R.id.view_source_id) {
+             currentTab.getWebContents().getNavigationController().loadUrl(new LoadUrlParams("view-source:"+currentTab.getUrlString()));
+         } else if (id == R.id.request_desktop_site_id || id == R.id.request_desktop_site_check_id) {
+-            final boolean reloadOnChange = !currentTab.isNativePage();
+             final boolean usingDesktopUserAgent =
+                     currentTab.getWebContents().getNavigationController().getUseDesktopUserAgent();
+-            currentTab.getWebContents().getNavigationController().setUseDesktopUserAgent(
```
csagan5, line 116: `setUseDesktopUserAgent` is used only here; it is OK to leave it behind (unless it causes build failure, will see) but I think we should preserve the `reloadOnChange` behavior. I am assuming that it is an optimisation used because native pages do not need to reload when agent changes.

No need to pass an extra parameter here, it can be checked in the plumbing of `SetOverrideUserAgentForAllTabs`/`SetOverrideUserAgent`. What do you think?

csagan5:
> thanks to your persistence I found another way that I like more.
> among other things, I'm curious to see how those of brave did it (later I look at it).
> tell me if there is something wrong
> 
> obviously I leave the options in English to you!

I left a comment; it looks impressive so far! Many thanks.

Brave seems to have done it like in your previous version: https://github.com/bromite/bromite/issues/780#issuecomment-716124810

I have not checked their commits, features are never grouped into separate commits so it is hard to find them.

Regardless, it does not change anything about what we discussed, we make our own conclusions and choices for the features of Bromite.

build/patches/Add-a-flag-to-always-view-desktop-site-for-all-websites.patch
uazo: > `setUseDesktopUserAgent` is used only here; 

I moved that call to `currentTab.SetOverrideUserAgent()` for code cleanup, but of course you can leave it there

> but I think we should preserve the reloadOnChange behavior

yes, it' so, see line 262

> I am assuming that it is an optimisation used because native pages do not need to reload when agent changes

yes are java ui

> No need to pass an extra parameter here, it can be checked in the plumbing `SetOverrideUserAgentForAllTabs`/`SetOverrideUserAgent`

yes, absolutely not necessary, do you want me to change it?

uazo:
> I have not checked their commits, features are never grouped into separate commits so it is hard to find them.

Yes, I saw, I thought it was easier to read. 
I just don't understand how they handle that code, quite challenging

build/patches/Add-a-flag-to-always-view-desktop-site-for-all-websites.patch
csagan5: The parameter is not there, so there is nothing to change.

csagan5 (approved):

csagan5:
@uazo I would say this is ready for merge; have you tested behaviour with different combinations?

uazo:
> have you tested behaviour with different combinations?

I tried it with active dchecks, nothing breaks. tested with anonymous browsing always active and without, activating the ntp and without, browsing from the history and the recents tabs, did I forget something? Obviously it would be better to activate the test suites in bromite too, but I don't know yet how google does and it and I think we lack the resources.
instead, if you mean on multiple devices, I have only tried it on the ones I own. I haven't been able to activate an android virtual machine yet, my processor doesn't allow it without breaking hyper-v support, which I need to keep active, if you know some other way, I'm all ears

I realize, it's unprofessional, that's probably something we should fix.
Sure I am available to fix any bugs

csagan5:
> Obviously it would be better to activate the test suites in bromite too, but I don't know yet how google does and it and I think we lack the resources.

Yes, that is the main problem.

> instead, if you mean on multiple devices

No, I meant the feature itself e.g. existing tabs vs new tabs with/without the new setting enabled.

> I haven't been able to activate an android virtual machine yet, my processor doesn't allow it without breaking hyper-v support

I do not follow; your processor has Hyper-V support?

Issue #469, opened 2020-01-18
Consider turning "Search Ready Omnibox" ON by default


Sangeppato:
### Bromite version

Version: 79.0.3945.123
Arch: arm64
Android version: 10
Device model: Xiaomi Mi A2

### Is this bug about the SystemWebView?
No

### Is the bug reproducible with latest version?

Yes

### Can the bug be reproduced with corresponding Chromium version?

No

### Is the bug a crash?

No

### Describe the bug

In #350 it was incorrectly stated that it's not possible to edit the URL anymore with the "Search Ready Omnibox" feature enabled and for this reason the feature was disabled (or at least, this is what I understood). I personally prefer the newer Search Ready function and it has been the default in Chromium, Chrome and Brave for some time now, I'd ask you to consider setting it back as default, mainly because it looks to me that the only complaint about it was based on wrong assumptions.
In case you prefer not to enable the function, I'd at least suggest to mention the flag in the README, under the "Flag" section.

### Steps to reproduce the bug
 Search Ready Omnibox is disabled by default

### Expected behavior

Search Ready Omnibox should be enabled by default

csagan5:
@Sangeppato although it might seem that the default was changed only because of that issue, that is not the main reason. The biggest motivation to change it is to keep the functionality to edit/select directly without another tap.

How would you mention it in the README? There is plenty of other defaults changed which are not,  as the README explains that reading the changelog and patches is necessary to have the full picture.

Sangeppato:
I understand then.

About the README: sorry, I wasn't clear, I mean that since the flag was retired from the upstream Chromium (or at least, I can't find it), it should probably be mentioned amongst the "Flags which have been retired from upstream Chromium but are still available in Bromite".

csagan5:
Alright, I have mentioned it; thanks for spotting this. 

If many users miss the functionality we could add a FAQ for it, but it's not currently the case.

Sangeppato:
Got it, thanks!

csagan5:
@Sangeppato next time feel free to make a PR to the README!

Issue #404, opened 2019-11-06
"Search ready" omnibox unavailable


brianclemens:
### Bromite version
Version: `78.0.3904.72`
Arch: `arm64`
Android version: `8.1.0`
Device model: `angler`

### Is this bug about the SystemWebView?
No

### Is the bug reproducible with latest version?
Yes

### Can the bug be reproduced with corresponding Chromium version?
No

### Is the bug a crash?
No

### Describe the bug
The chromium "search-ready" omnibox is unavailable in chrome://flags.  It should be available at chrome://flags/#enable-search-ready-omnibox.  I can see that the search-ready omnibox was disabled by default after #350, in which @jungleballs did not understand that they could edit the URL by pushing the 'pencil' button.  The new omnibox is very handy because it has one-button URL copying (instead of fidgeting with a share menu or having to select all the URL text and long-pressing to copy), so I think it's foolish to disable it by default.

### Steps to reproduce the bug
1. Go to 'chrome://flags/#enable-search-ready-omnibox'
2. Observe lack of flag

### Expected behavior
1. Go to 'chrome://flags/#enable-search-ready-omnibox'
2. Find and enable flag
3. Enjoy new omnibox

csagan5:
> in which @jungleballs did not understand that they could edit the URL by pushing the 'pencil' button. The new omnibox is very handy because it has one-button URL copying (instead of fidgeting with a share menu or having to select all the URL text and long-pressing to copy), so I think it's foolish to disable it by default.

This whole part is off-topic; we don't discuss here which feature is best or not, everyone has opinions.

The bug report is about the missing flag; upstream has completely removed the flag leaving the feature enabled and that Bromite has restored the flag to allow to use it or not; somehow the flag is not showing up and will have to look into it.

brianclemens:
#response_container_BBPPID{font-family: initial; font-size:initial; color: initial;}  Right then.  Is there a mailing list or something to discuss the decision to have the old style inbox set as default?                                                                                                                                                                                                                                                                                                                        From: notifications@github.comSent: November 6, 2019 17:22To: bromite@noreply.github.comReply-to: reply@reply.github.comCc: brian@tiuxo.com; author@noreply.github.comSubject: Re: [bromite/bromite] "Search ready" omnibox unavailable (#404)  
[...]

Issue #350, opened 2019-08-09
URL box request


jungleballs:
After the new update, when I click the URL box, it erases the URL and gives me a drop down menu. I'm not able to edit the URL of the page I'm on. Is there any way to change this so I can edit URLs?

Zoraver:
In chrome://flags, set #enable-search-ready-omnibox to disabled.

csagan5:
I will see to change the default, but @jungleballs all issues must be submitted with the correct template.

PR #796, opened 2020-10-31
Fix Omnibox flag not working

### LINKED ISSUE #789: Search Ready Omnibox flag not working
### Bromite version

Version: `86.0.4240.112`
Arch: `arm64`
Android version: `9.0`
Device model: `Galaxy S20`

Yes

No

### Describe the bug

Enabling the `Search Ready Omnibox` flag does not work, the feature stays off.

uazo:
fixes #789 as written here

https://github.com/bromite/bromite/issues/565#issuecomment-680679575

tell me if I have misunderstood

### DIFF
--- a/build/patches/Restore-Search-Ready-Omnibox-flag.patch
+++ b/build/patches/Restore-Search-Ready-Omnibox-flag.patch
@@ -7,49 +7,34 @@ This reverts commit ae458edcc8422d0815d0e82261e71fe10d7d6fc2.
 
 Disable search-ready omnibox by default
 ---
- .../suggestions/DropdownItemViewInfoListBuilder.java     | 9 +++++++--
- chrome/browser/about_flags.cc                            | 3 +++
- chrome/browser/flag-metadata.json                        | 5 +++++
- chrome/browser/flag_descriptions.cc                      | 5 +++++
- chrome/browser/flag_descriptions.h                       | 3 +++
- chrome/browser/flags/android/chrome_feature_list.cc      | 4 ++++
- chrome/browser/flags/android/chrome_feature_list.h       | 1 +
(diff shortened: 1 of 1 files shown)


csagan5:
@uazo with this patch does Bromite have no Omnibox by default and after enabling the flag Omnibox is working?

csagan5 (changes requested):
This PR does not fix the Omnibox flag, it makes Omnibox always enabled - regardless of the flag value.

uazo:
https://github.com/bromite/bromite/issues/565#issuecomment-680696966

>> if the purpose is not to "Cleanup Search Ready Omnibox"

>It's a revert of the cleanup.

No, this PR is for that.
Why the need to disable omnibox?

csagan5:
> Why the need to disable omnibox?

The default behaviour is to have Omnibox always enabled; there is a patch in Bromite to allow to disable it. What does this PR do exactly? It does not fix #789. If one wanted Omnibox always enabled then the Bromite patch could simply be dropped.

uazo:
> What does this PR do exactly? 

clear omnibox textbox

with that flag enabled:
image

disabled:
image

also see that buttons to share/copy/edit are active

csagan5:
Ok, that is an improvement, however not a fix for #789

uazo:
I honestly don't understand you.
I may be wrong, but I am referring to

https://github.com/bromite/bromite/issues/350
https://github.com/bromite/bromite/issues/469
https://github.com/bromite/bromite/issues/404

could you explain how you would like to fix it #789 ?

csagan5:
Flag enabled: like Chromium
Flag disabled: like Bromite is now

csagan5 (approved):

csagan5:
Great job @uazo! Many thanks

PR #800, opened 2020-11-06
Always desktop view mode: exclude internal schemes

uazo:
as https://github.com/bromite/bromite/issues/780#issuecomment-721597527 I added an option to exclude internal schemes.
just tell me what you think.


csagan5:
@uazo no need for a new options. We just never use the desktop UA for native pages.

uazo:
ok, I take it off.

PR #807, opened 2020-11-13
Allow custom user agent

### LINKED ISSUE #798: Allow custom user agent
Yes, it allows for greater privacy by being able to set a more generic user agent, causing the browser fingerprint to be less unique and identifyable. As a side effect, users would gain the ability to circumvent site/feature blockades that are becoming more and more common.

Not to my knowledge.

An option to manually enter a custom user agent, or at the very least an option to choose a user agent from a list.

Tried using the desktop mode, but even though it changes the user agent - it does exactly what it's intended for.. displaying pages as if they're on desktop. This is not needed, nor preferred, in most cases, when only the user agent needs to be changed.

uazo:
fixes #798 

added possibility to define a custom useragent for mobile and desktop mode.

added possibility to reactivate the metatag view for desktop mode, so that users can choose to use the flag in the hambumger menu to navigate with a custom useragent leaving the standard navigation unchanged


csagan5:
> added possibility to reactivate the metatag view for desktop mode, so that users can choose to use the flag in the hamburger menu to navigate with a custom useragent leaving the standard navigation unchanged

I do not understand this part; what does it do?

build/patches/Add-flag-to-always-view-the-desktop-site-for-all-websites.patch
```diff
@@ -471,6 +1123,38 @@ diff --git a/content/browser/frame_host/navigation_controller_android.h b/conten
...
+ #include "base/time/time.h"
+ #include "base/trace_event/trace_event.h"
+@@ -2319,7 +2320,8 @@ const WebPreferences WebContentsImpl::ComputeWebPreferences() {
+ 
+   prefs.viewport_enabled = command_line.HasSwitch(switches::kEnableViewport);
+ 
+-  if (IsOverridingUserAgent())
++  if (IsOverridingUserAgent() && 
++      !command_line.HasSwitch(switches::kDesktopModeViewportMetaEnabled))
+     prefs.viewport_meta_enabled = false;
```
uazo, line 1155: >> added possibility to reactivate the metatag view for desktop mode, so that users can choose to use the flag in the hamburger menu to navigate with a custom useragent leaving the standard navigation unchanged

> I do not understand this part; what does it do?

related code is here.
in desktop mode the viewport tag (ex: `<meta name="viewport" content="width=device-width, initial-scale=1.0">`) is ignored by blink, so the css declared for the desktop are used

csagan5:
@uazo I have pushed the v87 patches so that you can already rebase this and other work.

build/patches/Add-flag-to-always-view-the-desktop-site-for-all-websites.patch
```diff
@@ -188,11 +412,109 @@ new file mode 100644
...
++            public void onTextChanged(CharSequence newText) {
++                PrivacyPreferencesManager.getInstance().setOverrideUserAgentValue(
++                    newText.toString(), true);
++            }
++        });
++
++        return viewGroup;
++    }
++
++    // @Override
```
csagan5, line 480: If this code is not necessary, it should be deleted.

csagan5 (commented):
Can you provide some screenshots for the various modes?

uazo:
Some screenshot:

**Settings:**| **Page with viewport flag disabled (default):**|**Same page with viewport flag enabled:**
------------ | -------------| -------------
image|image|image

csagan5:
Impressive work!

build/patches/Flag-for-always-desktop-mode.patch
```diff
@@ -1,52 +1,201 @@
...
 +++ b/chrome/android/chrome_java_sources.gni
-@@ -1322,6 +1322,7 @@ chrome_java_sources = [
-   "java/src/org/chromium/chrome/browser/privacy/settings/PrivacyPreferencesManager.java",
-   "java/src/org/chromium/chrome/browser/privacy/settings/PrivacySettings.java",
-   "java/src/org/chromium/chrome/browser/provider/BaseColumns.java",
+@@ -1308,6 +1308,7 @@ chrome_java_sources = [
+   "java/src/org/chromium/chrome/browser/photo_picker/DecoderService.java",
+   "java/src/org/chromium/chrome/browser/settings/AdBlockEditor.java",
+   "java/src/org/chromium/chrome/browser/settings/AdBlockPreferences.java",
 +  "java/src/org/chromium/chrome/browser/settings/UserAgentPreferences.java",
```
csagan5, line 78: Can you move this elsewhere? Being so close to files changed in another patch will cause problems.

build/patches/Flag-for-always-desktop-mode.patch
```diff
@@ -342,6 +678,214 @@ diff --git a/chrome/android/java/src/org/chromium/chrome/browser/tabmodel/TabWin
...
++        std::string();  // match content::GetOSVersion(false) on Linux
++    spoofed_ua.ua_metadata_override->architecture = "x86";
++    spoofed_ua.ua_metadata_override->model = std::string();
++    spoofed_ua.ua_metadata_override->mobile = false;
++
++    content::WebContents* web_contents =
++        content::WebContents::FromJavaWebContents(jweb_contents);
++    web_contents->SetUserAgentOverride(spoofed_ua, false);
++  }
++  else
```
csagan5, line 749: These 3 lines should be combined in `} else {`

csagan5 (changes requested):

csagan5:
I will merge this and do some adjustments.

csagan5:
@uazo FYI it does not build the SystemWebView target:
```
ld.lld: error: undefined symbol: switches::kUserAgent
>>> referenced by string_piece.h:166 (../../base/strings/string_piece.h:166)
>>>               thinlto-cache/Thin-11b025.tmp.o:(content::BrowserStartupComplete(int))

ld.lld: error: undefined symbol: g_browser_process
>>> referenced by browser_startup_controller.cc:35 (../../content/browser/android/browser_startup_controller.cc:35)
>>>               thinlto-cache/Thin-11b025.tmp.o:(content::BrowserStartupComplete(int))

ld.lld: error: undefined symbol: prefs::kOverrideUserAgentEnabled
>>> referenced by browser_startup_controller.cc:37 (../../content/browser/android/browser_startup_controller.cc:37)
>>>               thinlto-cache/Thin-11b025.tmp.o:(content::BrowserStartupComplete(int))

ld.lld: error: undefined symbol: prefs::kOverrideUserAgent
>>> referenced by browser_startup_controller.cc:39 (../../content/browser/android/browser_startup_controller.cc:39)
>>>               thinlto-cache/Thin-11b025.tmp.o:(content::BrowserStartupComplete(int))

ld.lld: error: undefined symbol: prefs::kDesktopModeViewportMetaEnabled
>>> referenced by browser_startup_controller.cc:48 (../../content/browser/android/browser_startup_controller.cc:48)
>>>               thinlto-cache/Thin-11b025.tmp.o:(content::BrowserStartupComplete(int))
clang: error: linker command failed with exit code 1 (use -v to see invocation)

```

csagan5:
@uazo I have moved it to `chrome/browser/chrome_content_browser_client.cc`, but what is the logic behind this?

Local options should not override switches, it is the other way around: switches have priority over anything else. Perhaps you are using switches to pass options around? We need to check if everything works on next release and possibly fix this.

uazo:
> FYI it does not build the SystemWebView target

sure, never tried :(

> it is the other way around .... Perhaps you are using switches to pass options around? 

no, the first version (never released) was as you say, I had tried but, as the access is through different threads it doesn't work. Chromium itself uses switches, I guess for the same reason.

csagan5:
> Chromium itself uses switches, I guess for the same reason.

Have you seen an example in Chromium of copying local state values into switches?

uazo:
> but what is the logic behind this?

as they are currently handled by switches, and in android it is not possible to pass values to the command line (which is disabled in the non-debug version), so it is as if I entered them. that method is called in the startup phase before the native init.
then, as the switches pass to the child processes (both in startup and at runtime), I used that way to pass information to the rendering process.

> Have you seen an example in Chromium of copying local state values into switches

Features and variants

csagan5:
> that method is called in the startup phase before the native init.

This part is not correct. It should be done elsewhere, like for all the other switches in Chromium. For now I have put it in `chrome/browser/chrome_content_browser_client.cc` but I do not know if it will work; you can check the patch after the next release, it will be done shortly.

uazo:
> This part is not correct. It should be done elsewhere, like for all the other switches in Chromium

I think you can't, those parameters come from java which is initialized before the native one and the preferences are still not accessible because they are not initialized. Can you let me know exactly what you're changing? what block? from your comments I do not understand.

uazo:
>> FYI it does not build the SystemWebView target

> sure, never tried :(

thinking about this, I guess the patch should not be applied to systemwebview, it already natively supports changing of the user agent by the embedder, I have to see which #if to apply

csagan5:
There is no ifdef to achieve that; @uazo why would these flags be any different from the others? Can't we do the same as Chromium already does?

See latest version of the patch.

csagan5:
It does not work with current patch in `87.0.4280.68`, I have just pushed it; I have opened a new issue: #819

PR #816, opened 2020-11-21
Intent handling fix for local apps on Android 11

nikolowry:
See https://github.com/brave/brave-browser/issues/12330 for more details.

I was experiencing this issue for the past week on my new Pixel 5 running ProtonAOSP, and can confirm the new `QUERY_ALL_PACKAGES` permission resolved the issue.

I see there's currently 5 other patches that target `chrome/android/java/AndroidManifest.xml`, so feel free to reject this requested approach if you think there's a better way to add the permission -- whatever's easiest to maintain.

Sidebar: Google Chrome's official builds seem to have already added this permission, but they overlooked porting it back to Chromium. Wouldn't be surprised if it got forgotten during all the recent build targets shuffling

### DIFF
--- a/build/bromite_patches_list.txt
+++ b/build/bromite_patches_list.txt
@@ -152,3 +152,4 @@ Flag-for-always-desktop-mode.patch
 Restore-Simplified-NTP-launch.patch
 Revert-the-removal-of-an-option-to-block-autoplay.patch
 Automated-domain-substitution.patch
+Add-query-all-packages-permission-for-R.patch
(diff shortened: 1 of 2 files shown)


csagan5:
Thanks for your submission @nikolowry; what is the issue exactly?

> Wouldn't be surprised if it got forgotten during all the recent build targets shuffling

You mean in the code? Was it discussed in Google Groups? I am curious to read about it, I have only seen the effects in code so far.

csagan5:
The upstream commit is this: https://github.com/chromium/chromium/commit/cdc6f38e3e4a994df8c40c6f0b4b7db14da2dd35

csagan5:
I will merge this but then use upstream's commit, it seems better.

csagan5:
Fixed in `87.0.4280.68`.

nikolowry:
Thjanks @csagan5 .

Feel silly for not going back to check upstream once I found the patch. Was searching Chromium's Gerrit and Issues but never found a mention of this problem, so I wrongly assumed it wasn't upstream once I found Brave's issue/patch.

I'll make sure to double-check next time against upstream.

PR #823, opened 2020-11-24
Fix allow custom user agent patch

### LINKED ISSUE #819: User agent customization patch
The patch for user agent customization in `87.0.4280.68` does not work; original version does not allow SystemWebView to be built.

See also: #798 #807

uazo:
Fixes #819 

Changes made:
- correctly passed user agent to blink
- in settings, flag assigned at focus of the control
- loading of overrides at startup

### DIFF
--- a/build/patches/User-agent-customization.patch
+++ b/build/patches/User-agent-customization.patch
@@ -11,31 +11,33 @@ to use the flag in the hamburger menu to navigate with a custom useragent leavin
  base/base_switches.h                          |   2 +
  chrome/android/chrome_java_resources.gni      |   2 +
  chrome/android/chrome_java_sources.gni        |   1 +
- .../layout/custom_useragent_preferences.xml   | 108 +++++++++++
+ .../layout/custom_useragent_preferences.xml   | 108 ++++++++++
  .../android/java/res/xml/main_preferences.xml |   5 +
  .../java/res/xml/useragent_preferences.xml    |  31 +++
  .../chrome/browser/app/ChromeActivity.java    |  20 +-
- .../settings/PrivacyPreferencesManager.java   |  30 +++
- .../settings/UserAgentPreferences.java        | 176 ++++++++++++++++++
- .../chromium/chrome/browser/tab/TabImpl.java  |  83 ++++++++-
+ .../init/ChromeBrowserInitializer.java        |   3 +
+ .../settings/PrivacyPreferencesManager.java   |  35 ++++
+ .../settings/UserAgentPreferences.java        | 184 ++++++++++++++++++
+ .../chromium/chrome/browser/tab/TabImpl.java  |  83 +++++++-
  .../browser/tabmodel/TabWindowManager.java    |  23 +++
- .../browser/android/content/content_utils.cc  |  64 +++++--
+ .../browser/android/content/content_utils.cc  |  29 +++
  .../preferences/browser_prefs_android.cc      |   7 +
- .../privacy_preferences_manager.cc            |  93 +++++++++
+ .../privacy_preferences_manager.cc            | 113 +++++++++++
  chrome/browser/android/tab_android.cc         |   5 +-
  chrome/browser/android/tab_android.h          |   3 +-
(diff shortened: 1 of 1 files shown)


build/patches/User-agent-customization.patch
```diff
@@ -286,13 +288,42 @@ diff --git a/chrome/android/java/src/org/chromium/chrome/browser/app/ChromeActiv
...
+ import org.chromium.ui.resources.ResourceExtractor;
++import org.chromium.chrome.browser.privacy.settings.PrivacyPreferencesManager;
+ 
+ import java.io.File;
+ import java.util.ArrayList;
+@@ -324,11 +325,13 @@ public class ChromeBrowserInitializer {
+ 
+                         @Override
+                         public void onSuccess() {
++                            PrivacyPreferencesManager.getInstance().UpdateOverrideUserAgent();
```
uazo, line 306: couldn't find a better place.
unfortunately I couldn't insert it where you wanted it because I need to change the parameters of the browser process

build/patches/User-agent-customization.patch
```diff
@@ -823,11 +833,28 @@ diff --git a/chrome/browser/android/preferences/privacy_preferences_manager.cc b
 +    parsed_command_line->AppendSwitchASCII(switches::kUserAgent, ua);
 +  }
 +
++  for (auto iter = content::RenderProcessHost::AllHostsIterator(); !iter.IsAtEnd();
++      iter.Advance()) {
++    if (iter.GetCurrentValue()->IsInitializedAndNotDead()) {
++      if (overrideUserAgentEnabled) {
++        iter.GetCurrentValue()->GetRendererInterface()->SetUserAgent(ua);
```
uazo, line 840: the problem was that blink caches the value. this way I update it

build/patches/User-agent-customization.patch
```diff
@@ -1064,6 +1094,27 @@ diff --git a/chrome/common/pref_names.h b/chrome/common/pref_names.h
...
  #endif  // CHROME_COMMON_PREF_NAMES_H_
+diff --git a/components/browser_ui/widget/android/java/src/org/chromium/components/browser_ui/widget/RadioButtonWithEditText.java b/components/browser_ui/widget/android/java/src/org/chromium/components/browser_ui/widget/RadioButtonWithEditText.java
+--- a/components/browser_ui/widget/android/java/src/org/chromium/components/browser_ui/widget/RadioButtonWithEditText.java
++++ b/components/browser_ui/widget/android/java/src/org/chromium/components/browser_ui/widget/RadioButtonWithEditText.java
+@@ -167,6 +167,17 @@ public class RadioButtonWithEditText extends RadioButtonWithDescription {
+             mEditText.setCursorVisible(false);
+             KeyboardVisibilityDelegate.getInstance().hideKeyboard(mEditText);
+         }
++        if (mRadioButtonWithEditTextFocusListener != null) {
++            mRadioButtonWithEditTextFocusListener.onRadioButtonWithEditTextFocusChanged(hasFocus);
```
uazo, line 1105: it practically does not send the checked change to the focus. this way I add that functionality

build/patches/User-agent-customization.patch
csagan5: I am neutral; as far as it compiles for all targets it's fine.

csagan5:
I will merge it and try to build; thanks!

uazo:
it does not work yet! changing ua without closing the tab does not work.
but this time I don't understand, my compiled version has no problems, just double-checked, the downloaded one doesn't. I will have to investigate, but on Monday I have a go-live and for now I don't have time.

csagan5:
> my compiled version has no problems, just double-checked, the downloaded one doesn't.

On same device? Maybe different state/configuration?

The patch used is the same published here, and is identical to the one of this PR.

uazo:
> On same device? 

yes same device.

> Maybe different state/configuration?

I think the same too, but it would be a problem because by heart the only different ones are debugging in GN.
but maybe it's something more banal. 
do you find the same problem?

Issue #791, opened 2020-10-30
Revert the removal of an option to block autoplay from Chrome


baybal:
### Is your feature request related to privacy?

Yes

### Is there a patch available for this feature somewhere?

No, but I believe a quick search in chromium git will avail for a quick undoing of the change.

JS based solutions are available https://github.com/Eloston/disable-html5-autoplay

### Describe the solution you would like

Revert to autoplay blocking the way it was done in chrome 7x.x, where autoplay block was blocking all videos.

Android Chrome 82.x+ will **still play** videos even if they are autoplay blocked if they pass some obscure "engagement score:" https://www.theregister.com/2020/09/08/google_chrome_autoplay/

### Describe alternatives you have considered

Reverting to older versions works

baybal:
Removals in commit history:

https://github.com/chromium/chromium/commit/1785f13de1aede35cddacd92f1e462f7f81eff43
https://github.com/chromium/chromium/commit/f286512e7a812a4b666d549208ea6e818a5ab53e

csagan5:
Those commits are about "muted autoplay", not autoplay as a whole. We need to look deeper into how and why this was changed.

Related: https://www.theregister.com/2020/09/08/google_chrome_autoplay/

uazo:
> Removals in commit history:

I think he's right, the only real change is here,

https://github.com/chromium/chromium/commit/f286512e7a812a4b666d549208ea6e818a5ab53e#diff-817bdb7b9f231d251a99cdb41048f4b4592efa614697f02b532c88bfc30aa3edL313

everything else is for options (and bringing them to the rendering process), as well as for testing.
to re-enable the old behavior for all sites, seems it's enough

```
return true; // !(IsEligibleForAutoplayMuted();
```

i tried and it seems autoplay is off (tested with youtube **and** reloading page, the goal is perhaps to block all videos making user gesture mandatory?).
Can someone provide some other test cases?

csagan5:
Ok, so the commits mention "muted autoplay" but they are actually removing access to all of the autoplay settings.

@uazo what I think we should restore is also the `Media` site settings which contained the option to enable/disable autoplay

uazo:
yes, but maybe we should wait for the refactoring of the content settings code (by the chromium team) to finish.
for now I suggest to change only that line.

csagan5:
Do you have an issue for that refactoring? I am not aware.

uazo:
[...]

PR #825, opened 2020-11-25
fix global "Autoplay" site-setting

uazo:
fix toggling on/off in Settings > Site Settings > Autoplay as written https://github.com/bromite/bromite/issues/791#issuecomment-733887208

### DIFF
--- a/build/patches/Revert-the-removal-of-an-option-to-block-autoplay.patch
+++ b/build/patches/Revert-the-removal-of-an-option-to-block-autoplay.patch
@@ -10,15 +10,15 @@ Subject: Revert the removal of an option to block autoplay
  .../res/drawable-xxhdpi/settings_autoplay.png | Bin 0 -> 792 bytes
  .../drawable-xxxhdpi/settings_autoplay.png    | Bin 0 -> 1064 bytes
  .../res/xml/single_website_preferences.xml    |   2 ++
- .../res/xml/site_settings_preferences.xml     |   4 ++++
+ .../res/xml/site_settings_preferences.xml     |   4 +++
  .../ContentSettingsResources.java             |   5 ++++
- .../site_settings/SingleCategorySettings.java |   4 ++++
- .../site_settings/SingleWebsiteSettings.java  |  22 ++++++++++++++++++
+ .../site_settings/SingleCategorySettings.java |   4 +++
+ .../site_settings/SingleWebsiteSettings.java  |  22 +++++++++++++++++
  .../site_settings/SiteSettingsCategory.java   |   9 +++++--
- .../browser_ui/site_settings/Website.java     |   8 +++++++
+ .../browser_ui/site_settings/Website.java     |   8 ++++++
  .../WebsitePermissionsFetcher.java            |   1 +
  .../WebsitePreferenceBridge.java              |  14 +++++++++++
- .../android/website_preference_bridge.cc      |  22 ++++++++++++++++++
+ .../android/website_preference_bridge.cc      |  23 ++++++++++++++++++
  .../strings/android/site_settings.grdp        |   9 +++++++
  .../core/browser/content_settings_registry.cc |   2 +-
  .../core/browser/content_settings_utils.cc    |   2 ++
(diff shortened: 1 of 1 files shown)


csagan5:
Great job!

nikolowry:
Just ran `./bromite-builder --arch=arm64 --upstream` locally and wasn't able to apply this patch due to the trailing whitespace added by `Timezone-customization.patch` at https://github.com/bromite/bromite/blob/d446f0e453833d367348b869986bbed1254363f2/build/patches/Timezone-customization.patch#L1145

I was able to start building by removing the trailing whitespace from `Timezone-customization.patch`, adding the modified patch to a directory named `overrides` and then running: `./bromite-builder --arch=arm64 --patches=overrides --upstream`

nikolowry:
I can confirm the updated patch is working as expected. Thanks again!

uazo:
> wasn't able to apply this patch due to the trailing whitespace

it's true, I'm still a rookie in these things! only recently have I understood the importance of removing spaces ...

does anyone know any plugins for vscode that help me find these errors while developing? for now I solved it in another way, https://github.com/uazo/bromite-buildtools/blob/b34479fe9a150a63dd852eb1ba2f5d9d33bd887d/apply-single-patch.sh#L16 but I have to run the build twice and it takes too long for my taste.

@csagan5 could you make your script available to keep the timestamps of the files? thanks!

uazo:
> https://github.com/nikolowry/bromite-builder/blob/a2d051bbc35be4da950a20c58a3fb5049af811a9/bromite-builder#L127
>
>        && fetch --nohooks android

@nikolowry if you want to improve fetch times, I recommend this change:

```
git fetch --depth 2 https://chromium.googlesource.com/chromium/src.git +refs/tags/$VERSION:chromium_$VERSION
gclient sync -D

```

csagan5:
> @csagan5 could you make your script available to keep the timestamps of the files? thanks!

Sure; I will release it tomorrow.

nikolowry:
@uazo @csagan5 my script was designed for applying small customizations on-top of release builds. I could add an option for using a local Bromite git folder instead of the release archives -- then it be easy for you to run `./bromite-builder prepare` to verify everything applies cleanly.

If either of you would find that helpful, I'll make the necessary changes. Just let me know.

csagan5:
@uazo it is here: https://github.com/bromite/mtool

There is a script to show how to use it with Chromium.

uazo:
> I could add an option for using a local Bromite git folder instead of the release archives

thanks @nikolowry but I think until we have a CI it doesn't make much sense. maybe in the future :)
my personal opinion, obviously carl decides here, maybe it would be useful instead and I don't know.

nikolowry:
> thanks @nikolowry but I think until we have a CI it doesn't make much sense. maybe in the future :)

@uazo, running `./bromite-builder prepare --no-skip-patches` (with an option to use your local bromite working directory) would have alerted you to the trailing whitespace issue that I commented about earlier. Regardless of CI and/or future plans, it can be helpful now for verifying that patches are being cleanly applied to fresh builds.

> my personal opinion, obviously carl decides here, maybe it would be useful instead and I don't know.

`bromite-builder` is just a personal hobby project because I wanted some opinionated customizations (like a dark navbar) and also so I could have an excuse to work on different shell scripting patterns/techniques. However, I'll never hesitate on making changes or feature-additions if it can help you two in anyway.

---

Sidebar: @uazo I've been experimenting with ways to speed up the initial checkout -- which is the only place `fetch` is used. I tried your earlier suggestion but `gclient` threw some error about not being properly configured. However, running `fetch --nohooks --no-history android` speed things up quite a bit. I'm also thinking of changing the `gclient sync` command to `gclient sync --no-history --shallow --force -reset`.

It be greatly appreciated if you could link me to any resources with info on how to bootstrap Android builds without `depot_tools`. I need to look over the `ungoogled-chromium-android` builds and see what they ended up doing

uazo:
> but gclient threw some error about not being properly configured.

sorry you are right, that is if you have already initialized the directory, try this

```
gclient config https://chromium.googlesource.com/chromium/src.git
echo "target_os=['android']" >>.gclient

VERSION=$( cat ~/bromite/build/RELEASE )
git clone --depth 2 https://chromium.googlesource.com/chromium/src.git +refs/tags/$VERSION:chromium_$VERSION
gclient sync --no-history -D

git remote remove origin
```
I think it should still work, it's been a while since I cleaned the home.
excellent the `no-history` suggestion, thanks.

> It be greatly appreciated if you could link me to any resources with info on how to bootstrap Android without depot_tools.

sorry, I have nothing about it.


instead if you have suggestions on how to speed up the build (without upgrading the ram :) I'm all ears!
I've seen this but haven't tried it yet https://chromium.googlesource.com/infra/goma/server

nikolowry:
@uazo I'll be creating new feature branches for bromite-builder for:
- Checkout optimizations
- Using a local bromite repo instead of the release tar

I'll @ you in those pull requests when they're ready for review. Hopefully I'll be able to find some time this week for both of those

csagan5:
@uazo did you use the mtool I shared? that speeds up builds massively already. I use it in conjunction with ccache.

uazo:
> did you use the mtool I shared? 

yes, very useful. thanks it works great.

PR #830, opened 2020-12-01
Add random seed to TimezoneOverride patch

### LINKED ISSUE #828: Timezone override display the same value when it sets to "Random"
### Bromite version

Version: `87.0.4280.81`
Arch: `arm`
Android version: `10`
Device model: `Redmi 2` / `wt88047`

No

Yes

No. I did not find such feature in chromium.

### Is the bug a crash?
No

### Describe the bug
All websites that use javascript to check browser timezone display the same result when the timezone override is set to random.

Steps to reproduce the bug:
1. Go to Settings > Site settings > Timezone override. Set it to "Random"
2. Press exit button. Then open bromite again (relaunch)
3. Go to https://whatismytimezone.com/ 
4. Go to http://browserspy.dk/date.php
5. Go to https://webbrowsertools.com/timezone/
4. All websites display same timezone

### Expected behavior
At least one of those websites should display different timezone.

### Screenshots
25fcc6bd-47dc-4a9f-a9e5-abefa2e14294
81f8247d-20cd-415b-aa93-0818efdd1c33
8721dbdd-e363-48fa-b2d5-f6f9991b7157

uazo:
fixes #828

### DIFF
--- a/build/patches/Timezone-customization.patch
+++ b/build/patches/Timezone-customization.patch
@@ -39,10 +39,10 @@ See also: https://github.com/bromite/bromite/wiki/TimezoneOverride
  .../core/common/content_settings_types.h      |   3 +
  .../core/common/pref_names.cc                 |   3 +
  .../content_settings/core/common/pref_names.h |   2 +
- .../renderer/content_settings_agent_impl.cc   |  88 ++++++++
+ .../renderer/content_settings_agent_impl.cc   |  89 ++++++++
  .../renderer/content_settings_agent_impl.h    |   4 +
  .../WebLayerSiteSettingsClient.java           |   3 +
- 36 files changed, 713 insertions(+), 10 deletions(-)
+ 36 files changed, 714 insertions(+), 10 deletions(-)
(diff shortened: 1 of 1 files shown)


build/patches/Timezone-customization.patch
```diff
@@ -1287,7 +1287,8 @@ diff --git a/components/content_settings/renderer/content_settings_agent_impl.cc
 +  } else if (setting == CONTENT_SETTING_BLOCK) {
 +    // timezone random
 +    UErrorCode ec = U_ZERO_ERROR;
-+    int32_t rawOffset = ((rand() % 24) - 12) * 3600 * 1000;
++    std::srand(std::time(nullptr));
```
csagan5, line 1290: It is a bad practice to do this every single time a random number is used. Please use `base::RandInt` instead

csagan5 (changes requested):

build/patches/Timezone-customization.patch
uazo: fix in https://github.com/bromite/bromite/pull/830/commits/76992d379bc632413a9ec828457715737642b208

csagan5 (approved):

Issue #890, opened 2021-01-26
Proxy page does not load


chromer030:
### Bromite version

Version: `88.0.4324.95`
Arch: `arm64`
Android version: `7.1.2`
Device model: `Lenovo`

### Is this bug about the SystemWebView?
No

### Is the bug reproducible with latest version?
Yes

### Can the bug be reproduced with corresponding Chromium version?
It's not related to Chromium.

### Is the bug a crash?
No

### Describe the bug
1- Proxy page is broken and it is stuck on Loading... forever.

### Steps to reproduce the bug

Steps to reproduce the bug:
1 -Open chrome://proxy or from settings.

### Screenshots

![Screenshot_20210126-103237](https://user-images.githubusercontent.com/13361552/105812580-a8c9f000-5fc3-11eb-8221-c5a61ff39ef0.png)

csagan5:
I cannot reproduce this bug; please follow this documentation and see what errors are reported there in the development console: https://developers.google.com/web/tools/chrome-devtools/remote-debugging

uazo:
we need to add this line to `chrome/browser/resources/proxy_config.html`

`<script src="chrome://resources/js/assert.js"></script>`

as show here https://source.chromium.org/chromium/chromium/src/+/e2575bf8c6c0c0360862da29bcb4463f3118fe4f

here https://gist.github.com/uazo/94dde7027a559386d53d4a003548a58e the complete patch, it may be slightly different in some places as mentioned yesterday.

csagan5:
Fixed in `88.0.4324.141`.

PR #892, opened 2021-01-28
fix proxy and bookmark page

### LINKED ISSUE #891: Bookmarks don't work on 88.0.4324.95
### Bromite version
Version: `88.0.4324.95`
Arch: `arm`
Android version: `9.0`
Device model: `nx569j`

No.

Yes
.
I don't know.

### Is the bug a crash?
No,

### Describe the bug
Bookmarks don't work on 88.0.4324.95.

1. Open any bookmark.
2. Nothing happens - it just redirects to empty Mobile Bookmarks.

### Expected behavior
When you click on a bookmark, it opens a site in bookmark.

uazo:
fixes https://github.com/bromite/bromite/issues/891 https://github.com/bromite/bromite/issues/890

I deleted the useless `Revert-Remove-ChromeActivity-reference-from-BookmarkPage.patch`.
Tell me if it's okay


csagan5:
For the proxy I had already applied a similar fix; for the bookmarks, glad you could find a simpler way, will use your patch.

csagan5 (approved):

uazo:
perfect, if you want I can also fix the problem of the wrong thread and missing write permissions (dcheck failed, always present and visible only in debug). Maybe I also put the config of the destination folder in it...

csagan5:
Sure, but please wait for the new patches to be released.

csagan5:
@uazo I have pushed the patches, the release will come later as builds finish; you can already use them, if there are no build issues they will not change.

PR #898, opened 2021-01-30
add reading of kProxyReverseBypass from config

### LINKED ISSUE #884: "Reverse the meaning of bypass rules" broken
## Bromite version
Version: 87.0.4280.131
Arch: arm64
Android version: 11.0
Device model: Google Pixel 4a 5g (GD1YQ)

No

Yes

No

### Is the bug a crash?
No

### Describe the bug
In the chrome://proxy configuration page, "Reverse the meaning of bypass rules" unchecks itself upon clicking "Apply". This renders the setting unusable.

1. Go to Settings -> Privacy -> Proxy configuration
2. Enter any valid proxy configuration
3. Enter any valid "bypass rules"
4. Check "Reverse the meaning of bypass rules"
5. Click apply
6. Observe the previous checkbox un-checks itself

### Expected behavior
The check box should remain checked and saved properly.

uazo:
For #884

### DIFF
--- a/build/patches/Add-a-proxy-configuration-page.patch
+++ b/build/patches/Add-a-proxy-configuration-page.patch
@@ -23,11 +23,12 @@ Offer auto-complete for the proxy page URL.
  chrome/common/webui_url_constants.cc          |   4 +
  chrome/common/webui_url_constants.h           |   2 +
  .../core/browser/proxy_policy_handler.cc      |   2 +-
- .../proxy_config/proxy_config_dictionary.cc   |  22 +-
- .../proxy_config/proxy_config_dictionary.h    |   6 +-
+ .../pref_proxy_config_tracker_impl.cc         |   1 +
+ .../proxy_config/proxy_config_dictionary.cc   |  30 +-
+ .../proxy_config/proxy_config_dictionary.h    |   7 +-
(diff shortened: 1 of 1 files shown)


csagan5 (approved):

build/patches/Add-a-proxy-configuration-page.patch
```diff
@@ -1091,7 +1103,22 @@ diff --git a/components/proxy_config/proxy_config_dictionary.cc b/components/pro
...
  }  // namespace
  
-@@ -79,29 +81,30 @@ const base::Value& ProxyConfigDictionary::GetDictionary() const {
+@@ -72,6 +74,14 @@ bool ProxyConfigDictionary::HasBypassList() const {
+   return dict_.FindKey(kProxyBypassList);
+ }
+ 
++bool ProxyConfigDictionary::HasReverseBypass() const {
++  const base::Value* value = dict_.FindKey(kProxyReverseBypass);
++  if (!value || !value->is_bool()) {
```
csagan5, line 1112: When can it be not a boolean?

uazo: only if for some reason there is something that cannot be transformed into a boolean. `base::Value` can contain anything, even another dictionary. It is a check that is done in many places.

PR #901, opened 2021-02-02
Enable proxy config for the system network context

uazo:
The current patch comes from the implementation for ChromeOS which gets the proxy information from the local_state for the system network context.

This is not completely correct in Android, since the network context is never connected to any user profile and only gets the proxy information from the system.

With this change settings are moved from the Profile to LocalState, allowing the transparent use of the proxy also for SimpleURLLoaders (for example used for downloading Adblock filters and for favicons, which currently do not use any proxy even if set). Existing settings are automatically migrated on first startup.

### DIFF
--- a/build/patches/Add-a-proxy-configuration-page.patch
+++ b/build/patches/Add-a-proxy-configuration-page.patch
@@ -12,6 +12,9 @@ Offer auto-complete for the proxy page URL.
  .../privacy/settings/PrivacySettings.java     |   5 +-
  .../chrome_autocomplete_provider_client.cc    |   2 +
  chrome/browser/browser_resources.grd          |   6 +
+ chrome/browser/net/proxy_service_factory.cc   |  23 +-
+ chrome/browser/net/proxy_service_factory.h    |   3 +
+ chrome/browser/prefs/browser_prefs.cc         |   4 +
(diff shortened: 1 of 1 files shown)


csagan5:
Great job in spotting this problem; I assume it is tested to work also when using the default system settings.

csagan5 (approved):

csagan5:
@uazo even with this improvement DoH will not use the proxy, right?

uazo:
well, but in theory, isn't it the proxy that resolves the address?

csagan5:
The proxy uses DNS resolution to resolve the proxy address, then maybe after that it could use it. It seems complicated but in practice I verify that the DoH requests are not going through the proxy.

uazo:
> The proxy uses DNS resolution to resolve the proxy address

do you mean chromium?

but I think it is correct because otherwise it would be a counter circuit.
just tried, if you try with a non-existent domain, the proxy responds with "lookup failed".

csagan5:
> > The proxy uses DNS resolution to resolve the proxy address
> 
> do you mean chromium?

Also Bromite.

Yes, it is correct. I have documented the limitation here: https://github.com/bromite/bromite/wiki/DNS-over-HTTPS#limitations

Issue #836, opened 2020-12-05
consider enabling network isolation key partitioning features


thestinger:
Google appears to be testing some of these via field trials. The tricky part is figuring out which portions of this are ready to enable. It might all be ready already. These features are designed to split up everything other than cookies into partitions keyed based on the first party origin similar to Safari (and frame, going beyond what Safari does).

Extending it to cookies should be enough to fully partition site data, caches, connection pools, etc. when combined with the baseline feature. It's probably still not at the point that this is actually the case yet, but that's the goal for the feature.

Alternatively, could just wait a few months for this to roll out itself. However, some things have been getting delayed due to increased reluctance to make aggressive changes due to COVID-19 fallout. I don't think that's going to apply to this since it shouldn't break anything.

thestinger:
This is also called split cache in some places. It applies to a lot more than the cache though.

csagan5:
@thestinger have you already enabled this in Vanadium?

thestinger:
I want to figure out which subset of it they've started deploying first.

csagan5:
I can give a look tomorrow (+19h)

csagan5:
All the field trials can be found here: https://github.com/chromium/chromium/blob/87.0.4280.81/testing/variations/fieldtrial_testing_config.json

Look for `NetworkIsolation` there.

These are the upstream issues I have found:

* https://bugs.chromium.org/p/chromium/issues/detail?id=1018718
* https://bugs.chromium.org/p/chromium/issues/detail?id=963480

The features used by the field trials are:
* `PartitionConnectionsByNetworkIsolationKey`
* `PartitionHttpServerPropertiesByNetworkIsolationKey`
* `SplitHostCacheByNetworkIsolationKey`
* `AppendFrameOriginToNetworkIsolationKey`
* `SplitCacheByNetworkIsolationKey`
* `UseRegistrableDomainInNetworkIsolationKey`
* `PartitionSSLSessionsByNetworkIsolationKey`

See also: https://github.com/chromium/chromium/blob/87.0.4280.81/net/base/features.cc#L63

There are two field trials, `PartitionConnectionsByNetworkIsolationKey` and `SplitCacheByNetworkIsolationKey`; each field trial enables a certain set of features as you wrote; one of the two Chromium issues is marked as fixed, the other one seems "hanging", but there might be other private issues somewhere else (I hope not).

[...]

PR #905, opened 2021-02-04
Ability to see when adblock filters were last updated; button to manually update adblock filters

### LINKED ISSUE #811: Ability to see when adblock filters were last updated; button to manually update adblock filters

Yes, since it allows users to see if their adblock filters file is out of date. 

No

I would like to see a date when adblock filters we last updated and a button to manually update them if they are quite out of date to give users more control over their local filters. 

Can't think of any alternative.

uazo:
Patch for #811 

Added:
- InfoBar at startup (only if filters updated or in error)
- Displayed in the settings "Last Checked Date" and "Current Version"
- Possibility of manual update

### DIFF
--- a/build/patches/Bromite-AdBlockUpdaterService.patch
+++ b/build/patches/Bromite-AdBlockUpdaterService.patch
@@ -17,7 +17,7 @@ Fix RestoreForeignSessionTab by recreating the tab (issue #681)
  .../android/java/res/xml/main_preferences.xml |   5 +
  .../chrome/browser/app/ChromeActivity.java    |  18 +
  .../browser/settings/AdBlockEditor.java       |  92 +++++
- .../browser/settings/AdBlockPreferences.java  | 131 +++++++
+ .../browser/settings/AdBlockPreferences.java  | 130 +++++++
  .../chrome/browser/tabmodel/TabModelImpl.java |   2 +-
  chrome/app/generated_resources.grd            |  34 ++
  chrome/browser/after_startup_task_utils.cc    |   5 +
@@ -33,19 +33,19 @@ Fix RestoreForeignSessionTab by recreating the tab (issue #681)
  .../sessions/session_restore_android.cc       |   4 +-
  .../strings/android_chrome_strings.grd        |  14 +
  components/component_updater/BUILD.gn         |   6 +
- .../adblock_updater_service.cc                | 320 ++++++++++++++++++
+ .../adblock_updater_service.cc                | 321 ++++++++++++++++++
  .../adblock_updater_service.h                 | 118 +++++++
- .../download_filters_task.cc                  | 239 +++++++++++++
+ .../download_filters_task.cc                  | 237 +++++++++++++
  .../component_updater/download_filters_task.h | 129 +++++++
  ...ent_subresource_filter_throttle_manager.cc |  11 +
  .../content/browser/ruleset_service.cc        |  33 +-
  .../content/browser/ruleset_service.h         |   7 +-
  .../content/browser/ruleset_version.h         |   4 +
  .../browser/verified_ruleset_dealer.cc        |   3 +
- .../browser/subresource_filter_features.cc    | 113 +------
+ .../browser/subresource_filter_features.cc    | 113 +-----
  .../core/common/common_features.cc            |   2 +-
  .../navigation_throttle_runner.cc             |   5 -
- 36 files changed, 1444 insertions(+), 140 deletions(-)
+ 36 files changed, 1442 insertions(+), 140 deletions(-)
  create mode 100644 chrome/android/java/res/layout/adblock_editor.xml
  create mode 100644 chrome/android/java/res/xml/adblock_preferences.xml
  create mode 100644 chrome/android/java/src/org/chromium/chrome/browser/settings/AdBlockEditor.java
@@ -204,7 +204,7 @@ new file mode 100644
 +        app:allowDividerBelow="false" />
 +
 +    <org.chromium.components.browser_ui.settings.ButtonPreference
-+        android:key="adblock_startchek"
++        android:key="adblock_startcheck"
 +        android:title="@string/options_adblock_startcheck_label"/>
 +
 +    <org.chromium.components.browser_ui.settings.TextMessagePreference
@@ -384,7 +384,7 @@ diff --git a/chrome/android/java/src/org/chromium/chrome/browser/settings/AdBloc
 new file mode 100644
 --- /dev/null
 +++ b/chrome/android/java/src/org/chromium/chrome/browser/settings/AdBlockPreferences.java
-@@ -0,0 +1,131 @@
+@@ -0,0 +1,130 @@
 +// Copyright 2015 The Chromium Authors. All rights reserved.
 +// Use of this source code is governed by a BSD-style license that can be
 +// found in the LICENSE file.
@@ -423,7 +423,7 @@ new file mode 100644
 +    private static final String PREF_ADBLOCK_SWITCH = "adblock_switch";
 +    private static final String PREF_ADBLOCK_EDIT = "adblock_edit";
 +    private static final String PREF_ADBLOCK_INDEX_VERSION = "adblock_current_index_version";
-+    private static final String PREF_ADBLOCK_CHECK_NOW = "adblock_startchek";
++    private static final String PREF_ADBLOCK_CHECK_NOW = "adblock_startcheck";
 +    private static final String PREF_ADBLOCK_CURRENT_STATUS = "adblock_current_status";
 +
 +    private Preference mAdBlockEdit;
@@ -452,13 +452,12 @@ new file mode 100644
 +        mAdBlockEdit = findPreference(PREF_ADBLOCK_EDIT);
 +        updateCurrentAdBlockSettings();
 +
-+        AdBlockPreferences captured = this;
 +        ButtonPreference startUpdateButton =
 +            (ButtonPreference) findPreference(PREF_ADBLOCK_CHECK_NOW);
 +        startUpdateButton.setOnPreferenceClickListener(new OnPreferenceClickListener() {
 +            @Override
 +            public boolean onPreferenceClick(Preference preference) {
-+                CachedFeatureFlags.AdBlockStartCheck(captured);
++                CachedFeatureFlags.AdBlockStartCheck(AdBlockPreferences.this);
 +                return true;
 +            }
 +        });
@@ -978,7 +977,7 @@ diff --git a/components/component_updater/adblock_updater_service.cc b/component
 new file mode 100644
 --- /dev/null
 +++ b/components/component_updater/adblock_updater_service.cc
-@@ -0,0 +1,320 @@
+@@ -0,0 +1,321 @@
 +/*
 +    This file is part of Bromite.
 +
@@ -1114,14 +1113,15 @@ new file mode 100644
 +
 +  last_update_ = pref_service_->GetTime(kAdBlockLastCheckTime);
 +
-+  // Check if the request is too soon.
-+  if (is_foreground == false && !last_update_.is_null()) {
-+    base::TimeDelta delta =
-+        base::Time::Now() - last_update_;
-+    auto version = ruleset_service_->GetMostRecentlyIndexedVersion();
-+    if(!version.content_version.empty()) {
-+      if (is_updating_ || (delta < base::TimeDelta::FromSeconds(on_demand_check_delay))) {
-+        LOG(INFO) << "AdBlockUpdaterService: update not necessary.";
++  auto version = ruleset_service_->GetMostRecentlyIndexedVersion();
++  if (!version.content_version.empty()) {
++    // Check if the request is too soon.
++    if (!last_update_.is_null()) {
++      int deltaCheck = is_foreground == false ? next_check_delay : on_demand_check_delay;
++      base::TimeDelta delta = base::Time::Now() - last_update_;
++      if (is_updating_ || (delta < base::TimeDelta::FromSeconds(deltaCheck))) {
++        LOG(INFO) << "AdBlockUpdaterService: update delayed. Wait for "
++                  << (base::TimeDelta::FromSeconds(deltaCheck)-delta);
 +        return false;
 +      }
 +    }
@@ -1426,7 +1426,7 @@ diff --git a/components/component_updater/download_filters_task.cc b/components/
 new file mode 100644
 --- /dev/null
 +++ b/components/component_updater/download_filters_task.cc
-@@ -0,0 +1,239 @@
+@@ -0,0 +1,237 @@
 +/*
 +    This file is part of Bromite.
 +
@@ -1586,8 +1586,6 @@ new file mode 100644
 +
 +  final_url_ = final_url;
 +  response_code_ = response_head.headers ? response_head.headers->response_code() : -1;
-+
-+
 +
 +  if (!response_head.headers->GetLastModifiedValue(&last_modified_))
 +    LOG(WARNING) << "DownloadFiltersTask: fetched URL '" << final_url.spec()
--- a/build/patches/Bromite-AdBlockUpdaterService.patch
+++ b/build/patches/Bromite-AdBlockUpdaterService.patch
@@ -11,41 +11,41 @@ Fix RestoreForeignSessionTab by recreating the tab (issue #681)
 ---
  chrome/android/chrome_java_resources.gni      |   2 +
  chrome/android/chrome_java_sources.gni        |   2 +
- .../java/res/layout/adblock_editor.xml        |  67 +++++
+ .../java/res/layout/adblock_editor.xml        |  67 ++++
  chrome/android/java/res/values/values.xml     |   2 +
- .../java/res/xml/adblock_preferences.xml      |  25 ++
+ .../java/res/xml/adblock_preferences.xml      |  37 ++
  .../android/java/res/xml/main_preferences.xml |   5 +
- .../browser/settings/AdBlockEditor.java       |  92 +++++++
- .../browser/settings/AdBlockPreferences.java  |  61 +++++
+ .../chrome/browser/app/ChromeActivity.java    |  18 +
+ .../browser/settings/AdBlockEditor.java       |  92 +++++
+ .../browser/settings/AdBlockPreferences.java  | 131 +++++++
  .../chrome/browser/tabmodel/TabModelImpl.java |   2 +-
- chrome/app/generated_resources.grd            |  10 +
+ chrome/app/generated_resources.grd            |  34 ++
  chrome/browser/after_startup_task_utils.cc    |   5 +
  chrome/browser/browser_process.h              |   7 +
  chrome/browser/browser_process_impl.cc        |  29 ++
  chrome/browser/browser_process_impl.h         |   3 +
  chrome/browser/chrome_browser_main.cc         |   2 +
- .../browser/chrome_content_browser_client.cc  |  16 --
- .../flags/android/cached_feature_flags.cc     |  11 +
- .../browser/flags/CachedFeatureFlags.java     |  10 +
- .../net/system_network_context_manager.cc     |   4 +
+ .../browser/chrome_content_browser_client.cc  |  16 -
+ .../flags/android/cached_feature_flags.cc     |  62 ++++
+ .../flags/android/cached_feature_flags.h      |   2 +
+ .../browser/flags/CachedFeatureFlags.java     |  57 ++++
+ chrome/browser/prefs/browser_prefs.cc         |   1 +
  .../sessions/session_restore_android.cc       |   4 +-
  .../strings/android_chrome_strings.grd        |  14 +
- chrome/common/pref_names.cc                   |   3 +
- chrome/common/pref_names.h                    |   1 +
  components/component_updater/BUILD.gn         |   6 +
- .../adblock_updater_service.cc                | 248 ++++++++++++++++++
- .../adblock_updater_service.h                 | 100 +++++++
- .../download_filters_task.cc                  | 222 ++++++++++++++++
- .../component_updater/download_filters_task.h | 129 +++++++++
+ .../adblock_updater_service.cc                | 320 ++++++++++++++++++
+ .../adblock_updater_service.h                 | 118 +++++++
+ .../download_filters_task.cc                  | 239 +++++++++++++
+ .../component_updater/download_filters_task.h | 129 +++++++
  ...ent_subresource_filter_throttle_manager.cc |  11 +
- .../content/browser/ruleset_service.cc        |  33 ++-
+ .../content/browser/ruleset_service.cc        |  33 +-
  .../content/browser/ruleset_service.h         |   7 +-
  .../content/browser/ruleset_version.h         |   4 +
  .../browser/verified_ruleset_dealer.cc        |   3 +
- .../browser/subresource_filter_features.cc    | 113 +-------
+ .../browser/subresource_filter_features.cc    | 113 +------
  .../core/common/common_features.cc            |   2 +-
  .../navigation_throttle_runner.cc             |   5 -
- 36 files changed, 1120 insertions(+), 140 deletions(-)
+ 36 files changed, 1444 insertions(+), 140 deletions(-)
  create mode 100644 chrome/android/java/res/layout/adblock_editor.xml
  create mode 100644 chrome/android/java/res/xml/adblock_preferences.xml
  create mode 100644 chrome/android/java/src/org/chromium/chrome/browser/settings/AdBlockEditor.java
@@ -174,7 +174,7 @@ diff --git a/chrome/android/java/res/xml/adblock_preferences.xml b/chrome/androi
 new file mode 100644
 --- /dev/null
 +++ b/chrome/android/java/res/xml/adblock_preferences.xml
-@@ -0,0 +1,25 @@
+@@ -0,0 +1,37 @@
 +<?xml version="1.0" encoding="utf-8"?>
 +<!-- Copyright 2015 The Chromium Authors. All rights reserved.
 +     Use of this source code is governed by a BSD-style license that can be
@@ -199,6 +199,18 @@ new file mode 100644
 +        android:title="@string/options_adblock_edit_label"
 +        android:fragment="org.chromium.chrome.browser.settings.AdBlockEditor" />
 +
++    <org.chromium.components.browser_ui.settings.TextMessagePreference
++        android:key="adblock_current_index_version"
++        app:allowDividerBelow="false" />
++
++    <org.chromium.components.browser_ui.settings.ButtonPreference
++        android:key="adblock_startchek"
++        android:title="@string/options_adblock_startcheck_label"/>
++
++    <org.chromium.components.browser_ui.settings.TextMessagePreference
++        android:key="adblock_current_status"
++        app:allowDividerBelow="false" />
++
 +</PreferenceScreen>
 diff --git a/chrome/android/java/res/xml/main_preferences.xml b/chrome/android/java/res/xml/main_preferences.xml
 --- a/chrome/android/java/res/xml/main_preferences.xml
@@ -215,6 +227,62 @@ diff --git a/chrome/android/java/res/xml/main_preferences.xml b/chrome/android/j
      <Preference
          android:fragment="org.chromium.chrome.browser.safety_check.SafetyCheckSettingsFragment"
          android:key="safety_check"
+diff --git a/chrome/android/java/src/org/chromium/chrome/browser/app/ChromeActivity.java b/chrome/android/java/src/org/chromium/chrome/browser/app/ChromeActivity.java
+--- a/chrome/android/java/src/org/chromium/chrome/browser/app/ChromeActivity.java
++++ b/chrome/android/java/src/org/chromium/chrome/browser/app/ChromeActivity.java
+@@ -59,6 +59,8 @@ import org.chromium.chrome.browser.ChromeActivitySessionTracker;
+ import org.chromium.chrome.browser.ChromeApplication;
+ import org.chromium.chrome.browser.ChromeWindow;
+ import org.chromium.chrome.browser.DeferredStartupHandler;
++import org.chromium.chrome.browser.flags.CachedFeatureFlags;
++import org.chromium.chrome.browser.flags.CachedFeatureFlags.AdBlockCallback;
+ import org.chromium.chrome.browser.IntentHandler;
+ import org.chromium.chrome.browser.IntentHandler.IntentHandlerDelegate;
+ import org.chromium.chrome.browser.IntentHandler.TabOpenType;
+@@ -106,6 +108,7 @@ import org.chromium.chrome.browser.gsa.ContextReporter;
+ import org.chromium.chrome.browser.gsa.GSAAccountChangeListener;
+ import org.chromium.chrome.browser.gsa.GSAState;
+ import org.chromium.chrome.browser.history.HistoryManagerUtils;
++import org.chromium.chrome.browser.infobar.InfoBarIdentifier;
+ import org.chromium.chrome.browser.init.AsyncInitializationActivity;
+ import org.chromium.chrome.browser.init.ProcessInitializationHandler;
+ import org.chromium.chrome.browser.init.StartupTabPreloader;
+@@ -133,6 +136,7 @@ import org.chromium.chrome.browser.printing.TabPrinter;
+ import org.chromium.chrome.browser.profiles.Profile;
+ import org.chromium.chrome.browser.settings.SettingsLauncher;
+ import org.chromium.chrome.browser.settings.SettingsLauncherImpl;
++import org.chromium.chrome.browser.settings.AdBlockPreferences;
+ import org.chromium.chrome.browser.share.ShareDelegate;
+ import org.chromium.chrome.browser.share.ShareDelegateImpl;
+ import org.chromium.chrome.browser.tab.AccessibilityVisibilityHandler;
+@@ -159,6 +163,7 @@ import org.chromium.chrome.browser.ui.TabObscuringHandler;
+ import org.chromium.chrome.browser.ui.appmenu.AppMenuBlocker;
+ import org.chromium.chrome.browser.ui.appmenu.AppMenuDelegate;
+ import org.chromium.chrome.browser.ui.appmenu.AppMenuPropertiesDelegate;
++import org.chromium.chrome.browser.ui.messages.infobar.SimpleConfirmInfoBarBuilder;
+ import org.chromium.chrome.browser.ui.messages.snackbar.SnackbarManager;
+ import org.chromium.chrome.browser.ui.messages.snackbar.SnackbarManager.SnackbarManageable;
+ import org.chromium.chrome.browser.ui.messages.snackbar.SnackbarManagerProvider;
+@@ -1018,6 +1023,19 @@ public abstract class ChromeActivity<C extends ChromeActivityComponent>
+      */
+     @CallSuper
+     protected void initDeferredStartupForActivity() {
++        DeferredStartupHandler.getInstance().addDeferredTask(() -> {
++            CachedFeatureFlags.AdBlockRegisterCallback((int result, int error) -> {
++                if (result == /*ADBLOCK_UPDATED*/ 4 || error == /*DOWNLOAD_ERROR*/ 4) {
++                    Tab tab = ChromeActivity.this.getActivityTabProvider().get();
++                    if (tab == null) return;
++
++                    SimpleConfirmInfoBarBuilder.create(tab.getWebContents(),
++                        InfoBarIdentifier.WINDOW_ERROR_INFOBAR_DELEGATE_ANDROID,
++                        AdBlockPreferences.GetAdBlockMessage(result, error), true);
++                }
++            });
++        });
++
+         DeferredStartupHandler.getInstance().addDeferredTask(() -> {
+             if (isActivityFinishingOrDestroyed()) return;
+             UpdateInfoBarController.createInstance(ChromeActivity.this);
 diff --git a/chrome/android/java/src/org/chromium/chrome/browser/settings/AdBlockEditor.java b/chrome/android/java/src/org/chromium/chrome/browser/settings/AdBlockEditor.java
 new file mode 100644
 --- /dev/null
@@ -316,7 +384,7 @@ diff --git a/chrome/android/java/src/org/chromium/chrome/browser/settings/AdBloc
 new file mode 100644
 --- /dev/null
 +++ b/chrome/android/java/src/org/chromium/chrome/browser/settings/AdBlockPreferences.java
-@@ -0,0 +1,61 @@
+@@ -0,0 +1,131 @@
 +// Copyright 2015 The Chromium Authors. All rights reserved.
 +// Use of this source code is governed by a BSD-style license that can be
 +// found in the LICENSE file.
@@ -324,28 +392,42 @@ new file mode 100644
 +package org.chromium.chrome.browser.settings;
 +
 +import android.os.Bundle;
++import android.content.Context;
 +import androidx.preference.Preference;
 +import androidx.preference.PreferenceFragmentCompat;
 +import org.chromium.components.browser_ui.settings.ChromeSwitchPreference;
 +
++import org.chromium.base.Log;
++import org.chromium.base.ContextUtils;
 +import org.chromium.components.embedder_support.browser_context.BrowserContextHandle;
 +import org.chromium.components.browser_ui.site_settings.SiteSettingsPreferenceFragment;
 +import org.chromium.components.browser_ui.site_settings.WebsitePreferenceBridge;
++import org.chromium.components.browser_ui.settings.TextMessagePreference;
++import org.chromium.components.browser_ui.settings.ButtonPreference;
 +import org.chromium.components.content_settings.ContentSettingsType;
 +import org.chromium.components.browser_ui.settings.SettingsUtils;
 +import org.chromium.chrome.browser.flags.CachedFeatureFlags;
++import org.chromium.chrome.browser.flags.CachedFeatureFlags.AdBlockCallback;
 +import androidx.annotation.VisibleForTesting;
++import androidx.preference.Preference.OnPreferenceClickListener;
 +import org.chromium.chrome.R;
++
++import java.text.DateFormat;
 +
 +/**
 + * Fragment that allows the user to configure AdBlock related preferences.
 + */
-+public class AdBlockPreferences extends SiteSettingsPreferenceFragment {
++public class AdBlockPreferences extends SiteSettingsPreferenceFragment
++                                implements AdBlockCallback {
 +    @VisibleForTesting
-+    public static final String PREF_ADBLOCK_SWITCH = "adblock_switch";
++    private static final String PREF_ADBLOCK_SWITCH = "adblock_switch";
 +    private static final String PREF_ADBLOCK_EDIT = "adblock_edit";
++    private static final String PREF_ADBLOCK_INDEX_VERSION = "adblock_current_index_version";
++    private static final String PREF_ADBLOCK_CHECK_NOW = "adblock_startchek";
++    private static final String PREF_ADBLOCK_CURRENT_STATUS = "adblock_current_status";
 +
 +    private Preference mAdBlockEdit;
++    private TextMessagePreference currentIndexVersion;
 +
 +    @Override
 +    public void onCreatePreferences(Bundle savedInstanceState, String rootKey) {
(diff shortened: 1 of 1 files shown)


build/patches/Bromite-AdBlockUpdaterService.patch
```diff
@@ -199,6 +199,18 @@ new file mode 100644
 +        android:title="@string/options_adblock_edit_label"
 +        android:fragment="org.chromium.chrome.browser.settings.AdBlockEditor" />
 +
++    <org.chromium.components.browser_ui.settings.TextMessagePreference
++        android:key="adblock_current_index_version"
++        app:allowDividerBelow="false" />
++
++    <org.chromium.components.browser_ui.settings.ButtonPreference
++        android:key="adblock_startchek"
```
csagan5, line 207:
```suggestion
+        android:key="adblock_startcheck"
```

build/patches/Bromite-AdBlockUpdaterService.patch
```diff
@@ -316,36 +384,50 @@ diff --git a/chrome/android/java/src/org/chromium/chrome/browser/settings/AdBloc
...
 + */
-+public class AdBlockPreferences extends SiteSettingsPreferenceFragment {
++public class AdBlockPreferences extends SiteSettingsPreferenceFragment
++                                implements AdBlockCallback {
 +    @VisibleForTesting
-+    public static final String PREF_ADBLOCK_SWITCH = "adblock_switch";
++    private static final String PREF_ADBLOCK_SWITCH = "adblock_switch";
 +    private static final String PREF_ADBLOCK_EDIT = "adblock_edit";
++    private static final String PREF_ADBLOCK_INDEX_VERSION = "adblock_current_index_version";
++    private static final String PREF_ADBLOCK_CHECK_NOW = "adblock_startchek";
```
csagan5, line 426:
```suggestion
+    private static final String PREF_ADBLOCK_CHECK_NOW = "adblock_startcheck";
```

build/patches/Bromite-AdBlockUpdaterService.patch
```diff
@@ -1224,19 +1557,27 @@ new file mode 100644
 +    return;
 +  }
 +
-+  // ignoring 'headers' as 'Last-Modified' has already been picked up by OnResponseStarted
-+  const base::TimeDelta dt =
-+          last_modified_ - min_last_modified_;
++  int net_error = simple_url_loader_->NetError();
++  simple_url_loader_.reset();
```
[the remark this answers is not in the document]
uazo: ah, one more thing: do you ever know that I am in doubt if quic is really disabled? When digging, I've seen here and there some #if **no android** when ask if disabled.

I don't know exactly the protocol (I will study it) but from what I understood the problem was in the session cache, which would allow the host to track you if it is not deleted, because it would keep the local cryptographic key, obviously different for client and therefore with possibility tracking. I am wrong? You know something about this?

between the flags there is a way to disable it, and I thought it was applied, but with those #if they make me doubt.

build/patches/Bromite-AdBlockUpdaterService.patch
uazo: fixed in https://github.com/bromite/bromite/pull/905/commits/78a9c7eb85cadf9ba849f400a646e8be3999b629

build/patches/Bromite-AdBlockUpdaterService.patch
uazo: fixed in https://github.com/bromite/bromite/pull/905/commits/78a9c7eb85cadf9ba849f400a646e8be3999b629

build/patches/Bromite-AdBlockUpdaterService.patch
```diff
@@ -364,18 +446,74 @@ new file mode 100644
...
-+        updateCurrentAdBlockUrl();
++        updateCurrentAdBlockSettings();
++
++        AdBlockPreferences captured = this;
++        ButtonPreference startUpdateButton =
++            (ButtonPreference) findPreference(PREF_ADBLOCK_CHECK_NOW);
++        startUpdateButton.setOnPreferenceClickListener(new OnPreferenceClickListener() {
++            @Override
++            public boolean onPreferenceClick(Preference preference) {
++                CachedFeatureFlags.AdBlockStartCheck(captured);
```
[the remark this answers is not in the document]
uazo: fixed in https://github.com/bromite/bromite/pull/905/commits/78a9c7eb85cadf9ba849f400a646e8be3999b629

build/patches/Bromite-AdBlockUpdaterService.patch
```diff
@@ -845,13 +1112,19 @@ new file mode 100644
...
++  if (is_foreground == false && !last_update_.is_null()) {
 +    base::TimeDelta delta =
-+        base::TimeTicks::Now() - last_update_;
-+    if (is_updating_ || (delta < base::TimeDelta::FromSeconds(on_demand_check_delay)))
-+      LOG(INFO) << "AdBlockUpdaterService: update not necessary.";
-+      return false;
++        base::Time::Now() - last_update_;
++    auto version = ruleset_service_->GetMostRecentlyIndexedVersion();
++    if(!version.content_version.empty()) {
++      if (is_updating_ || (delta < base::TimeDelta::FromSeconds(on_demand_check_delay))) {
```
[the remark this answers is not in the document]
uazo: fixed in https://github.com/bromite/bromite/pull/905/commits/78a9c7eb85cadf9ba849f400a646e8be3999b629

build/patches/Bromite-AdBlockUpdaterService.patch
```diff
@@ -775,20 +1018,40 @@ new file mode 100644
...
++// Last check time
++const char kAdBlockLastCheckTime[] = "adblock.last_check_time";
++
++// Last check time (only with ok)
++const char kAdBlockLastCheckTimeOk[] = "adblock.last_check_time_ok";
++
 +// all constants express seconds
 +// these could be made configurable
 +const int initial_check_delay = 5,
 +      next_check_delay = 60*60*24*7, // 1 week
```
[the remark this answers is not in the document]
csagan5: It is currently 2 weeks; but why do we need to save the preference?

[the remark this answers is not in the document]
csagan5: Sure, we can.

build/patches/Bromite-AdBlockUpdaterService.patch
```diff
@@ -1060,27 +1379,41 @@ new file mode 100644
 +  // Will schedule automatic updates, run in background.
 +  void Start();
 +
++  std::string GetAdBlockFiltersURL();
++  void SetAdBlockFiltersURL(const std::string url);
++  std::string GetMostRecentIndexedVersion();
++
++  // return date/time (in millis) of last successfully request
++  long GetLastOkUpdateUs();
```
csagan5, line 1387: If it is milliseconds then it should not be `Us`.

build/patches/Bromite-AdBlockUpdaterService.patch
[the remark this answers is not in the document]
csagan5: > I have repeatedly tried in different ways to force the socket to close but I have not succeeded, it remains open until the android recents open. I think it's standard chromium behavior, but I don't like it, I think I open an issue to remind me to check.

It probably goes back in a connection pool and is closed automatically when not needed anymore; why do you think it is a problem? For privacy reasons?



> I am in doubt if quic is really disabled

It is not disabled.

> I don't know exactly the protocol (I will study it) but from what I understood the problem was in the session cache, which would allow the host to track you if it is not deleted, because it would keep the local cryptographic key, obviously different for client and therefore with possibility tracking. I am wrong? You know something about this?

It was disabled for a while in Bromite, then Chromium fixed/mitigated the problem you are mentioning and the patch was dropped; see https://github.com/bromite/bromite/blob/master/CHANGELOG.md#830410346

When I dropped the patch I checked about the remaining concerns and there were none based on facts; also, QUIC is going to be replaced by HTTP/3 and Chromium implementation is aligned with that as far as I know.

If you find that there are still concerns we can look into this again, but there was no privacy concern left back then.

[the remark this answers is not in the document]
uazo: > It probably goes back in a connection pool and is closed automatically

certainly, but it seems too aggressive

> For privacy 

from what I have seen, the socket stays open for my entire browsing session and there is nothing that makes it close except by opening the recents, so theoretically github (via adblock updater) knows when I use bromite.
This particular situation is resolved with my changes, because, at the very least, now github is notified every 7 days (as much as entered in `next_check_delay`), and if the user really bothers this too, he can set up his own filter mirror

however it is not a problem of this patch, even the favicons sockets (opened by the new tab page, which fortunately I don't use) remain open.

The anomaly is that by clearing the cache (therefore, in the belief to erase all traces), chromium requests immediately the favicons for all sites in newtabs, reopening the sockets and never closing them: so, theoretically, all the sites that I would have in the newtab know when I delete the history! yes, they don't send any cookies (and I hope not even the etag), but still I'm sending some sort of ping, and would be even worse if it were true the quic issue.

All this, however, it's my thought, should be better documented and verified, but it should make us think why in anonymous browsing favicons are never loaded

> It is not disabled.

i disabled it via chrome://flags

> If you find that there are still concerns we can look into this again

i think i will check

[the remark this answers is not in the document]
csagan5: > the socket stays open for my entire browsing session and there is nothing that makes it close except by opening the recents

What are the "recents"?

> github is notified every 7 days

GitHub is not notified, the collection is passive; HTTP requests are logged and they surely will analyse traffic in bulk to see usage patterns, cost analysis and to provide data to agencies during investigations, as needed.

See https://docs.github.com/en/github/site-policy/github-privacy-statement#usage-information; I could not find information about for how long they keep logs though.

> reopening the sockets and never closing them

This cannot be; if it is true, you can report it as a bug to Chromium. I am afraid you are misinterpreting the TCP connection states. What is it exactly that you see? Do you see any TCP/UDP packets going back and forth during this time when sockets are never closed?

> it should make us think why in anonymous browsing favicons are never loaded

Source?

> I hope not even the etag

With the recent enabling in Bromite of the network isolation features (see https://github.com/bromite/bromite/issues/836) the concerns about shared connections are mitigated.

build/patches/Bromite-AdBlockUpdaterService.patch
```diff
@@ -775,20 +1018,40 @@ new file mode 100644
...
++      subresource_filter::RulesetService* ruleset_service,
++      PrefService* pref_service)
++ : ruleset_service_(ruleset_service),
++   shared_url_network_factory_(shared_url_network_factory),
++   scheduler_(std::move(scheduler)) {
 +  DCHECK(ruleset_service);
-+
-+  filters_url_ = filters_url;
++  pref_service_ = pref_service;
++  ruleset_service_->SetRulesetPublishedCallbackForTesting(
```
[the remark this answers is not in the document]
csagan5: We should not use the testing methods; clone it without the `ForTesting` suffix?

build/patches/Bromite-AdBlockUpdaterService.patch
```diff
@@ -1167,7 +1498,6 @@ new file mode 100644
 +
 +  auto resource_request = std::make_unique<network::ResourceRequest>();
 +  resource_request->url = filters_url_;
-+  resource_request->credentials_mode = network::mojom::CredentialsMode::kOmit;
```
csagan5, line 1170: This is not necessary?

build/patches/Bromite-AdBlockUpdaterService.patch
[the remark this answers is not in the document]
uazo: yes, I thought the same thing too

build/patches/Bromite-AdBlockUpdaterService.patch
uazo: duplicate: it's on line 1502

build/patches/Bromite-AdBlockUpdaterService.patch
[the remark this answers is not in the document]
uazo: > What are the "recents"?

I mean https://developer.android.com/guide/components/activities/recents

> GitHub is not notified, the collection is passive

yes, sure, sorry, "notified" is wrong

> Do you see any TCP/UDP packets going back and forth during this time when sockets are never closed?

no, only tcp port open, no traffic

> What is it exactly that you see?

maybe I do a report, I can always be wrong

> Source?

ok, sorry, since the ui doesn't show the icons, i guess they don't load. I haven't checked the code yet.

> With the recent enabling in Bromite of the network isolation features (see #836) the concerns about shared connections are mitigated. As for ETag, you can easily verify this.

perfect, I'll check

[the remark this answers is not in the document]
csagan5: > no, only tcp port open, no traffic
> maybe I do a report, I can always be wrong

Are keep-alive packets being sent? It might keep the connections open to avoid re-doing the TLS handshake.

build/patches/Bromite-AdBlockUpdaterService.patch
```diff
@@ -364,18 +446,73 @@ new file mode 100644
...
++        currentIndexVersion.setSummary(
++            CachedFeatureFlags.getAdBlockMostRecentIndexedVersion() +
++            "\n" + getString(R.string.adblock_last_checked_text) +
++            " "  + df.format(CachedFeatureFlags.getAdBlockLastUpdateUs()));
++    }
++
++    public static String GetAdBlockMessage(int event, int error) {
++        Context applicationContext = ContextUtils.getApplicationContext();
++        // see in components/component_updater/download_filters_task.h
++        // EVENT:
```
csagan5, line 479: Wouldn't it be better to specify these as constants instead?

csagan5:
It does not currently build:
```
../../chrome/browser/flags/android/cached_feature_flags.cc:79:13: error: functions that differ only in their return type cannot be overloaded
static long JNI_CachedFeatureFlags_GetAdBlockLastOkUpdateUs(JNIEnv* env) {
       ~~~~ ^
gen/chrome/browser/flags/jni_headers/CachedFeatureFlags_jni.h:87:14: note: previous declaration is here
static jlong JNI_CachedFeatureFlags_GetAdBlockLastOkUpdateUs(JNIEnv* env);
       ~~~~~ ^
../../chrome/browser/flags/android/cached_feature_flags.cc:84:13: error: functions that differ only in their return type cannot be overloaded
static long JNI_CachedFeatureFlags_GetAdBlockLastUpdateUs(JNIEnv* env) {
       ~~~~ ^
gen/chrome/browser/flags/jni_headers/CachedFeatureFlags_jni.h:79:14: note: previous declaration is here
static jlong JNI_CachedFeatureFlags_GetAdBlockLastUpdateUs(JNIEnv* env);
       ~~~~~ ^
2 errors generated.
```

build/patches/Bromite-AdBlockUpdaterService.patch
uazo: yes, I change it, I've already seen how it's done.
for CachedFeatureFlags.java I need to create a new patch specific file

uazo:
> 2 errors generated.

do you know i noticed it only compiling for x86? please change long in jlong

build/patches/Bromite-AdBlockUpdaterService.patch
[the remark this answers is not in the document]
uazo: > Are keep-alive packets being sent? It might keep the connections open to avoid re-doing the TLS handshake.

is the first thing I thought, but `simple_url_loader` has no option to control that. 
I have the trace of the net, I have to check. but above all I have to check because opening the recents the sockets are closed, maybe somewhere i find a timeout to configure.

build/patches/Bromite-AdBlockUpdaterService.patch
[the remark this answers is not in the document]
uazo: added in https://github.com/bromite/bromite/pull/905/commits/95ab44c7afab8259a0ba50108fad5e846d609eee

build/patches/Bromite-AdBlockUpdaterService.patch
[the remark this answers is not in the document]
uazo: fixed in https://github.com/bromite/bromite/pull/905/commits/95ab44c7afab8259a0ba50108fad5e846d609eee

build/patches/Bromite-AdBlockUpdaterService.patch
uazo: fixed in https://github.com/bromite/bromite/pull/905/commits/95ab44c7afab8259a0ba50108fad5e846d609eee

build/patches/Bromite-AdBlockUpdaterService.patch
uazo: fixed in https://github.com/bromite/bromite/pull/905/commits/95ab44c7afab8259a0ba50108fad5e846d609eee

build/patches/Bromite-AdBlockUpdaterService.patch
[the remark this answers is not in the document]
csagan5: If there are no TCP packets transferred to or received from the host, then it is not an issue; you can verify this with Wireshark.

build/patches/Bromite-AdBlockUpdaterService.patch
```diff
@@ -199,6 +205,24 @@ new file mode 100644
 +        android:title="@string/options_adblock_edit_label"
 +        android:fragment="org.chromium.chrome.browser.settings.AdBlockEditor" />
 +
++    <org.chromium.components.browser_ui.settings.SpinnerPreference
++        android:key="adblock_period_spinner"
```
csagan5, line 209: Can you please change all the occurrencies of this to `frequency`?

build/patches/Bromite-AdBlockUpdaterService.patch
```diff
@@ -775,20 +1245,43 @@ new file mode 100644
...
 +// all constants express seconds
 +// these could be made configurable
 +const int initial_check_delay = 5,
-+      next_check_delay = 60*60*24*7, // 1 week
 +      on_demand_check_delay = 60; // minimum 1 minute between each on-demand check
-+
-+AdBlockUpdaterService::AdBlockUpdaterService(scoped_refptr<network::SharedURLLoaderFactory> shared_url_network_factory, std::unique_ptr<component_updater::UpdateScheduler> scheduler,
-+   subresource_filter::RulesetService* ruleset_service, std::string filters_url)
-+ : ruleset_service_(ruleset_service), shared_url_network_factory_(shared_url_network_factory), scheduler_(std::move(scheduler)) {
++int next_check_delay = 60*60*24*7; // 1 week
```
csagan5, line 1270: If this is not a constant then it should be on the `AdBlockUpdaterService` class (`next_check_delay_`)

build/patches/Bromite-AdBlockUpdaterService.patch
```diff
@@ -820,20 +1316,32 @@ new file mode 100644
...
++      base::TimeDelta::FromSeconds(delay),
 +      base::TimeDelta::FromSeconds(next_check_delay),
 +      base::Bind(&AdBlockUpdaterService::OnDemandScheduledUpdate,
 +                 base::Unretained(this)), base::DoNothing());
 +}
 +
-+void AdBlockUpdaterService::OnDemandScheduledUpdate(component_updater::UpdateScheduler::OnFinishedCallback on_finished) {
++void AdBlockUpdaterService::OnDemandScheduledUpdate(
++        component_updater::UpdateScheduler::OnFinishedCallback on_finished) {
 +  //TODO: call on_finished
```
csagan5, line 1345: Is this a new TODO?

build/patches/Bromite-AdBlockUpdaterService.patch
```diff
@@ -983,12 +1508,54 @@ new file mode 100644
...
++void AdBlockUpdaterService::SetAdBlockFiltersURL(const std::string url) {
++  pref_service_->SetString(kAdBlockFiltersURL, url);
++}
++
++std::string AdBlockUpdaterService::GetMostRecentIndexedVersion() {
++  auto version = ruleset_service_->GetMostRecentlyIndexedVersion();
++  return version.content_version;
++}
++
++long AdBlockUpdaterService::GetLastUpdate() {
```
csagan5, line 1524: Perhaps we can document this function with a comment? Saying that it returns a timestamp in UNIX milliseconds

build/patches/Bromite-AdBlockUpdaterService.patch
```diff
@@ -1060,27 +1631,47 @@ new file mode 100644
...
++
 +  // To be called for an user-triggered update.
 +  // Will not result in an actual update if the last update was too recently triggered.
 +  bool OnDemandUpdate(Callback on_finished);
 +
++  static void RegisterPrefs(PrefRegistrySimple* registry);
++
 + private:
-+  void NotifyObservers(Event event);
++  void ReStart(int delay);
```
csagan5, line 1655: Can we just call it `Start`? Since it is not doing any logic specific to restarting

csagan5 (changes requested):
It is a solid patch/feature; I left some request for changes and will attempt to build it again now.

build/patches/Bromite-AdBlockUpdaterService.patch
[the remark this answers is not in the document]
uazo: > then it is not an issue

it depends, in my opinion maintaining an active tcp session is an issue, with or without traffic, especially if it remains open until the browser is closed.
however, let me check with chromium net-internals as soon as I have some time.

build/patches/Bromite-AdBlockUpdaterService.patch
uazo: fixed in https://github.com/bromite/bromite/pull/905/commits/b11ab2588e787d6349664120c0ccdb7827cab201

build/patches/Bromite-AdBlockUpdaterService.patch
uazo: fixed in https://github.com/bromite/bromite/pull/905/commits/b11ab2588e787d6349664120c0ccdb7827cab201

build/patches/Bromite-AdBlockUpdaterService.patch
uazo: fixed in https://github.com/bromite/bromite/pull/905/commits/b11ab2588e787d6349664120c0ccdb7827cab201

build/patches/Bromite-AdBlockUpdaterService.patch
uazo: there is already in `.h`

build/patches/Bromite-AdBlockUpdaterService.patch
uazo: actually it is needed, because it is used in case the user changes the value in the settings

build/patches/Bromite-AdBlockUpdaterService.patch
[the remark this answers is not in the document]
uazo: excuse my ignorance, but is it me or you who have to "resolve" the conversation? there is always something to learn ...

build/patches/Bromite-AdBlockUpdaterService.patch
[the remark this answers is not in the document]
csagan5: Again: without traffic, there is no issue. You might be confused by kernel optimisations that leave the port "pre-opened" for the next connection.

build/patches/Bromite-AdBlockUpdaterService.patch
csagan5: There is no code specific to "restarting" vs "starting"

build/patches/Bromite-AdBlockUpdaterService.patch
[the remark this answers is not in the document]
csagan5: Etiquette is probably to let who started the conversation resolve it.

build/patches/Bromite-AdBlockUpdaterService.patch
```diff
@@ -215,6 +227,62 @@ diff --git a/chrome/android/java/res/xml/main_preferences.xml b/chrome/android/j
...
+     @CallSuper
+     protected void initDeferredStartupForActivity() {
++        DeferredStartupHandler.getInstance().addDeferredTask(() -> {
++            CachedFeatureFlags.AdBlockRegisterCallback((int result, int error) -> {
++                if (result == /*ADBLOCK_UPDATED*/ 4 || error == /*DOWNLOAD_ERROR*/ 4) {
++                    Tab tab = ChromeActivity.this.getActivityTabProvider().get();
++                    if (tab == null) return;
++
++                    SimpleConfirmInfoBarBuilder.create(tab.getWebContents(),
++                        InfoBarIdentifier.WINDOW_ERROR_INFOBAR_DELEGATE_ANDROID,
```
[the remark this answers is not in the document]
csagan5: What is the concern?

build/patches/Bromite-AdBlockUpdaterService.patch
```diff
@@ -1589,7 +2201,18 @@ diff --git a/components/subresource_filter/content/browser/ruleset_service.cc b/
 diff --git a/components/subresource_filter/content/browser/ruleset_service.h b/components/subresource_filter/content/browser/ruleset_service.h
 --- a/components/subresource_filter/content/browser/ruleset_service.h
 +++ b/components/subresource_filter/content/browser/ruleset_service.h
-@@ -182,7 +182,7 @@ class RulesetService : public base::SupportsWeakPtr<RulesetService> {
+@@ -171,6 +171,10 @@ class RulesetService : public base::SupportsWeakPtr<RulesetService> {
+     publisher_->SetRulesetPublishedCallbackForTesting(std::move(callback));
+   }
+ 
++  void SetRulesetPublishedCallback(base::OnceClosure callback) {
++    publisher_->SetRulesetPublishedCallbackForTesting(std::move(callback));
```
csagan5, line 2209: I meant to completely avoid calling this method and have our own, doing the identical job.

build/patches/Bromite-AdBlockUpdaterService.patch
uazo: maybe I don't understand, it's here

```
void AdBlockUpdaterService::SetAdBlockUpdateTimePeriod(int days) {
  pref_service_->SetInteger(kAdBlockFiltersCheckPeriod, days);
  ReStart(0);
}
```

csagan5: Inside the body of `ReStart` there is no logic like:
```
if (already_started) {
     // ...
}
```

This is what I mean

build/patches/Bromite-AdBlockUpdaterService.patch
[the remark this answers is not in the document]
uazo: for a clean patch i should have created another `InfoBarIdentifier` but then i saw that it really doesn't matter, it's more for histograms, i wanted to know what you thought.
adding it means modifying other files for nothing.

build/patches/Bromite-AdBlockUpdaterService.patch
uazo: fixed in https://github.com/bromite/bromite/pull/905/commits/51b8c9b04f86cb1cb56e30abc152bfd539780bf9

build/patches/Bromite-AdBlockUpdaterService.patch
uazo: fixed in https://github.com/bromite/bromite/pull/905/commits/51b8c9b04f86cb1cb56e30abc152bfd539780bf9

build/patches/Bromite-AdBlockUpdaterService.patch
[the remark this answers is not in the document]
csagan5: I think it is fine like this.

build/patches/Bromite-AdBlockUpdaterService.patch
csagan5: It is fixed because now `on_finished` is called correctly?

build/patches/Bromite-AdBlockUpdaterService.patch
```diff
@@ -364,18 +472,109 @@ new file mode 100644
...
++                getString(R.string.options_adblock_frequency_day)));
++        options.add(new TimeFrequencySpinnerOption(4,
++                getString(R.string.options_adblock_frequency_days)));
++        options.add(new TimeFrequencySpinnerOption(7,
++                getString(R.string.options_adblock_frequency_days)));
++        options.add(new TimeFrequencySpinnerOption(10,
++                getString(R.string.options_adblock_frequency_days)));
++        options.add(new TimeFrequencySpinnerOption(30,
++                getString(R.string.options_adblock_frequency_days)));
++        return options.toArray(new TimeFrequencySpinnerOption[0]);
```
csagan5, line 558: I would suggest these options:
* Never
* 7 days (default)
* 15 days
* 30 days

In fact the filters are updated every 2 weeks, that is why the default is 7 days.

build/patches/Bromite-AdBlockUpdaterService.patch
```diff
@@ -405,6 +604,39 @@ diff --git a/chrome/app/generated_resources.grd b/chrome/app/generated_resources
...
++        Days
++      </message>
++      <message name="IDS_OPTIONS_ADBLOCK_STARTCHECK_LABEL" desc="The title of the Ad Blocking button to check update" formatter_data="android_java">
++        Check Now
++      </message>
++      <message name="IDS_OPTIONS_ADBLOCK_EVENT_1" desc="The text for ADBLOCK_CHECKING_FOR_UPDATES" formatter_data="android_java">
++        Adblock filter: checking for update
++      </message>
++      <message name="IDS_OPTIONS_ADBLOCK_EVENT_3" desc="The text for ADBLOCK_UPDATE_DOWNLOADING" formatter_data="android_java">
++        Adblock filter: downloading update...
```
csagan5, line 629:
```suggestion
+        Downloading new filters...
```

build/patches/Bromite-AdBlockUpdaterService.patch
```diff
@@ -405,6 +604,39 @@ diff --git a/chrome/app/generated_resources.grd b/chrome/app/generated_resources
...
++      </message>
++      <message name="IDS_OPTIONS_ADBLOCK_EVENT_4" desc="The text for ADBLOCK_UPDATED" formatter_data="android_java">
++        Adblock filter: update successfully installed
++      </message>
++      <message name="IDS_OPTIONS_ADBLOCK_ERROR_3" desc="The text for UPDATE_NOT_NEEDED" formatter_data="android_java">
++        Adblock filter: update not needed
++      </message>
++      <message name="IDS_OPTIONS_ADBLOCK_ERROR_4" desc="The text for DOWNLOAD_ERROR" formatter_data="android_java">
++        Adblock filter: download error
```
csagan5, line 638:
```suggestion
+        Download error
```

build/patches/Bromite-AdBlockUpdaterService.patch
```diff
@@ -405,6 +604,39 @@ diff --git a/chrome/app/generated_resources.grd b/chrome/app/generated_resources
...
++        Day
++      </message>
++      <message name="IDS_OPTIONS_ADBLOCK_FREQUENCY_DAYS" desc="Ad Blocking frequency multiple day" formatter_data="android_java">
++        Days
++      </message>
++      <message name="IDS_OPTIONS_ADBLOCK_STARTCHECK_LABEL" desc="The title of the Ad Blocking button to check update" formatter_data="android_java">
++        Check Now
++      </message>
++      <message name="IDS_OPTIONS_ADBLOCK_EVENT_1" desc="The text for ADBLOCK_CHECKING_FOR_UPDATES" formatter_data="android_java">
++        Adblock filter: checking for update
```
csagan5, line 626:
```suggestion
+        Checking for filters update...
```

build/patches/Bromite-AdBlockUpdaterService.patch
```diff
@@ -405,6 +604,39 @@ diff --git a/chrome/app/generated_resources.grd b/chrome/app/generated_resources
 +      <message name="IDS_OPTIONS_ADBLOCK_SUMMARY" desc="The title of the Ad Blocking summary on Android" formatter_data="android_java">
 +        Configure Ad Blocking and filters URL
 +      </message>
++      <message name="IDS_OPTIONS_ADBLOCK_CURRENT_INDEX_TITLE" desc="The title of the Ad Blocking current index version on Android" formatter_data="android_java">
++        Current Indexed Filter Version
++      </message>
++      <message name="IDS_ADBLOCK_LAST_CHECKED_TEXT" desc="The title of the Ad Blocking last checked datetime" formatter_data="android_java">
++        Last Checked:
```
csagan5, line 611:
```suggestion
+        Last checked:
```

build/patches/Bromite-AdBlockUpdaterService.patch
```diff
@@ -405,6 +604,39 @@ diff --git a/chrome/app/generated_resources.grd b/chrome/app/generated_resources
...
++        Check Now
++      </message>
++      <message name="IDS_OPTIONS_ADBLOCK_EVENT_1" desc="The text for ADBLOCK_CHECKING_FOR_UPDATES" formatter_data="android_java">
++        Adblock filter: checking for update
++      </message>
++      <message name="IDS_OPTIONS_ADBLOCK_EVENT_3" desc="The text for ADBLOCK_UPDATE_DOWNLOADING" formatter_data="android_java">
++        Adblock filter: downloading update...
++      </message>
++      <message name="IDS_OPTIONS_ADBLOCK_EVENT_4" desc="The text for ADBLOCK_UPDATED" formatter_data="android_java">
++        Adblock filter: update successfully installed
```
csagan5, line 632:
```suggestion
+        Updated filters successfully downloaded
```

build/patches/Bromite-AdBlockUpdaterService.patch
```diff
@@ -405,6 +604,39 @@ diff --git a/chrome/app/generated_resources.grd b/chrome/app/generated_resources
...
++      </message>
++      <message name="IDS_OPTIONS_ADBLOCK_ERROR_3" desc="The text for UPDATE_NOT_NEEDED" formatter_data="android_java">
++        Adblock filter: update not needed
```
csagan5, line 635:
```suggestion
+        Filters already up to date
```

csagan5:
I think there is a bug when in airplane mode: I tapped Check Now a few times and in the logs I can see:
```
02-14 09:48:33.082  3106  3106 I chromium: [INFO:adblock_updater_service.cc(120)] AdBlockUpdaterService: starting up. First update attempt will take place in 604800 seconds. Next update attempt will take place in 604800 seconds. 
02-14 09:48:34.368  3106  3106 I chromium: [INFO:adblock_updater_service.cc(187)] AdBlockUpdaterService: MostRecentIndexedVersion = 2.0.21.81035
02-14 09:48:57.283  3106  3106 I chromium: [INFO:adblock_updater_service.cc(160)] AdBlockUpdaterService: update delayed. Wait for 37.0845 s
02-14 09:49:44.383  3106  3106 I chromium: [INFO:adblock_updater_service.cc(160)] AdBlockUpdaterService: update delayed. Wait for -10.0154 s
02-14 09:50:05.711  3106  3106 I chromium: [INFO:adblock_updater_service.cc(160)] AdBlockUpdaterService: update delayed. Wait for -31.3438 s
```

But on the UI it still stays at `checking for update`; the first update should eventually time out and the UI show that there was an error?

csagan5:
2 other bugs: start with a clean install in airplane mode, this the relevant log:
```
02-14 09:52:19.334  3745  3745 I chromium: [INFO:adblock_updater_service.cc(187)] AdBlockUpdaterService: MostRecentIndexedVersion = 
02-14 09:52:19.334  3745  3745 W chromium: [WARNING:adblock_updater_service.cc(227)] AdBlockUpdaterService: failed to convert version to time.
02-14 09:52:19.334  3745  3745 I chromium: [INFO:download_filters_task.cc(116)] DownloadFiltersTask: Start downloading
```
It stays stuck there, however:
* when going to settings and tapping `Check Now` it crashes
* it should not try to parse the empty `MostRecentIndexedVersion`, the whole block can be under an `if (!version.content_version.emtpty()) {` or similar

build/patches/Bromite-AdBlockUpdaterService.patch
uazo: yes, the old patch didn't use it, now it's called up.
however, the called function is currently `base::DoNothing()`

https://source.chromium.org/chromium/chromium/src/+/master:components/component_updater/timer_update_scheduler.cc;l=21;drc=8cfbfc3a40cb9513286d6a94cbf17783d558745b;bpv=1;bpt=1

build/patches/Bromite-AdBlockUpdaterService.patch
uazo: fixed in https://github.com/bromite/bromite/pull/905/commits/80c163754a3afa025bd7004aec20d43df53a4c12

build/patches/Bromite-AdBlockUpdaterService.patch
uazo: fixed in https://github.com/bromite/bromite/pull/905/commits/80c163754a3afa025bd7004aec20d43df53a4c12

build/patches/Bromite-AdBlockUpdaterService.patch
uazo: fixed in https://github.com/bromite/bromite/pull/905/commits/80c163754a3afa025bd7004aec20d43df53a4c12

build/patches/Bromite-AdBlockUpdaterService.patch
uazo: fixed in https://github.com/bromite/bromite/pull/905/commits/80c163754a3afa025bd7004aec20d43df53a4c12

build/patches/Bromite-AdBlockUpdaterService.patch
uazo: fixed in https://github.com/bromite/bromite/pull/905/commits/80c163754a3afa025bd7004aec20d43df53a4c12

build/patches/Bromite-AdBlockUpdaterService.patch
uazo: fixed in https://github.com/bromite/bromite/pull/905/commits/80c163754a3afa025bd7004aec20d43df53a4c12

build/patches/Bromite-AdBlockUpdaterService.patch
uazo: fixed in https://github.com/bromite/bromite/pull/905/commits/80c163754a3afa025bd7004aec20d43df53a4c12

uazo:
> But on the UI it still stays at checking for update; the first update should eventually time out and the UI show that there was an error?

I couldn't replicate it, but I got around by checking the status first

> it should not try to parse the empty 

fixed

> I tapped Check Now a few times...
> when going to settings and tapping Check Now it crashes

ah ah fixed! I always forget to check the repetitive impulsive click :) ! although also my customers often do it too... 

let me know if it's all fine

csagan5:
> I always forget to check the repetitive compulsive click

In this case I waited a bit after each click and saw that the wait time goes negative. Is that fixed as well? Does a negative time get converted into ~2^64 seconds?

csagan5 (approved):

build/patches/Bromite-AdBlockUpdaterService.patch
```diff
@@ -983,12 +1531,56 @@ new file mode 100644
...
++  base::Time lastOk = pref_service_->GetTime(kAdBlockLastCheckTimeOk);
++  return lastOk.ToJavaTime();
++}
++
++int AdBlockUpdaterService::GetAdBlockUpdateTimeFrequency() {
++  return pref_service_->GetInteger(kAdBlockFiltersCheckFrequency);
++}
++
++void AdBlockUpdaterService::SetAdBlockUpdateTimeFrequency(int days) {
++  pref_service_->SetInteger(kAdBlockFiltersCheckFrequency, days);
```
csagan5, line 1562: When selecting the `Never` option it looks like the update will still be started.

csagan5 (changes requested):
There is only 1 issue left.

Thanks for your work!

uazo:
> In this case I waited a bit after each tap and saw that the wait time goes negative. Is that fixed as well?

Yes, it's fixed

build/patches/Bromite-AdBlockUpdaterService.patch
uazo: it's correct, update does not start and the message is

chromium: [INFO:adblock_updater_service.cc(122)] AdBlockUpdaterService: user disabled.

build/patches/Bromite-AdBlockUpdaterService.patch
```diff
@@ -820,22 +1321,42 @@ new file mode 100644
...
++void AdBlockUpdaterService::StartWithDelay(int delay) {
++  int days = GetAdBlockUpdateTimeFrequency();
++  if (days < 0) days = 7;
++  next_check_delay_ = 60*60*24*days;
++
++  if (delay <= 0) delay = next_check_delay_;
++
++  scheduler_->Stop();
++  if (next_check_delay_ == 0) {
++    LOG(INFO) << "AdBlockUpdaterService: user disabled.";
```
csagan5, line 1336:
```suggestion
+    LOG(INFO) << "AdBlockUpdaterService: disabled by user.";
```

csagan5 (approved):

csagan5:
Yet another case of brilliant work, many thanks!

uazo:
checking the content filter source, I saw that the patch use a wrong activation flag (the settings on/off is wrong).
I also saw that it is possible to have multiple files of filters, and that if we wanted to, given what I learned with saf, we could load the .dat from disk (useful for diagnosing a possible filter problem).

csagan5:
> checking the content filter source, I saw that the patch use a wrong activation flag (the settings on/off is wrong).

I do not understand; which patch? We are working with patches to patches here...do you mean the original patch or the version you changed in this PR?

> I also saw that it is possible to have multiple files of filters, and that if we wanted to, given what I learned with saf, we could load the .dat from disk (useful for diagnosing a possible filter problem).

With SAF one can specify local file URLs? Are you sure that it would work?
As for multiple filter files: where would this be exactly possible? I don't think that you can load multiple filters by pasting multiple URLs in the UI textbox.

uazo:
> I do not understand; which patch? We are working with patches to patches here...do you mean the original patch or the version you changed in this PR?

sorry, I mean `build/patches/Bromite-AdBlockUpdaterService.patch`. This code seems incorrect:

```
        ChromeSwitchPreference mAdBlockSwitch =
                (ChromeSwitchPreference) findPreference(PREF_ADBLOCK_SWITCH);
        boolean isAdBlockEnabled = !WebsitePreferenceBridge.isCategoryEnabled(browserContextHandle, ContentSettingsType.ADS);
        mAdBlockSwitch.setChecked(isAdBlockEnabled);
        mAdBlockSwitch.setOnPreferenceChangeListener((preference, newValue) -> {
            WebsitePreferenceBridge.setCategoryEnabled(browserContextHandle, ContentSettingsType.ADS, !(boolean) newValue);
            return true;
        });
```
If I turn it off, the adblock continues to work but i see this error:
```
[FATAL:document_subresource_filter.cc(28)] Check failed: activation_state_.activation_level != mojom::ActivationLevel::kDisabled (ActivationLevel::kDisabled vs. ActivationLevel::kDisabled)'
Abort message: '[FATAL:document_subresource_filter.cc(28)] Check failed: activation_state_.activation_level != mojom::ActivationLevel::kDisabled (ActivationLevel::kDisabled vs. ActivationLevel::kDisabled)'
```
the correct flag is `kSafeBrowsingSubresourceFilter`, see https://source.chromium.org/chromium/chromium/src/+/master:components/subresource_filter/core/browser/subresource_filter_features.cc;l=180;bpv=1;bpt=1
it may not matter, but sooner or later I would like to re-enable automatic tests

> With SAF one can specify local file URLs? Are you sure that it would work?

yes, wait until I finish the other patch (export bookmarks), then I'll show you

csagan5:
> the correct flag is `kSafeBrowsingSubresourceFilter`

Yes and no: the problem is that there is no content setting accessible by the user for `kSafeBrowsingSubresourceFilter`; do you have a solution in mind for this inconsistency?

> sooner or later I would like to re-enable automatic tests

The only reason I never looked into that is because of the extra development and testing time that would be needed, but they are indeed a necessity.

uazo:
> Yes and no: the problem is that there is no content setting accessible by the user for kSafeBrowsingSubresourceFilter; do you have a solution in mind for this inconsistency?

sure, it's not a problem, but too many things are overlapping, we have to put order.

> The only reason I never looked into that is because of the extra development and testing time that would be needed, but they are indeed a necessity.

I'm setting up a CI, for now I'm using it to bring bromite to v89, then let's see.

csagan5:
> sure, it's not a problem, but too many things are overlapping, we have to put order.

If you want order we could start from not having te discussion here on the merged PR :) it is not related and will be difficult to find even for ourselves in future

> I'm setting up a CI, for now I'm using it to bring bromite to v89, then let's see.

I would prefer that we have coordination in doing the patches update, to avoid double work; unless you will always re-port the patches even if I am doing that as an exercise?

uazo:
> it is not related and will be difficult to find even for ourselves in future

of course, you're right

> I would prefer that we have coordination in doing the patches update, to avoid double work; 

here too, I open an issue

> unless you will always re-port the patches even if I am doing that as an exercise?

both, the more is learning. the problem is that working on it I started to see that some patches are incomplete (in my opinion)

csagan5:
> both, the more is learning. the problem is that working on it I started to see that some patches are incomplete (in my opinion)

You are probably right; I am glad you have the patience to look into them, much appreciated!

Feel free to open an issue for every problem you spot.

uazo:
> Feel free to open an issue for every problem you spot.

ok, I will do it.

PR #909, opened 2021-02-06
Patch for HTTPS proxy auth support

csagan5:
Preliminary support for HTTPS proxy auth


csagan5:
@uazo this is an old patch of mine that I never finished up; interested? The idea is to add basic auth support for HTTPS proxies

chromer030:
@csagan5 What about Socks5 auth support ?

csagan5:
@chromer030 that is not an option, it transfers passwords in cleartext.

PR #922, opened 2021-02-09
remove TabModelSelectorBase.java changes

### LINKED ISSUE #899: Bromite crashes when opening links  (in incognito only mode)
Running the latest version. I'm pretty sure chromium works fine. 

### Bromite version

Version: `88.4324.95.`
Arch: arm64
Android version: 11
Device model: HD1903

No

Doesn't seem to be

### Is the bug a crash? 

Yes

### Describe the bug

When I click a link in apps, bromite opens up, but crashes while loading the page. Every time.

Steps to reproduce the bug:
1. Go to any link,that opens in your browser. 
2. Click it
3. Bromite opens than crahses

### Expected behavior

Links actually open up. 

Same as #453

uazo:
Patch for #899

### DIFF
--- a/build/patches/Add-an-always-incognito-mode.patch
+++ b/build/patches/Add-an-always-incognito-mode.patch
@@ -18,12 +18,11 @@ Enable incognito custom tabs and fix crashes for incognito/custom tab intents (c
  .../browser/init/StartupTabPreloader.java     | 11 ++-
  .../privacy/settings/PrivacySettings.java     |  4 +-
  .../browser/tabmodel/ChromeTabCreator.java    | 16 +++-
- .../tabmodel/TabModelSelectorBase.java        |  8 ++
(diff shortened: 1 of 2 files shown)


csagan5:
What is the fix that solves #899? I wonder why 2 patches are at play

uazo:
sorry, you're right, it was just in my head!
for #899 it is sufficient to commit https://github.com/bromite/bromite/pull/922/commits/634c5b2a52bc75e97558dbdec4a222b647664804, but checking https://github.com/bromite/bromite/issues/899#issuecomment-774433400 I noticed that I allow an empty ua which is not good, so I add https://github.com/bromite/bromite/pull/922/commits/e94630c2b622bd0ed7acda047d0e2ac3ef03fedb

csagan5:
Ok, understood. I need to know this so that can be documented in the CHANGELOG

PR #925, opened 2021-02-11
Update DIAL repeating discovery patch

Ahrotahn:
This PR adds a fix to the DIAL repeating discovery patch to resolve a potential crash when the network state changes.


Zoraver:
This looks [familiar](https://github.com/Eloston/ungoogled-chromium/commit/76c969401b10e2cbced802e915accb33382f9d0f) ;)

PR #927, opened 2021-02-13
Fix favicons search for fallback

### LINKED ISSUE #788: Favicons are occasionally swapped
### Bromite version

Version: `86.0.4240.112`
Arch: `arm64`
Android version: `9.0`
Device model: `Galaxy S20`

Yes

No

### Describe the bug

The favicons are occasionally swapped for websites.

Upstream issue: https://bugs.chromium.org/p/chromium/issues/detail?id=915374

uazo:
Patch for upstream issue #788

### DIFF
--- a/build/bromite_patches_list.txt
+++ b/build/bromite_patches_list.txt
@@ -157,3 +157,4 @@ webview-Hard-no-to-persistent-histograms.patch
 Ignore-managed-state-and-policies-for-secure-DNS.patch
 Do-not-permit-user-installed-certificates.patch
 Automated-domain-substitution.patch
+fix-favicons-fallback.patch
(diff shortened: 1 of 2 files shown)


csagan5 (approved):

csagan5:
Nice work; if you have the patience you could try submitting the patch directly upstream.

Note: in future please always put the patch before the last one, `Automated-domain-substitution.patch`.

uazo:
> if you have the patience you could try submitting the patch directly upstream.

by now you have understood that I have little to do with English ...
if you want (and you have time) can you do it? there would also be the other proxy bug ...

> Note: in future please always put the patch before the last one

ah ah, certainly, no problem, I remembered you told me to put it down!

csagan5:
> ah ah, certainly, no problem, I remembered you told me to put it down!

If I did, my mistake; it is better before `Automated-domain-substitution.patch` to avoid conflicts related to replacements

csagan5:
> if you want (and you have time) can you do it? there would also be the other proxy bug ...

You should get credit for it; do not worry about English, when you open them I can quickly review them for you.

uazo:
> You should get credit for it

thanks, but i don't care, if you want, do it in the name of the "bromite team"

Issue #941, opened 2021-02-21
New Adblock list checker shows last checked time Dec. 20, 1969


Pentaphon:
This is minor so it just needs a screenshot to confirm.  

![time](https://user-images.githubusercontent.com/49374058/108627458-8d0f0980-740a-11eb-80bc-0b9b7f3061ee.png)

chromer030:
Same here.
Sometimes it shows true date , sometimes wrong :

![Screenshot_20210221-184111-01](https://user-images.githubusercontent.com/13361552/108629358-bf555100-7474-11eb-9abc-ec9180e7b9d4.jpeg)

NineOutOfTen:
I'm also seeing ads on websites I visit too. I did use the update button but it didn't work. I did post this as a separate issue but it is closed without comment.

csagan5:
@uazo these are the bugs I could individuate:
* on startup it shows the wrong date and time
* it does not actually do the first update after 5 seconds
* after tapping check now, if you exit the page and then come back, it will still show the wrong date and time; returning to the page one more time will then show the right date and time

So it seems that until the download logic triggers the date is always wrong, and there is a missing update on the UI; perhaps a model observer problem?

uazo:
> perhaps a model observer problem?

I don't know, maybe some event that occurs when user press back button that I manage badly. however I think I can fix it easily, damn I didn't notice before.
however the filters are loaded so I don't think the adblock has any problems.

uazo:
>  I did use the update button but it didn't work.

@NineOutOfTen please post a screenshot

NineOutOfTen:
![Screenshot_20210221-161343.png](https://user-images.githubusercontent.com/60764728/108641989-d0479780-745f-11eb-8ef3-e226bedc3e86.png)

NineOutOfTen:
![Screenshot_20210221-161504.png](https://user-images.githubusercontent.com/60764728/108642014-fe2cdc00-745f-11eb-8eee-a5e6f64ccc45.png)

chromer030:
Same here, ad blocker not working.

uazo:
@csagan5 what kind of regression is it, in your opinion?

technically speaking, there are no changes related to how the filter works, the only one is that before the https request was always made at every start, now only every 7 days or at first installation.

more information is needed, @chromer030 @NineOutOfTen a log extraction would be useful.

NineOutOfTen:
[...]

PR #946, opened 2021-02-23
Add a flag (default enabled) for "Ignore enterprise policies for secure DNS" patch

uazo:
as anticipated https://github.com/bromite/bromite/issues/941#issuecomment-783665701
it is probably best to always add a flag to each patch


csagan5:
It is probably a good idea, ungoogled-chromium already does this, but I am not convinced we need to introduce this now for this feature.

uazo:
you see, for me it is useful in case of troubleshooting, and it doesn't hurt.
rejects the pr if you want.

csagan5:
I look at it from the perspective of extra patch maintenance and because of all the other features/changes that do not have it.

If we want to merge this we should probably invert it? e.g. `kEnableDeviceEnterpriseInfo` by default disabled

uazo:
> I look at it from the perspective of extra patch maintenance and because of all the other features/changes that do not have it.

I'm slowly reviewing all the patches, it's a bit long as an activity, we might think about adding them the first time we touch them.

> e.g. kEnableDeviceEnterpriseInfo by default disabled

sure

uazo:
sorry for the long wait! I'm cleaning the various pulls

I inverted the flag...
I moved the patch to the bottom because it was more convenient for me, sorry if this can take you more with the rebase, but I don't think

csagan5:
Thanks for your update, I am still not convinced about this patch for two reasons:

* who would be needing this patch? e.g. some user that has problems because an enterprise policy is not in act: do we have any?
* wouldn't we need a patch like this for any enterprise policy we override?

uazo:
> who would be needing this patch? e.g. some user that has problems because an enterprise policy is not in act: do we have any?

yes, you are right, it makes little sense, if I were an it manager I would hardly propose the use of bromite in the enterprise environment.
leave it as it is now `if(true)`

> wouldn't we need a patch like this for any enterprise policy we override?

no, for the reasons mentioned above.
eventually it would be interesting the kiosk mode with the always unknown (and possibly the block to access the settings) for those who let use the tablet in public places, if any company is interested (and contributed to the project :).

let's close the pr.

PR #956, opened 2021-02-27
[experiment] v89 patches

uazo:
it's a work in progress... as I do them there will be another commit

some patches are heavily modified, just because every patch can compile

NOTE: only some of these patches will be used in next Bromite release


hefty-fading-culprit-epiphany-drew:
Are these uc patches?

uazo:
> Are these uc patches?

yes, "slightly" modified because they compile now.
the approach was different: I removed the dependencies from the .gn and recompiled to success, in the meantime modified all the code in reference.
doing so was easier and I also eliminated the dependencies from the webview (which @csagan5 as I told you I can't test, however, particularly
https://github.com/bromite/bromite/blob/736cca95faebff6d041ca0d8e4a645726e4d8e7f/build/patches/Disable-safe-browsing.patch#L694 and https://github.com/bromite/bromite/blob/736cca95faebff6d041ca0d8e4a645726e4d8e7f/build/patches/Disable-safe-browsing.patch#L702 they worry me a little)

csagan5:
> yes, "slightly" modified because they compile now.

They are not ungoogled-chromium patches, these are Bromite patches. Some of them are from ungoogled-chromium.

> they worry me a little)

It does not worry me; in all these cases I checked that the browser returns `nullptr` when the service is not available. Checking again of course does not hurt.

hefty-fading-culprit-epiphany-drew:
> > yes, "slightly" modified because they compile now.
> 
> They are not ungoogled-chromium patches, these are Bromite patches. Some of them are from ungoogled-chromium.
> 
> > they worry me a little)
> 
> It does not worry me; in all these cases I checked that the browser returns `nullptr` when the service is not available. Checking again of course does not hurt.

Are you re writing them to be compatible with chromium 89?

uazo:
> Are you re writing them to be compatible with chromium 89?

Yes.
I'm taking a while because this time I want to try to verify (unfortunately again by hand) each patch before moving on to the next.

@csagan5 i think i got the chromium version wrong though.

I started from 89.0.4389.57 but it seems a bit buggy to me, especially there are a lot of failed dchecks (not because of me) in
cc/metrics/compositor_frame_reporter.cc, DCHECK which are no longer present in the current version.
I don't think however that this work is useless though, how do you know which version to use?

csagan5:
> how do you know which version to use?

I use always the latest; it corresponds to Beta until it is released (then becomes Stable).

csagan5:
@uazo basically I cannot use these unless I go through each patch to see if there are differences or not; next time please annotate the changes that you have made in the PR description.

csagan5:
My findings: for some patches there are only minor changes, but even when the patch summary does not change much there can be important changes.

In future if you want to merge a PR which also does this type of refactoring it is necessary to annotate all the choices you made for the changes.

uazo:
>  patch summary does not change much there can be important changes.

apart from the patches related to ungoogle that I have modified, for the others I have only made changes for adapt to v89

> next time please annotate the changes that you have made in the PR description.

what exactly do you mean? if for example a method or a class no longer exists in v89 should I note it?
do you mean such a thing?

Add-an-always-incognito-mode: they changed many methods from string to gurl
Bromite AdBlockUpdaterService: They have removed some styles from the resources
Increase number of autocomplete matches from 5 to 10: they moved the method that was used to define that number
Multiple fingerprinting mitigations: they have eliminated the method used for shuffle
etc..

I can always do it if needed

> In future if you want to merge a PR

no intention to merge, did you know, for me it was a test to understand if it is possible to simplify the builds for testing and c.i.

uazo:
> for testing and c.i.

in this regard, as I told you, I was able to recover 2 (old) biprocessor servers (8core each) with 8gb of ram each and ssd disk.
one I have already set up, putting 16gb of ram, but during the compilation I saw that 8gb are always free.
the times still do not satisfy me: to prepare the environment it takes 2 hours and 6 hours for the build, probably due to the dated processor, maybe I recover a better one.
I am going to check if it is possible to modify gclient.py to improve download times, because the history seems to me that it continues to download event if I tell that I do not want it, while for 6 hours I am setting up goma, I tried but unfortunately not still compiles.
I have also seen that the fwrapv patch increases my time considerably (and therefore for the test build we might not even include it) and that compiling blink is the slowest part of the whole process (I don't know how to improve this yet) .
no problem with the link phase which is really fast with the ssd.

csagan5:
> apart from the patches related to ungoogle that I have modified, for the others I have only made changes for adapt to v89

Yes, but when the difference is not documented there is no other option than going through each patch and as you know it is very difficult to see "the diff of a diff".



> what exactly do you mean? if for example a method or a class no longer exists in v89 should I note it?
> do you mean such a thing?

No, only the choices: if it is a forced choice you can note it as well, but mostly it is about what you have decided to change.

> I can always do it if needed

I was trying to see if I could use some patches, in some cases I looked at yours to see how you did it, but in the end I could not use them "in bulk".

> no intention to merge, did you know, for me it was a test to understand if it is possible to simplify the builds for testing and c.i.

Sure, no problem; go ahead. But it is a PR and PRs are supposed to be merged, I concluded this one cannot be. I wanted to make it clear because I see all the effort you are putting into it; I know about your C.I. experiments, here I am just handling the PR.

uazo:
> there is no other option than going through each patch and as you know it is very difficult to see "the diff of a diff".

Yes, I realize, it is very difficult this way.
i saw https://sourcegraph.com/github.com/chromium/chromium, if we used it too in a bromite repo with chromium sources?

> but mostly it is about what you have decided to change.

i think i will propose a merge of all ungoogle patches, they are the only ones that have bothered me.

> but in the end I could not use them "in bulk".

I apologize, I thought it was faster, this time I started even earlier, but I didn't think that after 2 weeks I still hadn't finished.

> But it is a PR and PRs are supposed to be merged,

next time I write "don't merge"

csagan5:
> Yes, I realize, it is very difficult this way.
> i saw https://sourcegraph.com/github.com/chromium/chromium, if we used it too in a bromite repo with chromium sources?

I am open to install and maintain any tool that can help us or any other developer try out patches. Good idea; there is also Google's one right? We could give a look.

> next time I write "don't merge"

No worries, I closed it because it may cause confusion if someone else is looking for v89 patches for Bromite. I can re-open it and change title.

> I apologize, I thought it was faster, this time I started even earlier, but I didn't think that after 2 weeks I still hadn't finished.

No need to apologise, as I said I am just doing PR handling.

build/patches/Disable-promos-displayed-in-bookmarks-manager.patch
```diff
@@ -5,81 +5,22 @@ Subject: Disable promos displayed in bookmarks manager
...
 bookmarks, never reach signin thresholds.
 ---
- chrome/android/chrome_java_resources.gni      |   1 -
- ...rsonalized_signin_promo_view_bookmarks.xml |  30 -----
- .../bookmarks/BookmarkItemsAdapter.java       |  30 +----
- .../bookmarks/BookmarkPromoHeader.java        | 121 +-----------------
- .../browser/signin/SigninPromoController.java |  23 +---
- 5 files changed, 5 insertions(+), 200 deletions(-)
- delete mode 100644 chrome/android/java/res/layout/personalized_signin_promo_view_bookmarks.xml
+ .../bookmarks/BookmarkItemsAdapter.java       |  29 +----
```
csagan5, line 8: For example for this patch there is a moved file in `chrome/browser/signin/ui/android/java/src/org/chromium/chrome/browser/signin/ui/SigninPromoController.java` that was missed; without documenting this, nobody can review it.

uazo: but you have to see here

https://github.com/uazo/bromite/blob/v89/build/patches/Disable-promos-displayed-in-bookmarks-manager.patch

maybe doing a different commit each time is not good

csagan5: > but you have to see here

Can you explain what to see there? The changes to `SigninPromoController` are missing.

uazo: I didn't understand.. damn you're right.

i think the problem is in the wiggle options i used, i was missing "--no-ignore", but i realized it too late, when i found myself in a git am in progress, four patches ahead...
maybe I also need something that verifies the number of files modified before and after

I thought I reprocessed it, but once again I have to learn, luckily you checked, nothing escapes you :)

csagan5: At the beginning I was using some of your patches, then I noticed something like this ^ so I stopped; that's why I mentioned the importance of documenting the changes :)

So in this case I thought it was intentional, but without documenting what has been (willingly) changed it is not possible to do this type of review.

uazo: you're absolutely right, now I understand better what you meant.
I hope they can help at least to see quickly what needs to be changed, because for some I had to go rummaging through the chromium history

csagan5:
@uazo can you please submit the updated patch for user agent customization?

csagan5:
Also, on the timezone patch, I believe the `<ListPreference />` in the `components/browser_ui/site_settings/android/java/res/xml/single_website_preferences.xml` file is incorrect; it works differently now, they are populated dynamically.

uazo:
> @uazo can you please submit the updated patch for user agent customization?

it's here https://github.com/uazo/bromite/blob/v89/build/patches/User-agent-customization.patch

> Also, on the timezone patch

yes it looks different, I look at it

uazo:
> Also, on the timezone patch

here the revised patch
https://github.com/uazo/bromite/blob/v89/build/patches/Timezone-customization.patch

csagan5:
Ok, thanks; will re-open the PR so that updates show up.

csagan5:
@uazo I plan to not include the patch for the favicons fix in next release, since it does not solve the problem.

I have pushed the v89 patches I am currently using; there are probably still a few build errors but I have not met them yet. I have merged the block of patches that removes binary/cloud integrations, I hope it helps for future patch porting efforts.

I will fix build errors as I come across them; if you spot any other bug/problem I will be happy to fix it as well.

csagan5:
The v89 patches are released; if you have taken note of the problems then we could fix those now.

uazo:
sure, I start.

Issue #985, opened 2021-03-08
enable PartitionExpectCTStateByNetworkIsolationKey too


thestinger:
This feature is included in the `PartitionConnectionsByNetworkIsolationKey` field trial too. There are 7 features enabled there on top of the baseline `SplitCacheByNetworkIsolationKey`. Bromite has 5/7 of these enabled right now.

thestinger:
You might as well also enable `PartitionNelAndReportingByNetworkIsolationKey` too even if it's not expected to do anything for Bromite. The `PartitionDomainReliabilityByNetworkIsolationKey` feature is already enabled by default, although domain reliability is similarly disabled.

csagan5:
Related #836

Are these all enabled in Vanadium too?

thestinger:
They're all going to be enabled now to do the same thing as that field trial.

csagan5:
Fixed in `89.0.4389.78`.

PR #995, opened 2021-03-09
Set enable_reporting to false

nikolowry:
`enable_reporting` has been enabled since v83.0.4103.46, https://github.com/bromite/bromite/commit/65bd5c1df9f2e4b552bbf2ef32b8e43ec58a0499. 

This PR sets it back to false and adds `ungoogled-chromium-Fix-building-without-enabling-reporting.patch`
to fix build issues. See https://github.com/Eloston/ungoogled-chromium/pull/1416 for more details.


csagan5:
Nice finding, but what is the impact of `enable_reporting` being set to true? Is any undesired connection happening?

csagan5:
It was enabled in v83 because of COEP: https://bugs.chromium.org/p/chromium/issues/detail?id=887967

nikolowry:
@csagan5 I thought it was just an oversight setting it to true. I'll dig into the linked Chromium issue and report back in the next few hours.

Setting `enable_reporting=true` shouldn't negatively effect privacy-related functionality due to the domain substitution patches, correct? The domain substitution is a catch-all to prevent unnoticed tracking features from creeping into builds, right? 

Regardless, I'll report back after I review.

wchen342:
From what I read COEP reporting and the policy itself are two separate thing? I had a look at the navigation and render host part and there is nothing hidden behind the reporting flag.

nikolowry:
@csagan5 isn't the point of these lines https://github.com/bromite/bromite/blob/9824847bda54591cffe0db70506c1b57e6816645/build/patches/ungoogled-chromium-Fix-building-without-enabling-reporting.patch#L1-L24 to not make COEP/COOP-support dependent on `enable_reporting` being true?

@Ahrotahn explains why the other changes are needed in this comment  https://github.com/Eloston/ungoogled-chromium/pull/1416#issuecomment-791502070.

I haven't had any issues today after building v89 with `enable_reporting=false` and the `ungoogled-chromium-Fix-building-without-enabling-reporting.patch`.

csagan5:
>  isn't the point of these lines to not make COEP/COOP-support dependent on `enable_reporting` being true?

Yes; but as far as I know no reporting of any other type is happening right now; it would be nice to confirm/deny that.

> I haven't had any issues today after building v89 with `enable_reporting=false` and the `ungoogled-chromium-Fix-building-without-enabling-reporting.patch`.

Yes, but for all changes we need to have a reason to make them and then document it.

nikolowry:
> Yes; but as far as I know no reporting of any other type is happening right now; it would be nice to confirm/deny that.

Any suggestions on a reliable way to profile and test requests?

> Yes, but for all changes we need to have a reason to make them and then document it.

I'll be the first to admit I have very limited knowledge of Chromium's net architecture -- hopefully someone more familiar can chime in on whether `enable_reporting=true` cause negative effects.

Feel free to close this or keep it open until we have more context to make an informed decision.

csagan5:
> Any suggestions on a reliable way to profile and test requests?

None other than Wireshark; see also #471

By following the code it should be possible to understand what reporting does (or not).

uazo:
ah, I didn't notice, thanks @csagan5 
so it is not only active, but also with cookies!

uazo:
> By following the code it should be possible to understand what reporting does (or not).

I got there by checking something else, but here it is explained:
https://source.chromium.org/chromium/chromium/src/+/master:services/network/expect_ct_reporter.cc;l=116;drc=70365c5dbd0628df048dcc4e81740b3932411574
however, it is not the only point that uses that call:
https://source.chromium.org/chromium/chromium/src/+/master:services/network/network_context.cc;l=2241;drc=70365c5dbd0628df048dcc4e81740b3932411574;bpv=1;bpt=1

> Any suggestions on a reliable way to profile and test requests?

can this be enough?
https://source.chromium.org/chromium/chromium/src/+/master:services/network/expect_ct_reporter_unittest.cc

csagan5:
> so it is not only active, but also with cookies!

1. are cookies there by design of the certificate reporting (necessary) or they are unnecessary and a privacy violation?
2. certificate reporting is currently happening in Bromite for websites that have an invalid certificate and specify the headers for certificate reporting?
3. is invalid certificate reporting to be considered a privacy-invasive feature for the user? Can it be used/misused in such sense?
4. the patch in this PR would allow to still have COEP/COOP but no reporting, correct?

(3) is the most important aspect here

uazo:
> are cookies there by design of the certificate reporting (necessary) or they are unnecessary and a privacy violation?

forget it, a more careful check allowed me to understand that cookies are **NOT** active.

regarding the other points, personally I think that any request not authorized by the user should never be made.

csagan5:
I still do not have a resolution for the other questions (2,4)

> regarding the other points, personally I think that any request not authorized by the user should never be made.

The requests are sent to the same site the user is visiting, not a third-party. What are the conditions under which these requests are sent?

uazo:
> not a third-party

should it be verified or are you sure?

> What are the conditions under which these requests are sent?

the documentation and code must be studied and the test cases verified, it can be a long thing.

csagan5:
> should it be verified or are you sure?

I am sure; it must be the same domain; feel free to double-check!

> the documentation and code must be studied and the test cases verified, it can be a long thing.

Sure; but it is supposedly a feature for the protection of users (in case they are under MiTM or similar security problems), at the moment I am not sure how it can be abused against users.

uazo:
ok, i'll check.
in any case the correct build flag to disable this, appears to be `IS_CT_SUPPORTED`

csagan5:
> in any case the correct build flag to disable this, appears to be `IS_CT_SUPPORTED`

Related #985, where we enabled `PartitionExpectCTStateByNetworkIsolationKey`

uazo:
> is invalid certificate reporting to be considered a privacy-invasive feature for the user?

```
...
it can’t be the case that this general benefit be allowed to take priority over the ability 
of a user to individually opt-out of such a system
...
User agents MUST allow users to disable reporting with some reasonable amount 
of granularity in order to maintain the priority of constituencies espoused in [HTML-DESIGN-PRINCIPLES].
```
https://w3c.github.io/reporting/#disable

so, there should be a flag that deactivates it, but there isn't

csagan5:
> https://w3c.github.io/reporting/#disable
> 
> so, there should be a flag that deactivates it, but there isn't

Since adding a `chrome://` flag does not seem possible I will merge this PR and we go without the feature. Thoughts?

nikolowry:
@csagan5 should I update this PR with the latest patch that's located at https://github.com/Eloston/ungoogled-chromium/blob/a9eb6fd87445465a2e1881f7fc6e718fd17db606/patches/core/ungoogled-chromium/fix-building-without-enabling-reporting.patch?

uazo:
> Since adding a chrome:// flag does not seem possible 

why you say that?

> I will merge this PR and we go without the feature. 

I don't know, the alternative would be put a flag and act only on https://source.chromium.org/chromium/chromium/src/+/master:services/network/network_context.cc;l=2241;drc=70365c5dbd0628df048dcc4e81740b3932411574 however the code must be checked.

uazo:
I came up with another idea.

I don't know if you are aware of [traffic annotator script](https://source.chromium.org/chromium/chromium/src/+/main:tools/traffic_annotation/scripts/README.md), ([i think the result](https://source.chromium.org/chromium/chromium/src/+/main:tools/traffic_annotation/summary/grouping.xml)).

the idea could be to create an ui that allows the user to select which ones he wants to activate.
for implementation I guess the point to act is [simple_url_loader](https://source.chromium.org/chromium/chromium/src/+/main:services/network/public/cpp/simple_url_loader.cc): when blocked, we return a network error, a bit like the replacement of the url done by `Automated-domain-substitution.patch`, theoretically at that point we could even do without it, simplifying the build process.

also theoretically it would also be possible to act in [DefineNetworkTrafficAnnotation()](https://source.chromium.org/chromium/chromium/src/+/main:net/traffic_annotation/network_traffic_annotation.h;l=59;drc=c14f6f4b9c44fe479a8d004576b42723b2a5feb6;bpv=1;bpt=1), by modifying the call in such a way that the compiler communicates new annotators to us going into error.

about the implementation, I don't think it's impossible, a bit 'of doubts about what to return in case it is blocked, because I would not want there were infinite retries, but just check.

csagan5:
> > Since adding a chrome:// flag does not seem possible
> 
> why you say that?

I say that because it is currently a build flag, just assuming that it would be complicated at runtime. But if we have already covered similar complexity/functionality and you think we can have it as a flag it would be better.



> I don't know if you are aware of [traffic annotator script](https://source.chromium.org/chromium/chromium/src/+/main:tools/traffic_annotation/scripts/README.md), ([i think the result](https://source.chromium.org/chromium/chromium/src/+/main:tools/traffic_annotation/summary/grouping.xml)).

I know about traffic annotations but I have never checked the scripts used during tests for them.

> the idea could be to create an ui that allows the user to select which ones he wants to activate.
> for implementation I guess the point to act is [simple_url_loader](https://source.chromium.org/chromium/chromium/src/+/main:services/network/public/cpp/simple_url_loader.cc): when blocked, we return a network error, a bit like the replacement of the url done by `Automated-domain-substitution.patch`

Before doing this we should survey what are the traffic annotations and if there is an use-case for enabling/disabling some. If not, this would not be necessary.

As for the error: the one returned when ad-blocking is also fine.

> theoretically at that point we could even do without it, simplifying the build process.

The domain substitution patch changes strings, which end up in multiple places. Even when they end up in URLs being loaded, they are not always via `simple_url_loader` and they do not have a shared annotation.

> about the implementation, I don't think it's impossible, a bit 'of doubts about what to return in case it is blocked, because I would not want there were infinite retries, but just check.

For this specific PR #995 I suggest a flag, for the traffic annotations we can study it separately in a new issue.

> @csagan5 should I update this PR with the latest patch that's located at https://github.com/Eloston/ungoogled-chromium/blob/a9eb6fd87445465a2e1881f7fc6e718fd17db606/patches/core/ungoogled-chromium/fix-building-without-enabling-reporting.patch?

@nikolowry if we go with the flag then the reporting code would be always compiled but disabled at runtime, so I suggest to hold on that for now.

csagan5:
I am going to try building with `enable_reporting=false` and use the latest ungoogled-chromium patch as needed; having a flag here would be nice, since user might want to be sending this information to the domain they are willingly visiting, but I also would like to speed up the resolution of issues/PRs.

nikolowry:
I'll do the same tonight and update the patch in the PR. Feel free to cool if you beat me to it @csagan5

csagan5:
@nikolowry not necessary, I was already going to merge the PR after verifying that it does not introduce bugs.

nikolowry:
@csagan5 there's been two commits targeted for v94 in Ungoogled's `fix-building-without-enabling-reporting.patch` that appear to be more than just the result of updating via `quilt`:
- https://github.com/Eloston/ungoogled-chromium/commit/a40aba5785042319f0ca770b8c01d2cbf66266cb#diff-ac3a53258dd1326652a15dd19039c927793dcb4d5431d361e1a9f2c943672b35
- https://github.com/Eloston/ungoogled-chromium/commit/7eaad9b7267212d3ce9393b9d9f019734801df39

Maybe the commit author, @Ahrotahn, could weigh in with their thoughts?

Ahrotahn:
That was my understanding as well, but I haven't looked into it since that update.  It seemed like they were still adding more changes at the time.  

Those changes to the reporting patch were made because `kDocumentReporting` is gated behind the `ENABLE_REPORTING` buildflag elsewhere, but they added that section later and forgot to include the buildflag there as well.

csagan5:
Yes, they forgot to gate the `kDocumentReporting` behind the build flag, I have already fixed this in the upcoming v94 patches.



> from what I read, the `kDocumentReporting` flag only activates the modification of the url by reading a header in the response, it would have been more correct to call it `kParsedReportingEndpointsHeader`.
> 
> I also checked the current master and the flag still does just that.

Ok, then we still need this patch.

Issue #570, opened 2020-05-14
Bookmark all tabs


DragoCubed:
### Is your feature request related to privacy?

I think it's more about freedom. It's also related to performance. Bromite allows you to import and export your bookmarks which is an amazing addition. 

### Is there a patch available for this feature somewhere?

No
### Describe alternatives you have considered

a new folder could be created without prompt and allow users to access it quickly from a snackbar button, much like bookmarking a page.

csagan5:
This has nothing to do with privacy.

csagan5:
As I [commented here](https://github.com/wchen342/ungoogled-chromium-android/issues/26#issuecomment-637777685) it looks like this feature already exists for Desktop. We might be able to get it on Android too, minus the UI part.

csagan5:
Filed upstream here: https://bugs.chromium.org/p/chromium/issues/detail?id=1130144

csagan5:
Fixed in `88.0.4324.187`.

@uazo if you see the patch, I tried to use a dedicated folder but the bookmarks in it would disappear after exiting the browser. If you can spot where is the bug, we can restore that part. Right now it saves to mobile folder.

uazo:
I'll look at it but for now I'm fixing #831  by allowing the save upon request with the dialog and it seems to work (you can save directly in the cloud!). I just miss the preference and the strings.

PR #999, opened 2021-03-10
fix menu items display with 6 columns + fix disappear after exiting 

### LINKED ISSUE #963: Menu items are not properly displayed with tab overflow menu regroup flag
### Bromite version

Version: `88.0.4324.207`
Arch: `arm` 
Android version: `5.1`
Device model: `Moto G1 - XT1033`

No

Yes

No

### Is the bug a crash?
Yes

### Describe the bug
When opening the menu its header is empty, the view source icon is broken and the bookmark all tabs menu item is missing.

Steps to reproduce the bug:
1. Install the latest Bromite version (88.0.4324.207);
2. Open the menu;
3. See that the `menu header` is empty, the `View source icon` is broken and the `bookmark all tabs menu item` is missing.

### Expected behavior
When opening the menu all the menu items are displayed properly, as seen in the previous Bromite version.

__Question: What was wrong with the `exit and view source menu items icons`? Apparently they were working fine for me.__

_PS: I know this phone is very old, its my backup device, but everything was working fine in the previous version._

uazo:
fix #963 + see comment https://github.com/bromite/bromite/issues/570#issuecomment-782585897

### DIFF
--- a/build/patches/Add-menu-item-to-bookmark-all-tabs.patch
+++ b/build/patches/Add-menu-item-to-bookmark-all-tabs.patch
@@ -3,14 +3,17 @@ Date: Thu, 18 Feb 2021 21:22:52 +0100
 Subject: Add menu item to bookmark all tabs
 
 ---
- chrome/android/java/res/menu/main_menu.xml    |  6 ++
- .../java/res/menu/main_menu_regroup.xml       |  3 +
- .../chrome/browser/ChromeTabbedActivity.java  | 23 +++++++
+ chrome/android/java/res/menu/main_menu.xml    |  7 +++
+ .../java/res/menu/main_menu_regroup.xml       |  4 ++
+ .../chrome/browser/ChromeTabbedActivity.java  | 26 ++++++++
+ .../AppMenuPropertiesDelegateImpl.java        |  2 +-
  .../browser/bookmarks/BookmarkBridge.java     | 62 +++++++++++++++++++
  .../browser/bookmarks/BookmarkUtils.java      |  4 ++
  .../android/bookmarks/bookmark_bridge.cc      | 17 +++++
  .../android/bookmarks/bookmark_bridge.h       |  4 ++
  .../bookmarks/chrome_bookmark_client.cc       |  2 +
+ .../java/res/layout/icon_row_menu_item.xml    |  4 ++
+ .../browser/ui/appmenu/AppMenuAdapter.java    | 13 +++-
  .../strings/android_chrome_strings.grd        |  3 +
  components/bookmark_bar_strings.grdp          |  6 ++
  .../browser/bookmark_load_details.cc          |  4 ++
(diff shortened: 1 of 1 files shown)


uazo:
Assuming that what I wrote is correct, I would like to make you a proposal on how to write this patch: I think we can try a next step, componentization.

The goal would be to create no more a `.patch` file, but, using `DEPS`, hooks and `.gn`s, separate this code from the base code, in some way by exploiting the logic that chromium makes available to us + those that seem tested and used in brave.

It's a bit a revolution but the benefits could be:
1) improve reviews (no longer based on patches, but directly on the code, which becomes easily comparable)
2) potentially facilitate the transition between one version and another of chromium
3) in all probability simplify the development of tests

these are the steps I imagine:
1) setup of a folder `/src` with a structure to be agreed, I think that one as similar as possible to the one in use in chromium is an advantage
2) definition of a DEPS for gclient that downloads our folder in third_party during the sync phase
3) definition of a hook that executes a git-am with a patch to be built that allow us to fit our code in the base one, where it is not foreseen.
for example I am thinking of the simple modification of functions as virtual, or the introduction of the use of dependency injection constructs or the factory pattern where not foreseen, or simply by calling our functions instead of those present in the code, the best way will only be seen by trying
4) use of all the mechanisms already present in the code for embedders (I think I understand that they are called that)

the c++ part seems to me the most developed in this sense, I have seen this logic many times.
the java part, on the other hand, is a bit tougher, but lately I seem to have seen that they are also moving in this direction.

what do you say, shall we try?

csagan5:
> The goal would be to create no more a `.patch` file, but, using `DEPS`, hooks and `.gn`s, separate this code from the base code, in some way by exploiting the logic that chromium makes available to us + those that seem tested and used in brave.

I am skeptical: what will be the outcome? Will there be no changes needed to Chromium source or less changes?

> 1. improve reviews (no longer based on patches, but directly on the code, which becomes easily comparable)

I do not think all patches can become components.

> 3\. definition of a hook that executes a git-am with a patch to be built that allow us to fit our code in the base one

I think this is a bad idea. Patches are code changes, hooks are programs.

> what do you say, shall we try?

You can try for one of these and then I will evaluate; but no guarantees that it will be merged.

uazo:
> I am skeptical: what will be the outcome? Will there be no changes needed to Chromium source or less changes?

this obviously I do not know a priori, however, the goal is not less changes, but easier maintenance

> I do not think all patches can become components.

you have to try, obviously not the ungoogled patches, but those seem more easy to handle to me. In this regard, if you have time to check the ones I produced in my tests, it's seem more complete to me, therefore no need to apply the Remove-binary-blob-integrations and can be verified one by one (although the hypothesis of aggregating them is the best for me).

> hooks are programs.

hooks can also be bash scripts (https://chromium.googlesource.com/chromium/tools/depot_tools/+/master/gclient.py#236)

```
#     "action"   A list describing a command to run along with its arguments, if
#                any.  An action command will run at most one time per gclient
#                invocation, regardless of how many files matched the pattern.
#                The action is executed in the same directory as the .gclient
#                file.  If the first item in the list is the string "python",
#                the current Python interpreter (sys.executable) will be used
#                to run the command. If the list contains string
#                "$matching_files" it will be removed from the list and the list
#                will be extended by the list of matching files.
```

> You can try for one of these and then I will evaluate; but no guarantees that it will be merged.

no problem, just take it into consideration

Issue #1077, opened 2021-04-15
enable-horizontal-tab-switcher is missing from flags


oldphones68:
Have you read the README?
Yes
Have you searched the issue tracker?
Yes
Are you sure this is not a Chromium bug?
Yes. Since according to the README it is not related to Chromium bug.
Are you using the latest version of Bromite?
Yes 

### Bromite version

Version: `90.0.4430.74.`
Android version: `6.0`
Arch: `arm64`

### Flags

These are the flags changed under `chrome://flags`:

Nothing.

### Is this bug about the SystemWebView?

No

### Is the bug reproducible with latest version?

Yes

### Can the bug be reproduced with corresponding Chromium version?

This feature is NOT related to Chromium. It is Bromite specific according to the readme.

### Is this bug happening in an incognito tab?

No

### Allow ads for the site; is the bug caused by the adblocker?

No

### Is the bug a crash?

No

### Describe the bug

According to the README, there should have an option in the `chrome://flags` to `#enable-horizontal-tab-switcher` but it is not available in `chrome://flags`

### Steps to reproduce the bug

Steps to reproduce the bug:
1. Go to 'chrome://flags'
2. Search for 'enable-horizontal-tab-switcher'
3. See error

### Expected behavior

'enable-horizontal-tab-switcher' should be available in `chrome://flags`

kokukyocho:
Hello @oldphones68!

I am a bot :robot:

Thanks for submitting this issue!
Please make sure that the information under `Bromite version` is complete and correct.

kokukyocho:
Hello @oldphones68!

I am a bot :robot:

Thanks for submitting this issue!
I noticed that it is missing the template, please edit the issue to match the template for either a [bug](https://github.com/bromite/bromite/blob/master/.github/ISSUE_TEMPLATE/bug_report.md) or a [feature request](https://github.com/bromite/bromite/blob/master/.github/ISSUE_TEMPLATE/feature_request.md).

The issue will be automatically re-opened after that.

**Tip**: if you use GitHub in Desktop mode the template will be proposed to you when submitting issues.

csagan5:
It was removed upstream in https://github.com/chromium/chromium/commit/35156bbe110b1049894e7125e0ba2026340ada0a; I will try to restore it in next release but for UI-heavy functionalities like this we cannot keep them around for too long.

oldphones68:
Thank you brother. @csagan5

csagan5:
[...]

PR #1082, opened 2021-04-16
README.md

oldphones68:
Removing `#enable-horizontal-tab-switcher` since it is removed from upstream.


csagan5:
As explained in #1077 it will be fixed on next release. The README will stay like this until that.

oldphones68:
Ok!

PR #1098, opened 2021-04-20
fix update ui for pac url config

### LINKED ISSUE #1072: Proxy not being saved, and abnormal behavior
### Bromite version

Version: `90.0.4430.74`
Arch: ``arm64`
Android version: `10`
Device model: `SM-N960F`

### Flags

No

Yes

No

Yes

No

### Describe the bug

PAC URL option is reset to "system default" on Proxy Configuration settings.

Steps to reproduce the bug:
1. Go to Settings -> Privacy -> Proxy configuration
2. Select 'Use PAC URL'
3. Enter usual PAC URL, for example https://antizapret.prostovpn.org/proxy.pac or https://raw.githubusercontent.com/anticensority/generated-pac-scripts/master/anticensority.pac
4. Click 'Apply'

### Expected behavior

Proxy settings are applied. PAC is used.

Also, I'd like to add the following: the proxy is saved if, after having entered the PAC URL, you close the proxy setting window by clicking the X on the upper left corner.
Re-entering proxy setting will have the screen resetted, i. e.: no URL visible on the PAC URL bar, and the 'use system default' voice is selected. 

The bug seems still present in 90.0.4430.59

_Originally posted by @ezellohar in https://github.com/bromite/bromite/issues/908#issuecomment-820167464_

uazo:
fixes #1072 

the problem was not so much in saving the setting, but in rendering it in the ui, I can only think that the bug has been present for some time.

### DIFF
--- a/build/patches/Add-a-proxy-configuration-page.patch
+++ b/build/patches/Add-a-proxy-configuration-page.patch
@@ -563,14 +563,14 @@ new file mode 100644
 +    renderConfig_: function() {
 +      if (this.currentConfig.auto_detect) {
 +        $(kIdModeAutoDetect).checked = true;
-+      } else if (this.currentConfig.rules.type == "none") {
-+        $(kIdModeEmpty).checked = true;
-+      } else if (this.currentConfig.rules.type == "direct") {
-+        $(kIdModeDirect).checked = true;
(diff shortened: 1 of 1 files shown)


csagan5:
Have you tested both paths? e.g. going from one selected to the other?

uazo:
Yes, absolutely.

Issue #1047, opened 2021-04-07
Custom tab intents issue 


J-khatri:
### Bromite version

Version: `90.0.4430.59`
 `arm64`
Android version: (example: `Android 9`)
Device model: (example: `oneplus 6`)

### Flags

These are the flags changed under `chrome://flags`:

### Is this bug about the SystemWebView?

No

### Is the bug reproducible with latest version?

Yes

### Can the bug be reproduced with corresponding Chromium version?

No

### Allow ads for the site; is the bug caused by the adblocker?
No

### Describe the bug

If allow custom tab intents is enabled .. browser crashes if try to open any link received in msg or mail 

If custom tab intents is disabled ... everything works fine 

### Steps to reproduce the bug

Steps to reproduce the bug:
1. Go to '...settings'
2. Click on 'privacy and security'
3. Scroll down to 'allow custom tab intents' and enable it 
4. Load any links from msg or mail or telegram etc...

### Expected behavior

It should open link normally

frankjrp:
@J-khatri for me it's working fine with `allow custom tab intents` enabled.

I've tested with Gmail and Telegram.

Device: Samsung Galaxy S8 (SM-G950FD)
Android: 9

** Maybe you have some other flag enabled that's causing conflict with the custom tab intents.

J-khatri:
I've only disabled "color gamut" in flags 

I don't think it has anything to do with the custom tab intents 

Not sure though

csagan5:
@J-khatri you did not mention any changed flags in the issue though.

J-khatri:
I think it's  not needed , I've changed that flag after reporting that but... And there is no any relation if it's enabled or disabled ( color gamut )

uazo:
I can't reproduce this problem.. again automatic extraction of logs would be useful...
@J-khatri are you able to send us the logs? 
do you have a rooted phone?

J-khatri:
> I can't reproduce this problem.. again automatic extraction of logs would be useful...
> @J-khatri are you able to send us the logs? 
> do you have a rooted phone?
> 
> 

No ..I'm not rooted

534n0:
I can confirm this bug too and I'm not altering any flags. It works just fine before.

Version: 90.0.4430.59
arm64
Android version: Android 9
Device model: Xiaomi Redmi Note 7

douchesfeedboard:
Crashing with custom tabs...could it be related to the issue https://github.com/bromite/bromite/issues/1051 ?
[...]

Issue #1030, opened 2021-03-28
Closing Incognito Tab closes New Tabs under Normal Tabs and redirects to 1st tab


pixincreate:
### Bromite version

Version: `93.0.4577.83`
Arch: `arm64`
Android version: 10
Device model: Nokia 6.1 Plus (DRG)

### Flags
These are the flags changed under `chrome://flags`:
- Enable IPv6 Probing (_Enabled_)
- Parallel Downloading (_Enabled_)
- Download auto-resumption native (_Enabled_)
- tab engagement metrics (_Disabled_)
- Support for HTTPSSVC records in DNS (_Disabled_)

### Is this bug about the SystemWebView?
No

### Is the bug reproducible with latest version?
Yes

### Can the bug be reproduced with corresponding Chromium version?
No

### Allow ads for the site; is the bug caused by the adblocker?
No

### Is the bug a crash?
No

### Describe the bug
- I've some tabs opened for reference. Along with that I usually keep a newtab in the forefront of other previous tabs.  
- For random things, I use incognito tabs.  
- When I close them, once my research is done, it brings me to the 1st normal tab closing the newtab which'd be in the forefront.
- I literally feel it annoying as I'm supposed to open the newtab again every time I close an incognito tab, which doesn't happen in chromium.

### Steps to reproduce the bug

Steps to reproduce the bug:
1. Have some tabs open under normal tabs
2. Have **NewTab** opened at last.
3. Open an **incognito tab** and do your research. 
4. Close the incognito tab my any means. Either close from the notification bar/clear the RAM, or systematically close the incognito tab by clicking tab and **x** button(this doesn't close the newtab but highlights the first tab).

### Expected behavior
- It should not touch the newtab i.e., opened. 
- When I close the incognito tab, you should be redirected to the last tab atleast, not the beginning of the tab(1st tab).

### Recording
**Incognito recording has been turned black by default.**   
Here's the recording.(Initially, it showed no bug like as I said before, but on doing the same for the 2nd time, it happened.)  

**LINK:**   
https://user-images.githubusercontent.com/69745008/112744179-5bfe9780-8fbb-11eb-9519-5e94c6988081.mp4  
(Above link didn't seem to work, so, compressed the video into a zip uploaded below.)  
[Bug in Bromite.zip](https://github.com/bromite/bromite/files/6216933/Bug.in.Bromite.zip)

uazo:
do you have the new homepage option enabled with "custom web address" to nothing?

[...]

PR #1105, opened 2021-04-23
enable crash reporting ui, allow save report to primary download folder

### LINKED ISSUE #944: Check a way to automatically extract log information

No

I don't know, but I'll check

Sometimes it is necessary to ask users for more information about issues that do not occur in our test environments.
Since not everyone may be able to extract the technical information necessary to understand the extent of the problem, and, considering that the user is not always able to understand the complexity of development on a browser in such a heterogeneous environment as android, I think it is a priority to expose a specific function that is easily accessible and well documented, so that it is possible for us simple developers to better frame the problem without too much trouble.

None, for now.

uazo:
this patch simply re-enables chromium's standard interface accessible via chrome://crashes/ and modifies the behavior of the upload button by saving instead the report to the download folder.
I deliberately didn't add anything more before talking about it.

No upload is enabled and saving is only active with user intervention.

It lacks:
- Verify request for write permissions (must already be enabled)
- Not ready for A11 due to how the export path is constructed

considering that the file is the one that is sent to google's crashpad servers, it should be explored how to use what is produced to extract the necessary information.
here http://dev.chromium.org/developers/crash-reports at "Symbolizing Minidumps" some insights

fixes https://github.com/bromite/bromite/issues/944


build/patches/Enable-crash-reporting-ui.patch
```diff
@@ -0,0 +1,206 @@
...
+-                    String uploadId = responseContent != null ? responseContent : "unknown";
+-                    return Result.success(uploadId);
+-                } else {
+-                    // Return the remote error code and message.
+-                    return Result.uploadError(responseCode, connection.getResponseMessage());
+-                }
+-            } finally {
+-                connection.disconnect();
+-            }
++
```
csagan5, line 177: What is `fileToUpload`? The zip file, the content of the zip file or the `upload_file_minidump` alone?

uazo: The zip file, exactly this https://github.com/bromite/bromite/issues/944#issuecomment-825598966

build/patches/Enable-crash-reporting-ui.patch
```diff
@@ -0,0 +1,206 @@
...
+         }
+diff --git a/components/upload_list/upload_list.cc b/components/upload_list/upload_list.cc
+--- a/components/upload_list/upload_list.cc
++++ b/components/upload_list/upload_list.cc
+@@ -20,7 +20,7 @@ namespace {
+ // chrome://webrtc-logs and the feedback UI. See https://crbug.com/972526.
+ constexpr base::TaskTraits kLoadingTaskTraits = {
+     base::MayBlock(), base::TaskPriority::USER_BLOCKING,
+-    base::TaskShutdownBehavior::CONTINUE_ON_SHUTDOWN};
++    base::TaskShutdownBehavior::SKIP_ON_SHUTDOWN};
```
uazo, line 200: this is bad, and needs to be reviewed, although not particularly necessary right away: it happens that without it, a dcheck fails because somewhere it requires a singleton (not initialized) that it cannot generate in a (background) thread (the one IO that retrieves the crash list on file) because it would be temporary and _could_ lead to unwanted crashes.
I tried to look for it either, but I have to recompile everything with special flags that allow me to know the type of a template at runtime (the class is in fact a Lazy<T> and I need T).
But as I said, nothing happens and it does not crash, simply chromium is telling us that this is not good, but it works, let's say, since it's being used a lot, the benefits outweigh the risks

uazo: this is the stack of that dcheck, really impossible to tell where it came from

```
Symbolizing stack using ABI: x86
Build fingerprint: 'Android/sdk_phone_x86_64/generic_x86_64:9/PSR1.180720.012/4923214:userdebug/test-keys'
Revision: '0'
pid: 13049, tid: 13117, name: ThreadPoolForeg  >>> org.chromium.chrome.stable <<<
signal 6 (SIGABRT), code -6 (SI_TKILL), fault addr --------
[FATAL:thread_restrictions.cc(207)] Check failed: !g_singleton_disallowed.Get().Get(). LazyInstance/Singleton is not allowed to be used on this thread. Most likely it's because this thread is not joinable (or the current task is running>
Abort message: '[FATAL:thread_restrictions.cc(207)] Check failed: !g_singleton_disallowed.Get().Get(). LazyInstance/Singleton is not allowed to be used on this thread. Most likely it's because this thread is not joinable (or the current>

Stack Trace:
  RELADDR   FUNCTION                   FILE:LINE
  00000b59  __kernel_vsyscall+9        /home/uazo/chromium/src/[vdso:eb55e000]
  0001fc08  syscall+40                 /system/lib/libc.so
  000321f3  abort+115                  /system/lib/libc.so
  07428a6a  offset 0x1b94000           /data/app/org.chromium.chrome.stable-6fQWTqIq06SqUKHnkvEPnQ==/libchrome.so
  0737fe1f  offset 0x1b94000           /data/app/org.chromium.chrome.stable-6fQWTqIq06SqUKHnkvEPnQ==/libchrome.so
  07380134  offset 0x1b94000           /data/app/org.chromium.chrome.stable-6fQWTqIq06SqUKHnkvEPnQ==/libchrome.so
  07367703  offset 0x1b94000           /data/app/org.chromium.chrome.stable-6fQWTqIq06SqUKHnkvEPnQ==/libchrome.so
  074185d3  offset 0x1b94000           /data/app/org.chromium.chrome.stable-6fQWTqIq06SqUKHnkvEPnQ==/libchrome.so
  076d2f6b  offset 0x1b94000           /data/app/org.chromium.chrome.stable-6fQWTqIq06SqUKHnkvEPnQ==/libchrome.so
  076d307b  offset 0x1b94000           /data/app/org.chromium.chrome.stable-6fQWTqIq06SqUKHnkvEPnQ==/libchrome.so
  076f9943  offset 0x1b94000           /data/app/org.chromium.chrome.stable-6fQWTqIq06SqUKHnkvEPnQ==/libchrome.so
  076f9857  offset 0x1b94000           /data/app/org.chromium.chrome.stable-6fQWTqIq06SqUKHnkvEPnQ==/libchrome.so
  076f9a80  offset 0x1b94000           /data/app/org.chromium.chrome.stable-6fQWTqIq06SqUKHnkvEPnQ==/libchrome.so
  072a60ef  offset 0x1b94000           /data/app/org.chromium.chrome.stable-6fQWTqIq06SqUKHnkvEPnQ==/libchrome.so
  072a5ea1  offset 0x1b94000           /data/app/org.chromium.chrome.stable-6fQWTqIq06SqUKHnkvEPnQ==/libchrome.so
  09770a1c  offset 0x1b94000           /data/app/org.chromium.chrome.stable-6fQWTqIq06SqUKHnkvEPnQ==/libchrome.so
  097709cc  offset 0x1b94000           /data/app/org.chromium.chrome.stable-6fQWTqIq06SqUKHnkvEPnQ==/libchrome.so
  09770981  offset 0x1b94000           /data/app/org.chromium.chrome.stable-6fQWTqIq06SqUKHnkvEPnQ==/libchrome.so
  09771217  offset 0x1b94000           /data/app/org.chromium.chrome.stable-6fQWTqIq06SqUKHnkvEPnQ==/libchrome.so
  09770f36  offset 0x1b94000           /data/app/org.chromium.chrome.stable-6fQWTqIq06SqUKHnkvEPnQ==/libchrome.so
  097710b3  offset 0x1b94000           /data/app/org.chromium.chrome.stable-6fQWTqIq06SqUKHnkvEPnQ==/libchrome.so
  09771079  offset 0x1b94000           /data/app/org.chromium.chrome.stable-6fQWTqIq06SqUKHnkvEPnQ==/libchrome.so
  0977104f  offset 0x1b94000           /data/app/org.chromium.chrome.stable-6fQWTqIq06SqUKHnkvEPnQ==/libchrome.so
  0377d545  offset 0x1b94000           /data/app/org.chromium.chrome.stable-6fQWTqIq06SqUKHnkvEPnQ==/libchrome.so
  074109b8  offset 0x1b94000           /data/app/org.chromium.chrome.stable-6fQWTqIq06SqUKHnkvEPnQ==/libchrome.so
  07410cfc  offset 0x1b94000           /data/app/org.chromium.chrome.stable-6fQWTqIq06SqUKHnkvEPnQ==/libchrome.so
  0377d545  offset 0x1b94000           /data/app/org.chromium.chrome.stable-6fQWTqIq06SqUKHnkvEPnQ==/libchrome.so
  073d43c0  offset 0x1b94000           /data/app/org.chromium.chrome.stable-6fQWTqIq06SqUKHnkvEPnQ==/libchrome.so
  07400445  offset 0x1b94000           /data/app/org.chromium.chrome.stable-6fQWTqIq06SqUKHnkvEPnQ==/libchrome.so
  074000d9  offset 0x1b94000           /data/app/org.chromium.chrome.stable-6fQWTqIq06SqUKHnkvEPnQ==/libchrome.so
  073ffc44  offset 0x1b94000           /data/app/org.chromium.chrome.stable-6fQWTqIq06SqUKHnkvEPnQ==/libchrome.so
  074366c1  offset 0x1b94000           /data/app/org.chromium.chrome.stable-6fQWTqIq06SqUKHnkvEPnQ==/libchrome.so
  073ff687  offset 0x1b94000           /data/app/org.chromium.chrome.stable-6fQWTqIq06SqUKHnkvEPnQ==/libchrome.so
  07409895  offset 0x1b94000           /data/app/org.chromium.chrome.stable-6fQWTqIq06SqUKHnkvEPnQ==/libchrome.so
  0740962f  offset 0x1b94000           /data/app/org.chromium.chrome.stable-6fQWTqIq06SqUKHnkvEPnQ==/libchrome.so
  0740951b  offset 0x1b94000           /data/app/org.chromium.chrome.stable-6fQWTqIq06SqUKHnkvEPnQ==/libchrome.so
  07436f29  offset 0x1b94000           /data/app/org.chromium.chrome.stable-6fQWTqIq06SqUKHnkvEPnQ==/libchrome.so
  0009cce5  __pthread_start(void*)+53  /system/lib/libc.so
  00033c1b  __start_thread+75          /system/lib/libc.so
```

build/patches/Enable-crash-reporting-ui.patch
csagan5: Ok, then it is not necessary to compress it again? Since it is already a zip file.

build/patches/Enable-crash-reporting-ui.patch
csagan5: > this is the stack of that dcheck, really impossible to tell where it came from

Then why not initialize this when the UI starts?

build/patches/Enable-crash-reporting-ui.patch
uazo: > Ok, then it is not necessary to compress it again?

sorry I didn't understand, no it's me with that code that creates the zip

build/patches/Enable-crash-reporting-ui.patch
uazo: > Then why not initialize this when the UI starts?

I think it is something related to upload, like a service or some component that is under `IsMetricsAndCrashReportingEnabled` that remains false and therefore does not instantiated at startup, but I don't understand what, the code is a bit complex to read (a bit tangled)

uazo:
https://github.com/bromite/bromite/pull/1105/commits/a80678a873404f60db75b4194f6a3b4f66c17b53 I added this because it is more correct

build/patches/Enable-crash-reporting-ui.patch
csagan5: Could it be due to some other patch in Bromite? There are at least 2 patches touching the code related to uploading crash reports.

build/patches/Enable-crash-reporting-ui.patch
csagan5: Then you should call the file `.gz`, it is not a `.zip`.

csagan5: Would it be possible to avoid the MIME container and use `ZipOutputStream` directly instead? Then we don't mess with MIME and gzip. It is not a problem for developers but .zip would allow users to inspect more easily what they are submitting.

uazo: > Would it be possible to avoid the MIME container

I don't think it's easily possible.
for the zip no problem

build/patches/Enable-crash-reporting-ui.patch
uazo: > Could it be due to some other patch in Bromite?

I do not think so.
i have to try to change the clang flags, now it's more complex because i use goma, but i try.

build/patches/Enable-crash-reporting-ui.patch
csagan5: I went to check how the `upload()` function is called:

1. what happens when multiple local minidumps are available?
2. are the internal minidump files deleted after user saves them?

uazo: > what happens when multiple local minidumps are available?

a picture is more than a thousand words
image

> are the internal minidump files deleted after user saves them?

no, they stay there
there is a whole self-cleaning system already active (even today) that deletes the files sent and not after a number of days

build/patches/Enable-crash-reporting-ui.patch
uazo: i found the problem, i think chrome is also affected.

the failed dcheck `AssertSingletonAllowed` is the fault of `ui::FormatBytes()` https://source.chromium.org/chromium/chromium/src/+/master:chrome/browser/crash_upload_list/crash_upload_list_android.cc;l=140;drc=221e331b49dfefadbc6fa40b0c68e6f97606d0b3;bpv=1;bpt=1

because it ultimately goes to use the icu library as a singleton.
https://source.chromium.org/chromium/chromium/src/+/master:base/i18n/number_formatting.cc;l=68;drc=221e331b49dfefadbc6fa40b0c68e6f97606d0b3;bpv=1;bpt=1

I preferred to delete that call, also because the file_size field is not used anywhere in the code. the alternative is to initialize the library before the thread fork, indeed, it seems strange to me that it is not already .

---
I write it here maybe I need it other times, to retrieve the name of T without changing clang flags:
```
  template <typename T> const char *name()
  {
      return __PRETTY_FUNCTION__;
  }

```

build/patches/Enable-crash-reporting-ui.patch
uazo: added `ZipOutputStream` in https://github.com/bromite/bromite/pull/1105/commits/8fda0c2fb16dfc9b5381775596b8bd7191fb54b0

csagan5:
> * Verify request for write permissions (must already be enabled)
> * Not ready for A11 due to how the export path is constructed

So after the above 2 are fixed, this is ready?

uazo:
yes, but thinking about using it as it is, for https://github.com/bromite/bromite/issues/1047 and https://github.com/bromite/bromite/issues/1030 , so let's see if it's worth it?

csagan5:
> yes, but thinking about using it as it is, for #1047 and #1030 , so let's see if it's worth it?

Then some user on A11 will not be able to use it; I am not in favor of releasing it with known issues.

uazo:
> I am not in favor of releasing it with known issues.

the fact is that it seems complex to me, I have to retrieve an android activity where I think it is not available, it was for that, but I look at it better.
I wanted to evaluate if it is already useless for the a9 and a10 before putting my hands back on it

csagan5:
@uazo in such case then we should display some error in the UI in case the request for write permissions has not been asked or the user is running on A11, so that users will not report a bug.

[A11 is catching up fast](https://www.appbrain.com/stats/top-android-sdk-versions), and there might be bugs that only happen on A11 as we have seen in the past.

uazo:
i completed the patch from your suggestions, now it uses download ui and then works on A11 (tested in emulator). Permissions are automatically requested if needed.
Technically I have allowed the browser process to access the crash folder, in which I produce a .zip file this time in native, the java part has been deleted.
All files are deleted after 5 days (it was 30 in chromium).

build/patches/Enable-crash-reporting-ui.patch
```diff
@@ -3,18 +3,49 @@ Date: Fri, 23 Apr 2021 18:41:48 +0000
...
+   const char pending_uploads[] = ".dmp";
+   const char skipped_uploads[] = ".skipped";
+   const char manually_forced_uploads[] = ".forced";
++  const char zipped_uploads[] = ".zip";
+ 
+   base::FileEnumerator files(upload_log_path().DirName(), false,
+                              base::FileEnumerator::FILES);
+   for (base::FilePath file = files.Next(); !file.empty(); file = files.Next()) {
+     UploadList::UploadInfo::State upload_state;
++    if (file.value().find(zipped_uploads) != std::string::npos) {
```
csagan5, line 33: In which cases the files would not be `.zip`?

csagan5: I guess those would be the `.dmp` files?

build/patches/Enable-crash-reporting-ui.patch
```diff
@@ -27,13 +58,101 @@ diff --git a/chrome/browser/crash_upload_list/crash_upload_list_android.cc b/chr
...
++  content::WebContents* web_contents_;
+ 
+   DISALLOW_COPY_AND_ASSIGN(CrashesDOMHandler);
+ };
+ 
+-CrashesDOMHandler::CrashesDOMHandler()
+-    : list_available_(false), first_load_(true) {
++CrashesDOMHandler::CrashesDOMHandler(content::WebContents* web_contents)
++    : list_available_(false), first_load_(true),
++      web_contents_(web_contents) {
```
csagan5, line 151: I am concerned about the `web_contents`: what if it has been destroyed by the crash already?

csagan5: (Perhaps that is why they were doing this in Java?)

build/patches/Enable-crash-reporting-ui.patch
```diff
@@ -43,30 +162,131 @@ diff --git a/chrome/browser/ui/webui/crashes_ui.cc b/chrome/browser/ui/webui/cra
...
++  // foo.skipped.try0), remove it
++  base::FilePath zip_file = crash_file_path;
++  while (zip_file != zip_file.RemoveExtension())
++    zip_file = zip_file.RemoveExtension();
++
++  // make zip file name, like "ec708a7b-cb17-44e7-8dae-e32f6c45cb8c.zip"
++  zip_file = upload_log_path.Append(zip_file.BaseName())
++                            .AddExtensionASCII(".zip");
++  // if exists return it
++  if (base::PathExists(zip_file))
```
csagan5, line 203: Why would it already exist?

build/patches/Enable-crash-reporting-ui.patch
```diff
@@ -111,6 +331,46 @@ diff --git a/components/crash_strings.grdp b/components/crash_strings.grdp
...
    </message>
  </grit-part>
+diff --git a/components/minidump_uploader/android/java/src/org/chromium/components/minidump_uploader/CrashFileManager.java b/components/minidump_uploader/android/java/src/org/chromium/components/minidump_uploader/CrashFileManager.java
+--- a/components/minidump_uploader/android/java/src/org/chromium/components/minidump_uploader/CrashFileManager.java
++++ b/components/minidump_uploader/android/java/src/org/chromium/components/minidump_uploader/CrashFileManager.java
+@@ -107,6 +107,8 @@ public class CrashFileManager {
+ 
+     private static final Pattern TMP_PATTERN = Pattern.compile("\\.tmp\\z");
+ 
++    private static final String UPLOADED_MINIDUMP_ZIP_SUFFIX = ".zip";
```
csagan5, line 341: There is not really any upload, shouldn't we call it `SAVED_MINIDUMP_ZIP_SUFFIX`?

build/patches/Enable-crash-reporting-ui.patch
```diff
@@ -111,6 +331,46 @@ diff --git a/components/crash_strings.grdp b/components/crash_strings.grdp
...
+-            deleteFile(f);
+-        }
+         for (File f : getAllTempFiles()) {
+             deleteFile(f);
+         }
+@@ -586,6 +585,8 @@ public class CrashFileManager {
+                     && !f.getName().contains(UPLOAD_FORCED_MINIDUMP_SUFFIX)) {
+                 continue;
+             }
++            if (f.getName().contains(UPLOADED_MINIDUMP_ZIP_SUFFIX))
```
csagan5, line 369: It should not contain, it should end with that.

build/patches/Enable-crash-reporting-ui.patch
```diff
@@ -192,21 +441,82 @@ diff --git a/components/minidump_uploader/android/java/src/org/chromium/componen
...
+     }
+diff --git a/components/upload_list/text_log_upload_list.cc b/components/upload_list/text_log_upload_list.cc
+--- a/components/upload_list/text_log_upload_list.cc
++++ b/components/upload_list/text_log_upload_list.cc
+@@ -27,7 +27,7 @@ constexpr char kJsonLogKeyState[] = "state";
+ constexpr char kJsonLogKeySource[] = "source";
+ 
+ std::vector<std::string> SplitIntoLines(const std::string& file_contents) {
+-  return base::SplitString(file_contents, base::kWhitespaceASCII,
++  return base::SplitString(file_contents, "\n",
```
csagan5, line 459: How come this change?

build/patches/Enable-crash-reporting-ui.patch
uazo: in android the upload status is saved by the extension. this is a real example.

image

in `/cache/Crash Reports` the files to be uploaded are the files with `* .dmp *` in the form `<guid> .dmp <id> .try <number>`
I produce the .zip file removing all extensions, as in the example "73d35758-26ea-441f-9ab6-97ff7d3c8afa.zip".
uploaded files are listed in uploads.log and take the extension `*.forced*`

build/patches/Enable-crash-reporting-ui.patch
uazo: as the download is done from the ui download on which I have no feedback, it may happen that the user cancels the download.
in this case, I still allow to retry the download and in that case the zip is already present.

build/patches/Enable-crash-reporting-ui.patch
```diff
@@ -43,30 +162,131 @@ diff --git a/chrome/browser/ui/webui/crashes_ui.cc b/chrome/browser/ui/webui/cra
...
++    LOG(ERROR) << "Crash report: cannot create temp directory";
++    return std::string();
++  }
++
++  // temporary file for zip content
++  base::FilePath dest_file = temp_dir.GetPath().Append(crash_file_path.BaseName());
++
++  // make zip file
++  if (base::CopyFile(crash_file_path, dest_file)) {
++    if (zip::Zip(temp_dir.GetPath(), zip_file, false)) {
```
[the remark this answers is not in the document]
uazo: I create a temporary directory and put all the files I want in the zip.
that one called `zip::Zip()` is the only one I found in all the native side chromium code.
it needs a source folder and a destination file, there are no other methods

build/patches/Enable-crash-reporting-ui.patch
uazo: Yes you are right.

build/patches/Enable-crash-reporting-ui.patch
uazo: yes, you are right, although it doesn't matter

build/patches/Enable-crash-reporting-ui.patch
uazo: practically in the `uploads.log` file there are 3 elements divided by "," arranged on lines

`<uploadtime>,<local_id>,<capturetime>`

in local_id I put the path which have spaces ..

```
1620458520,/data/user/0/org.chromium.chrome.stable/cache/Crash Reports/73d35758-26ea-441f-9ab6-97ff7d3c8afa.forced30226.try0,97ff7d3c8afa
```

I could remove the path and not touch that, for the code currently the path is useless

congratulations, nice review

uazo:
I fix the last things and in the meantime I improve the html of the page that I forgot to do

build/patches/Enable-crash-reporting-ui.patch
uazo: > I don't know why that code is in java rather than native...

ah, I understand.
because it uses Android services, so it can send the log even without chromium active

uazo:
> So after the above 2 are fixed, this is ready?

:) I was thinking... the "extract now" button is missing... ahhhhh

build/patches/Enable-crash-reporting-ui.patch
csagan5: Can you update the comment to explain this?

build/patches/Enable-crash-reporting-ui.patch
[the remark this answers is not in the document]
csagan5: There is `zip::ZipFiles`

build/patches/Enable-crash-reporting-ui.patch
csagan5: It would matter if you would write an unit test for it.

build/patches/Enable-crash-reporting-ui.patch
csagan5: Why don't we store the relative path there? The prefix `/data/user/0/org.chromium.chrome.stable/cache/Crash Reports` never changes; reason: I am concerned it might break somewhere else.

csagan5: > congratulations, nice review

You are welcome! Thanks. I do not think I am doing my best but trying.

csagan5:
What would "extract now" do?

uazo:
> What would "extract now" do?

to generate the log extraction, we also need it without any crash

uazo:
see how they extract the logcat!

```
        Process p = Runtime.getRuntime().exec("logcat -d");
        BufferedReader reader = new BufferedReader(new InputStreamReader(p.getInputStream()));
        LinkedList<String> rawLogcat = new LinkedList<>();
        Integer exitValue = null;
```

I never thought about doing it like this ...

csagan5:
> Process p = Runtime.getRuntime().exec("logcat -d");

Wow. Why does the browser even have permissions to run programs...looks like a security vulnerability to me. I checked and there is another use in the `DexFixer` class; can all the apps run shell commands or only Chromium because of some granted permission?

build/patches/Enable-crash-reporting-ui.patch
uazo: fixed in https://github.com/bromite/bromite/pull/1105/commits/2eacdb9cfab05dd5fcbee058ce545fdd71253fe8

build/patches/Enable-crash-reporting-ui.patch
uazo: fixed in https://github.com/bromite/bromite/pull/1105/commits/2eacdb9cfab05dd5fcbee058ce545fdd71253fe8

build/patches/Enable-crash-reporting-ui.patch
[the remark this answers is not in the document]
uazo: no need to change it, if you look in the chromium code they all do exactly as I wrote

build/patches/Enable-crash-reporting-ui.patch
uazo: yes, I have to try

uazo:
I improved the html, fixed what you wrote and added buttons to delete everything and to generate a new file

build/patches/Enable-crash-reporting-ui.patch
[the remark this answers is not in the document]
csagan5: I am not reviewing Chromium code, I am reviewing the code in this patch: is the file copy necessary?

build/patches/Enable-crash-reporting-ui.patch
csagan5: Can you use the C++ equivalend of `endsWith`?

build/patches/Enable-crash-reporting-ui.patch
```diff
@@ -0,0 +1,872 @@
...
+     id = id.substr(pos + 1);
++    // Since current thread is an IO thread
++    // to avoid failed DCHECK ThreadRestrictions::AssertSingletonAllowed()
++    // remove ui::FormatBytes(): dcheck fail because it use base::FormatDouble()
++    // and LazyInstance<NumberFormatWrapper>::DestructorAtExit().
++    // also "upload.file_size" is unused.
++    std::u16string file_size_string;
+     UploadList::UploadInfo upload(id, info.creation_time, upload_state,
+-                                  ui::FormatBytes(file_size));
++                                  file_size_string /*ui::FormatBytes(file_size)*/);
```
csagan5, line 113: I don't think we need to keep commented code here.
```suggestion
+                                  "n/a");
```

build/patches/Enable-crash-reporting-ui.patch
```diff
@@ -0,0 +1,872 @@
...
++ */
++ function requestNewExtraction() {
++  chrome.send('requestNewExtraction');
++
++  // show spinner
++  $('spinner').hidden = false;
++
++  // Trigger a refresh in 3 seconds.  Clear any previous requests.
++  clearTimeout(refreshCrashListId);
++  refreshCrashListId = setTimeout(requestCrashes, 3000);
```
csagan5, line 614: How come here it's 3 seconds and in `requestCrashUpload` it's 5 seconds?

build/patches/Enable-crash-reporting-ui.patch
```diff
@@ -0,0 +1,872 @@
...
+-    private static final int MAX_CRASH_REPORT_AGE_IN_DAYS = 30;
++    private static final int MAX_CRASH_REPORT_AGE_IN_DAYS = 5;
+ 
+     // The maximum number of non-uploaded crashes to copy to the crash reports directory. The
+     // difference between this value and MAX_CRASH_REPORTS_TO_KEEP is that TO_KEEP is only checked
+@@ -460,9 +462,6 @@ public class CrashFileManager {
+     }
+ 
+     public void cleanOutAllNonFreshMinidumpFiles() {
+-        for (File f : getAllUploadedFiles()) {
```
csagan5, line 710: Do I understand correctly that the deletion is done elsewhere?

build/patches/Enable-crash-reporting-ui.patch
```diff
@@ -0,0 +1,872 @@
...
+-            deleteFile(f);
+-        }
+         for (File f : getAllTempFiles()) {
+             deleteFile(f);
+         }
+@@ -586,6 +585,8 @@ public class CrashFileManager {
+                     && !f.getName().contains(UPLOAD_FORCED_MINIDUMP_SUFFIX)) {
+                 continue;
+             }
++            if (f.getName().endsWith(SAVED_MINIDUMP_ZIP_SUFFIX))
```
csagan5, line 720: Perhaps a `!` is missing here?

build/patches/Enable-crash-reporting-ui.patch
```diff
@@ -0,0 +1,872 @@
...
++      file_path(upload_info.file_path) {}
+ 
+ UploadList::UploadInfo::~UploadInfo() = default;
+ 
+@@ -108,6 +109,11 @@ void UploadList::RequestSingleUpload(const std::string& local_id) {
+   NOTREACHED();
+ }
+ 
++void UploadList::RequestNewExtraction() {
++  // only available for Android.
```
csagan5, line 824: But this patch is for Android?

build/patches/Enable-crash-reporting-ui.patch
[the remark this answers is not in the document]
uazo: >  is the file copy necessary?

yes, there is no other way

> There is zip::ZipFiles

it's under a if-define and never used

```
#if defined(OS_POSIX)
bool ZipFiles(const base::FilePath& src_dir,
              const std::vector<base::FilePath>& src_relative_paths,
              int dest_fd) {
```

build/patches/Enable-crash-reporting-ui.patch
```diff
@@ -0,0 +1,872 @@
...
+-  display: flex;
+ }
+ 
+ #countBanner #crashesCount {
+@@ -81,3 +81,65 @@ html[dir=rtl] h1 {
+ .notUploaded {
+   color: #a0a0a0;
+ }
++
++label {
```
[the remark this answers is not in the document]
uazo: to arrange the page like this

test

build/patches/Enable-crash-reporting-ui.patch
uazo: no motivation, is that in my tests 5 seconds seemed too much!

build/patches/Enable-crash-reporting-ui.patch
uazo: that method is virtual, override is in `crash_upload_list_android.cc
`

build/patches/Enable-crash-reporting-ui.patch
uazo: fixed in https://github.com/bromite/bromite/pull/1105/commits/de5dacb811773f6d07423cec1e97e4689e85e7bc

build/patches/Enable-crash-reporting-ui.patch
uazo: fixed in https://github.com/bromite/bromite/pull/1105/commits/de5dacb811773f6d07423cec1e97e4689e85e7bc

build/patches/Enable-crash-reporting-ui.patch
uazo: fixed in https://github.com/bromite/bromite/pull/1105/commits/de5dacb811773f6d07423cec1e97e4689e85e7bc

build/patches/Enable-crash-reporting-ui.patch
uazo: add comment in https://github.com/bromite/bromite/pull/1105/commits/de5dacb811773f6d07423cec1e97e4689e85e7bc

build/patches/Enable-crash-reporting-ui.patch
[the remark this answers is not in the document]
csagan5: It does not generate a crash though; shouldn't it better be called `Generate report`?

build/patches/Enable-crash-reporting-ui.patch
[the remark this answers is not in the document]
uazo: > OS_POSIX is defined for Android.

ah, ok, I didn't know

>> yes, there is no other way

> Why?

maybe I don't understand the matter.
this is the header
```
// Convenience method for callers who don't need to set up the filter callback.
// If |include_hidden_files| is true, files starting with "." are included.
// Otherwise they are omitted.
bool Zip(const base::FilePath& src_dir, const base::FilePath& dest_file,
         bool include_hidden_files);

#if defined(OS_POSIX)
// Zips files listed in |src_relative_paths| to destination specified by file
// descriptor |dest_fd|, without taking ownership of |dest_fd|. The paths listed
// in |src_relative_paths| are relative to the |src_dir| and will be used as the
// file names in the created zip file. All source paths must be under |src_dir|
// in the file system hierarchy.
bool ZipFiles(const base::FilePath& src_dir,
              const std::vector<base::FilePath>& src_relative_paths,
              int dest_fd);
#endif  // defined(OS_POSIX)
```
unlike the other, `ZipFiles` is not used anywhere in the code
image

if I can I prefer code already used, even just to see how to use it.
what I do not understand?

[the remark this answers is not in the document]
csagan5: I know it is not used currently in Chromium, but it is available.

You have still not explained why the file copy is necessary.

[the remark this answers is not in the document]
uazo: > You have still not explained why the file copy is necessary.

also for the other patch you raised the same question.
why a copy in a temporary directory is bothering you, considering by the way that the entire chromium base code does that way? 

really, it is not the will not to do it, it is that I cannot understand the motivation.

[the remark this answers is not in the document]
csagan5: @uazo the file copy is not bothering me, your attitude is. You have not explained why the file copy is necessary and are being evasive.

Motivation: if it is not necessary it should not be there.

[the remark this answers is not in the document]
uazo: > You have not explained why the file copy is necessary and are being evasive.

evasive? I have already written you my reasons:

1. I have not found any other method to zip
2. It's used that way in all chromium
3. The function you indicated is used only in tests

permettimi in italiano: nello sviluppo di chromium mi baso sul framework che danno a disposizione, la ragione tecnica del fatto che zip() richieda una cartella in input non la conosco, ma quel metodo lo richiede, ed è fatto così. 

> Motivation: if it is not necessary it should not be there.

zipping the crash file was my proposal to remedy any encoding problems once posted on github.
if you think it is not a problem, we can keep the file unzipped.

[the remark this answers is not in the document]
csagan5: > 1. I have not found any other method to zip

I have told you that there is another method.

> 2\. It's used that way in all chromium

It does not matter.

> 3\. The function you indicated is used only in tests

It does not matter.

You have replied:
> > is the file copy necessary?
> 
> yes, there is no other way

And I showed you another way. The fact that it is done like that in Chromium or the method is currently used only in tests does not mean that there is no other way, that is your conclusion that I do not agree with.

If you do not want to change it, no worries, I will do it later.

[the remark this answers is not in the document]
uazo: >  The fact that it is done like that in Chromium or the method is currently used only in tests does not mean that there is **no other way**

that answer I had given before knowing that OS_POSIX is active.
later, checking, I saw that it is only used in tests and we have avoided the use of test methods other times as well.

> If you do not want to change it, no worries, I will do it later.

it's not that, in my opinion it's not worth it, but of course do what you think.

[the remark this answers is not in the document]
gdm85: > we have avoided the use of test methods other times as well.

This is not a test method. It is a method currently used only in tests, it is different. It has not been developed for using it only in tests (it does not have "test" mentioned in the name or description).

csagan5:
I will release the current v91 patches, there is a problem with the user agent customization one; not urgent, but if you want to give a look meanwhile.

There are still more errors to fix, I just started the builds.

uazo:
> but if you want to give a look meanwhile.

ok i will check, would you please give me the chromium version you checked out?

csagan5:
> ok i will check, would you please give me the chromium version you checked out?

It is always in `build/RELEASE`: `91.0.4472.50`.

uazo:
> there is a problem with the user agent customization one;
> There are still more errors to fix, I just started the builds.

https://gist.github.com/uazo/fb8e4f0bd0e9795c8105b8d90c081e1e 
here the diff with the missing parts to be able to compile, tried on the emulator, it runs.

https://gist.github.com/uazo/fb8e4f0bd0e9795c8105b8d90c081e1e#file-v91-diff-L244-L260
this part if you haven't already solved it I'll see it tomorrow

uazo:
I also found some crashes
```
05-13 18:54:00.256  6300  6300 E AndroidRuntime: java.lang.NullPointerException: Attempt to invoke interface method 'int org.chromium.chrome.browser.tabmodel.TabModel.indexOf(org.chromium.chrome.browser.tab.Tab)' on a null object reference
05-13 18:54:00.256  6300  6300 E AndroidRuntime:        at org.chromium.chrome.browser.tabmodel.ChromeTabCreator.createNewTab(ChromeTabCreator.java:140)
05-13 18:54:00.256  6300  6300 E AndroidRuntime:        at org.chromium.chrome.browser.tabmodel.ChromeTabCreator.launchUrl(ChromeTabCreator.java:328)
05-13 18:54:00.256  6300  6300 E AndroidRuntime:        at org.chromium.chrome.browser.tabmodel.ChromeTabCreator.launchUrl(ChromeTabCreator.java:311)
05-13 18:54:00.256  6300  6300 E AndroidRuntime:        at org.chromium.chrome.browser.tabmodel.TabCreator.launchNTP(TabCreator.java:95)
05-13 18:54:00.256  6300  6300 E AndroidRuntime:        at org.chromium.chrome.browser.ChromeTabbedActivity.lambda$initializeToolbarManager$4$ChromeTabbedActivity(ChromeTabbedActivity.java:727)
05-13 18:54:00.256  6300  6300 E AndroidRuntime:        at org.chromium.chrome.browser.ChromeTabbedActivity$$Lambda$10.onClick(Unknown Source:2)
05-13 18:54:00.256  6300  6300 E AndroidRuntime:        at org.chromium.chrome.browser.toolbar.top.TabSwitcherModeTTPhone.onClick(TabSwitcherModeTTPhone.java:103)
```

```
05-13 18:58:11.791  6659  6659 D AndroidRuntime: Shutting down VM
05-13 18:58:11.792  6659  6659 E AndroidRuntime: FATAL EXCEPTION: main
05-13 18:58:11.792  6659  6659 E AndroidRuntime: Process: org.chromium.chrome.stable, PID: 6659
05-13 18:58:11.792  6659  6659 E AndroidRuntime: java.lang.RuntimeException: The AccountInfoService is not yet initialized!
05-13 18:58:11.792  6659  6659 E AndroidRuntime:        at org.chromium.components.signin.identitymanager.AccountInfoService.get(AccountInfoService.java:62)
05-13 18:58:11.792  6659  6659 E AndroidRuntime:        at org.chromium.chrome.browser.signin.services.ProfileDataCache.<init>(ProfileDataCache.java:115)
05-13 18:58:11.792  6659  6659 E AndroidRuntime:        at org.chromium.chrome.browser.signin.services.ProfileDataCache.createWithDefaultImageSizeAndNoBadge(ProfileDataCache.java:125)
05-13 18:58:11.792  6659  6659 E AndroidRuntime:        at org.chromium.chrome.browser.ntp.RecentTabsManager.<init>(RecentTabsManager.java:103)
05-13 18:58:11.792  6659  6659 E AndroidRuntime:        at org.chromium.chrome.browser.native_page.NativePageFactory$NativePageBuilder.buildRecentTabsPage(NativePageFactory.java:130)
05-13 18:58:11.792  6659  6659 E AndroidRuntime:        at org.chromium.chrome.browser.native_page.NativePageFactory.createNativePageForURL(NativePageFactory.java:186)
05-13 18:58:11.792  6659  6659 E AndroidRuntime:        at org.chromium.chrome.browser.native_page.NativePageFactory.createNativePage(NativePageFactory.java:159)
05-13 18:58:11.792  6659  6659 E AndroidRuntime:        at org.chromium.chrome.browser.TabbedModeTabDelegateFactory.createNativePage(TabbedModeTabDelegateFactory.java:121)
05-13 18:58:11.792  6659  6659 E AndroidRuntime:        at org.chromium.chrome.browser.tab.TabImpl.maybeShowNativePage(TabImpl.java:1199)
05-13 18:58:11.792  6659  6659 E AndroidRuntime:        at org.chromium.chrome.browser.tab.TabImpl.loadUrl(TabImpl.java:485)
05-13 18:58:11.792  6659  6659 E AndroidRuntime:        at org.chromium.chrome.browser.ChromeTabbedActivity.onMenuOrKeyboardAction(ChromeTabbedActivity.java:1894)
05-13 18:58:11.792  6659  6659 E AndroidRuntime:        at org.chromium.chrome.browser.app.ChromeActivity.onOptionsItemSelected(ChromeActivity.java:1560)
05-13 18:58:11.792  6659  6659 E AndroidRuntime:        at org.chromium.chrome.browser.ui.appmenu.AppMenuHandlerImpl.onOptionsItemSelected(AppMenuHandlerImpl.java:291)
05-13 18:58:11.792  6659  6659 E AndroidRuntime:        at org.chromium.chrome.browser.ui.appmenu.AppMenu.onItemClick(AppMenu.java:365)
05-13 18:58:11.792  6659  6659 E AndroidRuntime:        at org.chromium.chrome.browser.ui.appmenu.AppMenuAdapter.lambda$setupStandardMenuItemViewHolder$3$AppMenuAdapter(AppMenuAdapter.java:402)
05-13 18:58:11.792  6659  6659 E AndroidRuntime:        at org.chromium.chrome.browser.ui.appmenu.AppMenuAdapter$$Lambda$3.onClick(Unknown Source:4)

```

csagan5:
> https://gist.github.com/uazo/fb8e4f0bd0e9795c8105b8d90c081e1e
> here the diff with the missing parts to be able to compile, tried on the emulator, it runs.
> 
> https://gist.github.com/uazo/fb8e4f0bd0e9795c8105b8d90c081e1e#file-v91-diff-L244-L260
> this part if you haven't already solved it I'll see it tomorrow

Thanks, but I will continue on my own. Then I will see if there are crashes.

csagan5:
@uazo what is the fix for the user agent customization patch? I only need that one.

uazo:
>  what is the fix for the user agent customization patch?

simply this

```
diff --git a/chrome/browser/android/preferences/browser_prefs_android.cc b/chrome/browser/android/preferences/browser_prefs_android.cc
index 0443076237..d17ab74f5f 100644
--- a/chrome/browser/android/preferences/browser_prefs_android.cc
+++ b/chrome/browser/android/preferences/browser_prefs_android.cc
@@ -11,6 +11,7 @@
 #include "chrome/browser/webauthn/android/cable_module_android.h"
 #include "components/pref_registry/pref_registry_syncable.h"
 #include "components/prefs/pref_registry_simple.h"
+#include "chrome/common/pref_names.h"

 namespace android {

diff --git a/components/browser_ui/site_settings/android/java/src/org/chromium/components/browser_ui/site_settings/ContentSettingsResources.java b/components/browser_ui/site_settings/android/java/src/org/chromium/components/browser_ui/site_settings/ContentSettingsResources.java
index 27f5af4a60..59839dac12 100644
--- a/components/browser_ui/site_settings/android/java/src/org/chromium/components/browser_ui/site_settings/ContentSettingsResources.java
+++ b/components/browser_ui/site_settings/android/java/src/org/chromium/components/browser_ui/site_settings/ContentSettingsResources.java
@@ -251,7 +251,7 @@ public class ContentSettingsResources {
                             R.string.website_settings_category_vr_ask,
                             R.string.website_settings_category_vr_blocked));
             localMap.put(ContentSettingsType.TIMEZONE_OVERRIDE,
-                    new ResourceItem(R.drawable.web_asset, R.string.timezone_override_permission_title,
+                    new ResourceItem(R.drawable.web_asset, 0, R.string.timezone_override_permission_title,
                             ContentSettingValues.ALLOW, ContentSettingValues.BLOCK,
                             R.string.website_settings_category_timezone_override_custom,
                             R.string.website_settings_category_timezone_override_random));
diff --git a/content/browser/web_contents/web_contents_impl.cc b/content/browser/web_contents/web_contents_impl.cc
index 4f2974698f..b92aa85a36 100644
--- a/content/browser/web_contents/web_contents_impl.cc
+++ b/content/browser/web_contents/web_contents_impl.cc
@@ -2488,7 +2488,8 @@ const blink::web_pref::WebPreferences WebContentsImpl::ComputeWebPreferences() {
       prefs.viewport_meta_enabled = false;
   }

-  if (IsOverridingUserAgent() &&
+  if (GetController().GetVisibleEntry() &&
+      GetController().GetVisibleEntry()->GetIsOverridingUserAgent() &&
       !command_line.HasSwitch(switches::kDesktopModeViewportMetaEnabled)) {
     prefs.viewport_meta_enabled = false;
   }

```

uazo:
> Java.lang.NullPointerException: Attempt to invoke interface method 'int org.chromium.chrome.browser.tabmodel.TabModel.indexOf

for the first crash, this is the patch, the error is a bit sneaky
```
diff --git a/chrome/android/java/src/org/chromium/chrome/browser/ChromeTabbedActivity.java b/chrome/android/java/src/org/chromium/chrome/browser/ChromeTabbedActivity.java
index b96fef970a..706c8fa6b9 100644
--- a/chrome/android/java/src/org/chromium/chrome/browser/ChromeTabbedActivity.java
+++ b/chrome/android/java/src/org/chromium/chrome/browser/ChromeTabbedActivity.java
@@ -1721,7 +1721,7 @@ public class ChromeTabbedActivity extends ChromeActivity<ChromeActivityComponent
             public void onTabStateInitialized() {
                 if (!mCreatedTabOnStartup) return;

-                TabModel model = mTabModelSelectorImpl.getModel(false);
+                TabModel model = mTabModelSelectorImpl.getModel(startIncognito);
                 TasksUma.recordTasksUma(model);
             }
         });
```

this to solve the second
```
diff --git a/chrome/android/java/src/org/chromium/chrome/browser/ntp/RecentTabsManager.java b/chrome/android/java/src/org/chromium/chrome/browser/ntp/RecentTabsManager.java
index 7e06f86fdd..b25b6a6c2e 100644
--- a/chrome/android/java/src/org/chromium/chrome/browser/ntp/RecentTabsManager.java
+++ b/chrome/android/java/src/org/chromium/chrome/browser/ntp/RecentTabsManager.java
@@ -18,7 +18,6 @@ import org.chromium.chrome.browser.ntp.ForeignSessionHelper.ForeignSession;
 import org.chromium.chrome.browser.ntp.ForeignSessionHelper.ForeignSessionTab;
 import org.chromium.chrome.browser.profiles.Profile;
 import org.chromium.chrome.browser.signin.services.IdentityServicesProvider;
-import org.chromium.chrome.browser.signin.services.ProfileDataCache;
 import org.chromium.chrome.browser.signin.services.SigninManager;
 import org.chromium.chrome.browser.signin.services.SigninManager.SignInStateObserver;
 import org.chromium.chrome.browser.signin.ui.PersonalizedSigninPromoView;
@@ -40,7 +39,7 @@ import java.util.List;
  * Provides the domain logic and data for RecentTabsPage and RecentTabsRowAdapter.
  */
 public class RecentTabsManager implements
-                                          SignInStateObserver, ProfileDataCache.Observer,
+                                          SignInStateObserver,
                                           AccountsChangeObserver {
     /**
      * Implement this to receive updates when the page contents change.
@@ -78,8 +77,6 @@ public class RecentTabsManager implements
     private UpdatedCallback mUpdatedCallback;
     private boolean mIsDestroyed;

-    private final ProfileDataCache mProfileDataCache;
-
     /**
      * Create an RecentTabsManager to be used with RecentTabsPage and RecentTabsRowAdapter.
      *
@@ -100,8 +97,6 @@ public class RecentTabsManager implements
                 ? sRecentlyClosedTabManagerForTests
                 : new RecentlyClosedBridge(profile);

-        mProfileDataCache = ProfileDataCache.createWithDefaultImageSizeAndNoBadge(context);
-
         mRecentlyClosedTabManager.setTabsUpdatedRunnable(() -> {
             updateRecentlyClosedTabs();
             postUpdate();
@@ -122,7 +117,6 @@ public class RecentTabsManager implements
     public void destroy() {
         mIsDestroyed = true;

-        mProfileDataCache.removeObserver(this);
         AccountManagerFacadeProvider.getInstance().removeObserver(this);

         mFaviconHelper.destroy();
@@ -151,7 +145,6 @@ public class RecentTabsManager implements
     }

     private void registerObservers() {
-        mProfileDataCache.addObserver(this);
         AccountManagerFacadeProvider.getInstance().addObserver(this);
     }

@@ -376,12 +369,6 @@ public class RecentTabsManager implements
         update();
     }

-    // ProfileDataCache.Observer implementation.
-    @Override
-    public void onProfileDataUpdated(String accountEmail) {
-        update();
-    }
-
     private void postUpdate() {
         if (mUpdatedCallback != null) {
             mUpdatedCallback.onUpdated();
```

csagan5:
> > what is the fix for the user agent customization patch?
> 
> simply this

The second block is for the timezone customization patch, but I understand.

> for the first crash, this is the patch, the error is a bit sneaky

For both crashes you have not mentioned how to reproduce them...

csagan5:
@uazo this patch does not apply on v91

build/bromite_patches_list.txt
```diff
@@ -146,4 +146,5 @@ Make-all-favicon-requests-on-demand.patch
 Add-Alt-D-hotkey-to-focus-address-bar.patch
 Remove-offline-measurement-background-task.patch
 User-agent-customization.patch
+Enable-crash-reporting-ui.patch
```
csagan5, line 149:
```suggestion
Logcat-crash-reports-UI.patch
```

build/patches/Enable-crash-reporting-ui.patch
```diff
@@ -0,0 +1,872 @@
+From: uazo <uazo@users.noreply.github.com>
+Date: Fri, 23 Apr 2021 18:41:48 +0000
+Subject: Enable crash report ui
```
csagan5, line 3:
```suggestion
Subject: Logcat crash reports UI
```

build/patches/Enable-crash-reporting-ui.patch
```diff
@@ -0,0 +1,872 @@
...
+@@ -136,8 +149,15 @@ void CrashUploadListAndroid::LoadUnsuccessfulUploadList(
+     RecordUnsuccessfulUploadListState(
+         UnsuccessfulUploadListState::ADDING_AN_UPLOAD_ENTRY);
+     id = id.substr(pos + 1);
++    // Since current thread is an IO thread
++    // to avoid failed DCHECK ThreadRestrictions::AssertSingletonAllowed()
++    // remove ui::FormatBytes(): dcheck fail because it use base::FormatDouble()
++    // and LazyInstance<NumberFormatWrapper>::DestructorAtExit().
++    // also "upload.file_size" is unused.
++    std::u16string file_size_string;
```
csagan5, line 119:
```suggestion
+    std::u16string file_size_string = "n/a";
```

build/patches/Enable-crash-reporting-ui.patch
```diff
@@ -0,0 +1,872 @@
...
+   </message>
+   <message name="IDS_CRASH_UPLOAD_NOW_LINK_TEXT" desc="Link text for manual uploads of a crash report">
+-    Send now
++    Save now
++  </message>
++  <message name="IDS_CRASH_CLEAR_ALL_TEXT" desc="Link text for clear all crash files">
++    Clear all
++  </message>
++  <message name="IDS_CRASH_EXTRACT_NOW_TEXT" desc="Link text for manual generation of a crash report">
++    Generate now
```
csagan5, line 691:
```suggestion
+    Generate report
```

csagan5:
About the user agent patch: the patch has no changes in `components/browser_ui/site_settings/android/java/src/org/chromium/components/browser_ui/site_settings/ContentSettingsResources.java`

uazo:
> For both crashes you have not mentioned how to reproduce them...

the first by simply pressing "home" in always incognito.
the other by opening the "Recent Tabs"

> About the user agent patch

i checked with the current master, there is no problem with that patch

> this patch does not apply on v91

now I look at it

csagan5:
The UI looks garbled:
image

Can this be fixed?

Also the translation is imprecise: it says that there was a crash at the time I generated the report, but that is untrue. For now leave it like that, but it is misleading.

uazo:
> Can this be fixed?

absolutely yes, with some changes to the css,
I didn't try it with smaller screens..

> Also the translation is imprecise: it says that there was a crash at the time I generated the report, but that is untrue

it is because manual creation is not provided by the original code.
I see if it is easy to mark the file with something that ui can understand

uazo:
> I see if it is easy to mark the file with something that ui can understand

it's not that easy ..
I have no way to force the name of the file generated by `base::debug::DumpWithoutCrashing()` unless I also change crashpad ..
if you agree I would leave it like this.

the alternative is to modify the `IDS_CRASH_CAPTURE_TIME_FORMAT` entry by removing the word "crash" with "report"... what do you say?

```
 <message name="IDS_CRASH_CAPTURE_TIME_FORMAT" desc="Format for the time at which the crash report was captured, to display on chrome://crashes">
    Report from <ph name="CRASH_TIME">$1<ex>Tuesday, January 25, 2011 2:58:02 PM</ex></ph>
  </message>
```

csagan5:
That's ok, leave it like that.

csagan5:
Nice work; I will fix the zipping part and then this can be on next release already.

uazo:
perfect, we hope it works the first time!

I remind you that unfortunately to read the minidumps we need another 1.1gb of artifacts, made with
```
components/crash/content/tools/generate_breakpad_symbols.py --build-dir=out/x86 \
   --symbols-dir=/tmp/my_symbols/ --binary=out/x86/lib.unstripped/libchrome.so --clear --verbose
```

csagan5:
> I remind you that unfortunately to read the minidumps we need another 1.1gb of artifacts

Is it necessary only for `libchrome.so`?

uazo:
> Is it necessary only for libchrome.so?

I don't know, the documentation only references that library, but the chromium documentation is old in many places. must be seen with time

uazo:
I'm sorry, the ui is still garbled on some phones, it seems that https://github.com/bromite/bromite/pull/1105/commits/0177840cce181c1b550fe4057ec5a42f7de4ea78 is not conclusive , I fix it as soon as possible

csagan5:
The latest version has all the symbols.

uazo:
strange, how come to me it was 1gb while there is 400mb there?
I wouldn't want it to be compiled in debug... we'll see.

PR #1106, opened 2021-04-24
TrialParam: explicit denial to fetch

### LINKED ISSUE #1021: enable-ephemeral-tab-bottom-sheet flag options are not working as intended.
### Bromite version

Version: 89.0.4389.100
Arch: arm64
Android version: 10
Device model: motorola one action and motorola one hyper

### Flags

These are the flags changed under `chrome://flags`: Only the affected flag

No

Yes

No

No

### Is the bug a crash?
No

### Describe the bug
The flag enable-ephemeral-tab-bottom-sheet has six options: default, enabled, enabled at peek state, enabled at half state, enabled at full state and disabled. When it is enabled, enabled at peek state, enabled at half state or enabled at full state, Bromite appears to use only the enable at peek state option, ignoring the others. Chromium uses the correct option.

1. Go to chrome://flags
2. Search for ephemeral flag and choose one of the enabled options
3. Restart the browser
4. Go to www.folha.uol.com.br  and long touch a link
5. Choose preview page
6. See the bug

### Expected behavior

It is expected to see the preview page covering half or almost all space of the tab window when the flag is enabled at half state or enabled at full state, respectively.

This flag is useful because it can avoid the refresh of the first page when opening a link to a second page and then returning to the first. This is in some cases time-consuming and frequently returns to a different point in the original page. Furthermore it appears that pages load faster using the preview.
Page auto refresh reduces utility because it closes the preview page also.

### Screenshots

Bromite - flag enabled at full state

Chromium - flag enabled at full state

uazo:
rehabilitation of `getFieldTrialParamByFeature()` and explicit denial to fetch
fixes #1021

### DIFF
--- a/build/patches/Disable-fetching-of-all-field-trials.patch
+++ b/build/patches/Disable-fetching-of-all-field-trials.patch
@@ -3,60 +3,9 @@ Date: Sun, 8 Jul 2018 18:16:34 +0200
 Subject: Disable fetching of all field trials
 
 ---
- .../browser/flags/ChromeFeatureList.java      | 19 ++++---------------
- .../variations/service/variations_service.cc  | 12 +-----------
- 2 files changed, 5 insertions(+), 26 deletions(-)
+ .../variations/service/variations_service.cc       | 14 ++------------
+ 1 file changed, 2 insertions(+), 12 deletions(-)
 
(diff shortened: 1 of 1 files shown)


csagan5:
I left a question in #1021

> And the defaults from `chrome://flags` lead to correct behaviour?

csagan5 (approved):

Issue #1051, opened 2021-04-08
Proxy configuration can't be open anymore with Settings menu if using always incognito mode


douchesfeedboard:
### Bromite version

Version: `90.0.4430.59`
Arch: `arm64`
Android version: `9.0`
Device model: TA-1033 (Nokia 6 2017)

### Flags

These are the flags changed under `chrome://flags`:

### Is this bug about the SystemWebView?

No

### Is the bug reproducible with latest version?

Yes

### Can the bug be reproduced with corresponding Chromium version?

No (this setting do not exist with Chromium)

### Allow ads for the site; is the bug caused by the adblocker?
No

### Is the bug a crash?

No

### Describe the bug

Proxy configuration can't be open anymore through "Settings" since the latest update 90.0.4430.59 if you have the settings 'Open links in incognito tabs always' activated. You get a white screen for a few seconds before going back to Privacy and security. 
But typing 'chrome://proxy/' manually in the urlbar still working.

### Steps to reproduce the bug

Steps to reproduce the bug:
1. Click on 'the three dots menu in the corner'
2. Click on 'Settings'
3. Click on 'Privacy and security.'
4. Check if 'Open links in incognito tabs always' is on or activate it.
5. Click on 'Proxy configuration'
6. Get a white screen and you're back in the 'Privacy and security' menu

### Expected behavior

A tab with the url 'chrome://proxy/' should be open when clicking on 'Proxy configuration' even when using 'Open links in incognito tabs always'

kokukyocho:
Hello @douchesfeedboard!

I am a bot :robot:

Thanks for submitting this issue!
I noticed that it is missing the template, please edit the issue to match the template for either a [bug](https://github.com/bromite/bromite/blob/master/.github/ISSUE_TEMPLATE/bug_report.md) or a [feature request](https://github.com/bromite/bromite/blob/master/.github/ISSUE_TEMPLATE/feature_request.md).

The issue will be automatically re-opened after that.

**Tip**: if you use GitHub in Desktop mode the template will be proposed to you when submitting issues.

csagan5:
I cannot reproduce this issue; please connect with [remote debugging](https://developer.chrome.com/docs/devtools/remote-debugging/) and check what is the problem on the Console.

douchesfeedboard:
Sorry, I don't have Chrome on my computer for the remote debugging.
But I did some experiments and found the issue is related to the "Open links in incognito tabs always" setting.
[...]

PR #1125, opened 2021-05-01
revert last changes + remove record incognito launch reason histogram

### LINKED ISSUE #1116: PWA crash in "Open links in incognito tabs always" mode
### Bromite version

Version: `90.0.4430.92`
Arch: `arm64`
Android version: `10`
Device model: `SM-A205G`

### Flags

These are the flags changed under `chrome://flags`: Reader Mode triggering

No

Yes

No

Yes

No

### Is the bug a crash?

Yes, this is the crash dump:
```
<I only have regular Windows at hand, so no grep, how do I do it?>
```

### Describe the bug

Trying to open a PWA while the "Open links in incognito tabs always" option is on will cause Bromite to crash.

I've seen other issues about this, but they were closed so I opened this one.

Steps to reproduce the bug:
1. Activate "Open links in incognito tabs always" in settings.
2. Install a PWA
3. Open the PWA

### Expected behavior

PWA opens as it normally does in non incognito mode.

uazo:
I checked every possible path, I think the best is simply to remove the histogram.
I put back the code that I had removed because the flag is active only for

`This feature is now used in Offline pages and Reader Mode.`

instead the patch also activates it for the custom tab. use this to reproduce
```
adb shell am start -a android.intent.action.VIEW -d http://www.stackoverflow.com --es android.support.customtabs.extra.SESSION "1"
```

checked https://github.com/bromite/bromite/issues/1051 it works

fix for #1116

### DIFF
--- a/build/patches/Add-an-always-incognito-mode.patch
+++ b/build/patches/Add-an-always-incognito-mode.patch
@@ -14,6 +14,8 @@ Enable incognito custom tabs and fix crashes for incognito/custom tab intents (c
  .../chrome/browser/app/ChromeActivity.java    |  4 +
  .../AppMenuPropertiesDelegateImpl.java        |  6 ++
  .../ChromeContextMenuPopulator.java           |  9 ++-
+ .../CustomTabActivityLifecycleUmaTracker.java |  5 --
+ .../CustomTabIntentDataProvider.java          |  5 +-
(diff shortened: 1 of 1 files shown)


csagan5:
Ok; in general preserving histogram functionality is not important, we do not do that in other patches.

PR #1166, opened 2021-05-24
fix saving tabs bookmarks folder

### LINKED ISSUE #1138: "Bookmark all tabs" to "Tabs collection" both temporary and non-exportable
### Bromite version

Version: `90.0.4430.204`
Arch: `arm64`
Android version: `9.0`
Device model: `Galaxy S10`

### Flags

No flags changed.

No

Yes

No

No

No

### Is the bug a crash?

No

### Describe the bug

When using `Bookmark all tabs`, the bookmarks are saved to the `Tabs collection` root folder in bookmark manager (this was a recent change, previously they would go into `Mobile bookmarks`). 
- __Bug:__ These `Tabs collection` bookmarks are not included when exporting bookmarks from the bookmark manager.
- __Bug:__ These `Tabs collection` bookmarks disappear when users close and restart Bromite.
- __Note:__ bookmarks in `Mobile bookmarks` are preserved across sessions and are included in exported html file).

Steps to reproduce the bug:

1. Select three dots menu.
2. Select `Bookmark all tabs`.
3. Select three dots menu.
4. Select `Bookmarks`. Verify that `Tabs collection` now contains the same number of bookmarks as open in Bromite.
5. Press down arrow to download, name file `bookmarks.html` and press `Export`. `bookmarks.html` should load and can verify by searching that bookmarks in `Tabs collection` are not included.
6. Press `X` to exit bookmarks manager if still open.
7. Select three dots menu and press `Exit`.
8. Re-open Bromite.
9. Select three dots menu.
10. Select `Bookmarks`. The `Tabs collection` is now empty even though bookmarks in `Mobile bookmarks` remain.

### Expected behavior

Expected that tabs bookmarked with `Bookmark all tabs` into `Tabs collection` would be preserved across sessions and be included in html file upon exporting bookmarks.

uazo:
Fix for #1138

### DIFF
--- a/build/patches/Add-menu-item-to-bookmark-all-tabs.patch
+++ b/build/patches/Add-menu-item-to-bookmark-all-tabs.patch
@@ -12,13 +12,16 @@ Subject: Add menu item to bookmark all tabs
  .../bookmarks/chrome_bookmark_client.cc       |  2 +
  .../strings/android_chrome_strings.grd        |  3 +
  components/bookmark_bar_strings.grdp          |  6 ++
+ .../bookmarks/browser/bookmark_codec.cc       | 24 ++++++--
+ components/bookmarks/browser/bookmark_codec.h |  7 ++-
  .../browser/bookmark_load_details.cc          |  4 ++
  .../bookmarks/browser/bookmark_load_details.h |  2 +
  .../bookmarks/browser/bookmark_model.cc       |  3 +-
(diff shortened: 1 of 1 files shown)


csagan5:
Really nice work, thanks!

uazo:
I'm sorry, but I noticed that the tabs collection does not appear in the html export, now I fix it

PR #1236, opened 2021-06-25
fix logcat crash garbled ui by clearing float style

uazo:
so it should be ok. I apologize again.


csagan5:
No need to apologize.

I had seen it already before release.

csagan5 (approved):

PR #1291, opened 2021-07-21
[wip] release 92.0.4515.105

uazo:
@csagan5 here you can find the temporary patches for v92, you also find my patches ready for v92
compile but I found some bugs,

- force tablet mode crashes bromite at startup
- floc ui is still present
- proxy ui not working
- import bookmarks ui does not display files

I keep checking if there are other bugs as well
I hope this time I did better than the other times, I don't get offended if you don't use them.


uazo:
fixed the bugs I found.

as for floc, you have to excuse me, I got carried away and completely removed everything.
for the "import bookmarks ui" there were no bugs, I just tried with A10 without activating the saf flag.

I keep checking if I find any other bugs and also fix the conflict (I started before you changed the `Update-i18n-zh_CN-support.patch`).
tomorrow I also try to compile webview

csagan5:
I had already done this in past days; if you want to avoid the waste of time you can coordinate with me first.

uazo:
> I had already done this in past days; if you want to avoid the waste of time you can coordinate with me first.

no problem, I learn by comparing with your version.
the previous time you put in some patch drafts, I was re-proposing your approach. 

> I had already done this in past days;

but how do you organize yourself, or rather, how soon do you start? which version did you start from?

csagan5:
> > I had already done this in past days; if you want to avoid the waste of time you can coordinate with me first.
> 
> no problem, I learn by comparing with your version.
> the previous time you put in some patch drafts, I was re-proposing your approach.

Your fix for the proxy configuration and the force tablet mode was useful, I took it from here. This time I did not submit any draft as I did not know you or someone else were interested.

> > I had already done this in past days;
> 
> but how do you organize yourself, or rather, how soon do you start? which version did you start from?

I start a few days before the Chromium release will happen, no specific rule. See the latest release.

uazo:
> Your fix for the proxy configuration and the force tablet mode was useful, I took it from here

it's better that way!
I found a little error [here](https://github.com/bromite/bromite/blob/master/build/patches/browser-ui-disable-warning-about-missing-API-keys.patch#L17): I also check the others (I have simplified some of them) and in case I let them know.
regarding `DCScan-Fix-builds-with-system-sysroots.patch`, i had it too but i thought it was a bug of my goma server!

csagan5:
> I found a little error [here](https://github.com/bromite/bromite/blob/master/build/patches/browser-ui-disable-warning-about-missing-API-keys.patch#L17):

If this error did not come up as a compilation error then it means this patch does nothing.

> (I have simplified some of them) and in case I let them know.

Feel free to submit other improvements, we can review them in the PRs.

PR #1302, opened 2021-07-25
reactivation of subresource filter

### LINKED ISSUE #1299: Adblocker is not working
### Bromite version

Version: `92.0.4515.103`
Arch: `arm`
Android version: : `5.0.1`
Device model: `TCL-8079`

### Flags

`#disable-webgl` / WebGL : Disabled
`#smooth-scrolling` / Smooth Scrolling : Enabled
`#enable-quic` / Experimental QUIC protocol : Enabled
`#enable-gpu-rasterization` / GPU rasterization : Enabled
`#omnibox-clipboard-suggestions-content-hidden` / Clipboard suggestion content hidden : Disabled
`#read-later` / Reading List : Enabled
`#bookmark-bottom-sheet` / Enables bookmark bottom sheet : Enabled

No

Yes

No

In an incognito and normal tab

No

### Is the bug a crash?

No

### Describe the bug

Adblocker is not working. I have tried restart adblocker and browser but it didn't work.

1. Go to ad-ful site.
2. See bug

### Expected behavior

Adblocker must work like previous versions

uazo:
fix #1299 

I tried, the problem is actually there, in this way the adblock is reactivated.
obviously it has nothing to do with the rules, if those are missing they continue to fail!


csagan5 (approved):

csagan5:
Interesting, just noticed this issue; thanks for the PR!

PR #1426, opened 2021-10-02
fix current build

uazo:
little things: I fixed the patch ungoogled-chromium-Fix-building-without-enabling-reporting that doesn't apply and removed a little bug in the timezone patch (dcheck failed)


csagan5:
> I fixed the patch ungoogled-chromium-Fix-building-without-enabling-reporting that doesn't apply

I had already fixed that but did not push the changes because there has been no new release.

PR #1427, opened 2021-10-02
Add recents, offlinepage and send to home screen for always incognito

uazo:
here is the patch, unfortunately a bit big. what's up:

### Add to homescreen in always incognito
it was enough to act in `AppMenuPropertiesDelegateImpl.java` to show the button and enable `isAddToHomeIntentSupported`.
Note that I only enabled it for always incognito because the open from the homescreen icon is always in primary mode (in normal navigation, so the homepage sites saved incognito are reopened not in incognito, and that's not good).

### Enable recent tabs menu in always incognito
it was sufficient to act in `AppMenuPropertiesDelegateImpl.java` to show the button and enable the registration of the tab in `TabPersistentStore.java` at the moment of closing. in `RecentTabsManager` I simply force the use of the primary profile, simply because otherwise the list of closed tabs would not be persisted. The modification in `HistoricalTabSaver` is simply to bring me the information of the incognito mode between java and native (the always incognito flag is only on the java side).
Note that this feature is also active only for always incognito because reopening from the page is always in primary mode.

### Save page in (always) incognito
A little more complex, first some technical notes (I hope not to write nonsense and not to offend anyone): first, it is not a simple creation of an mhtml file as I thought.
the offline mode is managed by `OfflinePageModelFactory` which derives from `SimpleKeyedServiceFactory`, which is chromium's way of managing dependency on services. The services are connected to `BrowserContext`, in our case by its derivation based on the profiles (`ProfileKey`) from which the current profile (normal or offtherecord, i.e. incognito) and where the preferences service can be retrieved. Side note: the profiles themselves are an extension of the BrowserContext and that from the incognito profiles you can retrieve the normal profile via `GetOriginalProfile()`, useful for having the real persisted preferences.

So, the work was to re-enable OfflinePageModelFactory and everything needed for his management, and I found `RequestCoordinatorFactory` which I re-enabled for incognito management.

The final step is to find where the saved pages are reopened, that is in the bridge between java and native (`OfflinePageBridge`): here I forced the page to open as if it were a normal mhtml file opened by the user. I preferred this way because alternatively I would have had to rehabilitate other services and the relationship between them, which do not seem to be necessary.

As I said here, too, I preferred activation only with always incognito: I do not have the ability to predict whether now or future functionalities can allow the passage of information between normal navigation and incognito (without always incognito active).
so for this I do not link any issues, where the request were different

The changes instead you see in `HistoryPage` and `HistoryManager` are for v94 where the new IncognitoPlaceholder does not allow viewing the history in incognito. In v94 it is active by default, in v93 seem to be under a flag

### Tests
a note on tests: I'm starting to insert the tests in the patches that I do and that you find in this one too, but the question is a bit complex and must be explained separately.

### DIFF
--- a/build/patches/Add-history-support-in-incognito-mode.patch
+++ b/build/patches/Add-history-support-in-incognito-mode.patch
@@ -2,7 +2,13 @@ From: uazo <uazo@users.noreply.github.com>
 Date: Fri, 7 Aug 2020 16:33:47 +0000
 Subject: Add history support in incognito mode
 
-Add recents, offlinepage and send to home screen for always incognito
+Add history, recents, offlinepages and send to home screen for always incognito.
+History, recent tabs and offline pages require the INCOGNITO_TAB_HISTORY_ENABLED 
+flag turned on.
+IncognitoPlaceholder is also deactivated, both in the phone and tablet version.
+The relative tests are also present.
+
+See also: https://github.com/bromite/bromite/pull/1427
 ---
  .../chrome_junit_test_java_sources.gni        |   4 +
  chrome/android/chrome_test_java_sources.gni   |   6 +
--- a/build/patches/Add-history-support-in-incognito-mode.patch
+++ b/build/patches/Add-history-support-in-incognito-mode.patch
@@ -2,17 +2,76 @@ From: uazo <uazo@users.noreply.github.com>
 Date: Fri, 7 Aug 2020 16:33:47 +0000
 Subject: Add history support in incognito mode
 
+Add recents, offlinepage and send to home screen for always incognito
 ---
- .../java/res/xml/privacy_preferences.xml       |  5 +++++
- .../privacy/settings/PrivacySettings.java      | 18 ++++++++++++++++++
- chrome/browser/history/history_tab_helper.cc   | 16 ++++++++++++++++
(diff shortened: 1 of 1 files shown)


uazo:
> Will not this breach the principle of not storing outside of RAM incognito navigation details?

we already do, but only in always incognito with history enabled

> Does it mean that the recent tabs feature requires the history in incognito to be enabled?

tecnically yes, recents tabs needs some internal history services, in incognito are not enabled, even with history flag enabled.
but i've activated that only in always incognito + history enabled

>  Or, put the other way round: if history in incognito is disabled, will the recent tabs show in the history once someone enables that history in incognito setting?

no, recents tabs do not read the history, they only record the closing of the tab, if nothing is active, nothing is recorded.

my patch activates that only with always incognito + history flag, I have not touched the code related to normal incognito mode, so even with history enabled recents tabs are not shown or recorded

uazo:
>  if the user has always-incognito AND always-incognito history, then they get the recent tabs feature working?

yes, that's right, with offlinepage and send to home screen

csagan5:
Ok, I see; can you expand the patch description to mention more of this information (also from PR description)?

So that it is not lost.

Issue #1339, opened 2021-08-15
Site setting for images


frogtile:
### Is your feature request related to privacy?
In some cases, maybe?

### Is there a patch available for this feature somewhere?
I haven't look for.

### Describe the solution you would like

I want to be able to disable loading of images, fonts etc.

Primary use case is browsing with limited bandwidth. (when using cellular internet package)

### Describe alternatives you have considered

chromium used have image blocking. But they removed it (in 2015)

https://bugs.chromium.org/p/chromium/issues/detail?id=505844

frogtile:
@kokukyocho any update on this

csagan5:
The use-case for this is very small but I guess it could be reinstated with some site setting for it; @uazo what do you think?

uazo:
what is missing seems is only the ui, the content settings is still active for non android platforms

https://source.chromium.org/chromium/chromium/src/+/main:components/content_settings/core/browser/content_settings_utils.cc;l=136;drc=cb65284bde8d00949f8b51e6b6c58f2502d8af2d

csagan5:
Ok; then I would say we can re-add it with a patch similar to those we have for timezone/UA. @uazo shall I assign it to you or myself?

uazo:
> shall I assign it to you or myself?

I do it, it shouldn't take long.

csagan5:
Fixed in `94.0.4606.94`.

frogtile:
Remote fonts?

PR #1447, opened 2021-10-05
Add site setting for images

uazo:
Re-added the content settings for images. The parameter is shown by default in the PageInfo regardless of the value, so that it is easy to define exceptions.

Fix for #1339


build/patches/Add-images-contentsettings.patch
```diff
@@ -0,0 +1,454 @@
...
+     // audio.
+diff --git a/components/page_info/page_info_ui.cc b/components/page_info/page_info_ui.cc
+--- a/components/page_info/page_info_ui.cc
++++ b/components/page_info/page_info_ui.cc
+@@ -138,7 +138,7 @@ base::span<const PermissionsUIInfo> GetContentSettingsUIInfo() {
+   DCHECK(base::FeatureList::GetInstance() != nullptr);
+   static const PermissionsUIInfo kPermissionsUIInfo[] = {
+     {ContentSettingsType::COOKIES, 0},
+-    {ContentSettingsType::IMAGES, IDS_PAGE_INFO_TYPE_IMAGES},
++    {ContentSettingsType::IMAGES, IDS_PAGE_INFO_TYPE_IMAGES}, // needed
```
uazo, line 448: added because in v94 it was removed.

uazo:
I do it tomorrow, but it won't apply cleanly to v94.
don't you prefer the v94 version directly or do you do it? because if you want a clean merge, I have to wait for you to make v94 available

Issue #706, opened 2020-08-23
Investigate showing update notification


csagan5:
Investigate whether it's feasible to use the update icon with some simple update mechanism (like the adblock filters do).

csagan5:
Some people do it with Tasker: https://www.reddit.com/r/tasker/comments/iduriz/bromite_updater/

uazo:
what about reactivating omaha?

here https://source.chromium.org/chromium/chromium/src/+/master:chrome/android/java/src/org/chromium/chrome/browser/omaha/RequestGenerator.java;l=35 the request and here https://source.chromium.org/chromium/chromium/src/+/master:chrome/android/java/src/org/chromium/chrome/browser/omaha/ResponseParser.java parsing the response

in android it seems the necessary ui is already present

uazo:
I guess you already know, but here's the patch ready to use

https://github.com/ungoogled-software/ungoogled-chromium-android/blob/master/patches/ungoogled-chromium-android/Enable-update-notification.patch

or maybe I misunderstood what you are asking?

csagan5:
> what about reactivating Omaha?

Yes, that was my plan. And point it to the GitHub releases.

> I guess you already know, but here's the patch ready to use

No, I did not know. This could be usable, but I wanted to use the same mechanism used for the AdBlock filters: a check on `Last-Modified` header time of the download URL.

This is how it would work:
* on start check if there is a `PREF_LAST_TIME_UPDATE_NOTIFICATION_KEY` timestamp stored in preferences (like in the patch of ungoogled-chromium); if not present then set it to the [build time](https://github.com/chromium/chromium/blob/99314be8152e688bafbbf9a615536bdbb289ea87/base/build_time.h)
* instead of the `UPDATE_VERSION_URL` it would contact directly https://fdroid.bromite.org/latest/bromite_arm64.apk or https://fdroid.bromite.org/latest/bromite_arm.apk (depending on architecture)
* as it happens for the AdBlock filters, the first request is just a HEAD request
* parse `Last-Modified` from the `HEAD` request and compare to `PREF_LAST_TIME_UPDATE_NOTIFICATION_KEY`; if newer then proceed with `GET` request

The logic above pretty much all exists for the current AdBlock filters patch.

[...]

PR #1454, opened 2021-10-07
Enable update notification

uazo:
in the end I took the patch of @wchen342 and modified it, it came easier for me.
the logic is this:

- two new properties `BUILD_TIME` and `BUILD_TARGET_CPU` are inserted in the java class `org.chromium.build.BuildConfig`, the first taken from the same source as `GetBuildTime()`, the second taken directly from the args
- if enabled by the user (default is disabled), every 7 days (configurable in the code), a `HEAD` request is made to the page built in `BromiteInlineUpdateController.java`, actually `UPDATE_VERSION_URL + "bromite_" + BuildConfig.BUILD_TARGET_CPU + ".apk"`
- if the `Last-Modified` is greater than `BUILD_TIME`, the notification appears: that leads to `getDownloadUrl()` in an incognito page

The request is made by the native to have support for the proxy (the java side is always direct) and is without cookies.

Currently nothing is downloaded automatically, it is always the user who will have to download and install. To make it easier to find the correct file I have added `arch` to download page
`
"https://github.com/bromite/bromite/releases/latest?arch=" + BuildConfig.BUILD_TARGET_CPU;
`
but we can find an alternative.

Fix for #706


csagan5:
As for the update URL, I found that it is possible to link directly to GitHub:
```
https://github.com/bromite/bromite/releases/latest/download/arm64_ChromePublic.apk
https://github.com/bromite/bromite/releases/latest/download/arm_ChromePublic.apk
https://github.com/bromite/bromite/releases/latest/download/x86_ChromePublic.apk
https://github.com/bromite/bromite/releases/latest/download/arm64_ChromePublic.apk
```

Can you please adjust the URL to use this pattern instead?

uazo:
one more thing, by chance you have the files available
```
build/util/LASTCHANGE
build/util/LASTCHANGE.committime

source: build/compute_build_timestamp.py
```
of the current build? can you keep the next ones for verification?
I'm curious to see if they are the same regardless of effective compile time.

can you also verify that the `Last-Modified` matches that value? because I guess you need to do a `touch` of the apk file

csagan5:
> one more thing, by chance you have the files available
> 
> ```
> build/util/LASTCHANGE
> build/util/LASTCHANGE.committime
> 
> source: build/compute_build_timestamp.py
> ```
> 
> of the current build? can you keep the next ones for verification? I'm curious to see if they are the same regardless of effective compile time.

They look like this:
```
10 Oct  7 23:54 build/util/LASTCHANGE
10 Oct  7 23:46 build/util/LASTCHANGE.committime
```
And the latter contains `1633568863`; `build/compute_build_timestamp.py` explains how these work.

> can you also verify that the `Last-Modified` matches that value? because I guess you need to do a `touch` of the apk file

The APK will always have a timestamp larger than the build timestamp because it is generated after the build, so I do not think we need to worry about this.

csagan5:
Can you please resolve the conflict? I am going to try this on v94.

uazo:
> We do not need anything related to build time and `Last-Modified`: the version check alone will suffice. 

OK, i will drop unused code

> Also: did you solve the problem about how to get the version from the first `location:` header?

yes, it works

csagan5:
There is a still conflict and the branch is protected so I cannot solve it myself...

csagan5:
It's not a big problem, I can always cherry-pick the commits - but then on GitHub it will not show as "merged" but rather "closed".

> do you know which option I need to activate to allow you the changes?

Probably adding me as collaborator would work?

> or maybe it is better that I create the branches directly in the Bromite repo?

Up to you, it is not a big deal.

uazo:
> Probably adding me as collaborator would work?

Done!

csagan5:
> > Probably adding me as collaborator would work?
> 
> Done!

I cannot accept that as it shares too much information for the invite.

I just noticed that it performs two HEAD checks, there must be somewhere a bug that makes it perform the operation twice; will try to figure this out.

csagan5:
> I did that to follow the redirect to the end anyway.

What would it need it for?

uazo:
> What would it need it for?

I left it if in the future (you or me) wanted to retrieve more from the actual `HEAD` of the file, I didn't remove it because I noticed that you don't like the commented codes.
The real `HEAD` is taken in `EndpointFetcher::OnURLLoadComplete`: consider that code must remain as it runs there in case there is no internet connection

csagan5:
It seems to be always empty; you can see how I adjusted the code as soon as I publish the patches for v94.

uazo:
> you can see how I adjusted the code as soon as I publish the patches for v94.

yes, how you did is fine

> Is that response_->redirect_url always empty or always non-empty?
> It seems to be always empty

in the hypothesis of having more redirects before reaching the url of the file, the first time is empty, the following ones are not.


```
1.  if (response_->redirect_url.empty()) {
2.    response_->redirect_url = url_.spec();
3.    response_->response = std::to_string(redirect_info.status_code);
4.  } else {
5.    LOG(INFO) << "BromiteUpdater: redirect URL is not empty, status code is " << redirect_info.status_code;
6.  }
7.  std::move(endpoint_fetcher_callback_).Run(std::move(response_));
```

so, having changed (7), the else to (5) will never be reached

Issue #1431, opened 2021-10-02
Enable encrypted Backups


famo:
### Is your feature request related to privacy?

Yes, with [Seedvault](https://github.com/seedvault-app/seedvault) it is possible to do privacy respecting backups - IF the app allows it...

### Is there a patch available for this feature somewhere?

Yes, see:
https://github.com/grote/Transportr/commit/4dc38f429f75909a088d8bd8a5b3b5ddd8030f71

from
[seedvault FAQ#why-do-some-apps-not-allow-to-get-backed-up](https://github.com/seedvault-app/seedvault/wiki/FAQ#why-do-some-apps-not-allow-to-get-backed-up)

### Describe the solution you would like

Bromite should allow backups which are encrypted.

### Describe alternatives you have considered

None.

Issue #244, opened 2019-02-06
Adding (self-hosted) Firefox Sync to Bromite


makapao:
**Is your feature request related to privacy?**
Yes.

**Is there a patch available for this feature somewhere?**
Not exactly an android-related patch, but there is some documentation from another project that might help.

**Describe the solution you'd like**

The ability to sync bookmarks, history, passwords and open tabs with Firefox.

At the moment there is no way for the user to share bookmarks, history or passwords with another browser, apart from manually importing bookmarks. Firefox Sync is a decent end-to-end encrypted option to share information between multiple browsers and an Android device on-the-go. 

Firefox Sync support would allow the end-to-end encrypted syncing of bookmarks, history, passwords, and open tabs with both desktop and mobile Firefox. For users of Firefox desktop it would make migrating to Bromite a more attractive prospect. For existing users it would add functionality.

**Describe alternatives you've considered**
I currently use either Firefox Notes (or Google Keep if I can't help it) to copy URLs and history I want to use between desktop Firefox on Linux and Bromite on Android. A lot more cumbersome, but it works.

**Other implementations**

**Epiphany (gnome browser)** have [added this feature](https://camo.githubusercontent.com/767d7f5b3f482217c4ce96203bfed8c0389e90d4/68747470733a2f2f692e696d6775722e636f6d2f57584a6374304a2e6a7067) and the following might be useful for understanding how they did it:

Description: [https://github.com/GNOME/epiphany/blob/70755404799e4393bf5605c9debec6877acf0177/lib/sync/README](https://blogs.gnome.org/mcatanzaro/2017/08/09/on-firefox-sync/)

Possibly related commit: [https://github.com/GNOME/epiphany/tree/70755404799e4393bf5605c9debec6877acf0177/lib/sync](https://github.com/GNOME/epiphany/tree/70755404799e4393bf5605c9debec6877acf0177/lib/sync)

Explainer on Epiphany Firefox Sync Support: [https://github.com/GNOME/epiphany/blob/70755404799e4393bf5605c9debec6877acf0177/lib/sync/README](https://github.com/GNOME/epiphany/blob/70755404799e4393bf5605c9debec6877acf0177/lib/sync/README)

**Further Reading**
Mozilla documentation

Mozilla host their sync client documentation here: [https://github.com/GNOME/epiphany/blob/70755404799e4393bf5605c9debec6877acf0177/lib/sync/README](https://moz-services-docs.readthedocs.io/en/latest/sync/)
[...]

PR #1500, opened 2021-10-21
Export & restore all settings

uazo:
this patch is a work in progress, I need to talk to you about what to do.

1) not a problem in itself, but the `ChromeBackupAgent` code seems unused to me, that is, I couldn't find anything that could enable it in the chromium base code.
in fact it is based on the presence of "`backup_key`" which however does not exist in any gn ([ref](https://source.chromium.org/chromium/chromium/src/+/main:chrome/android/java/AndroidManifest.xml;l=181))
2) the current code uses androidbackup in [key-value pairs mode](https://developer.android.com/guide/topics/data/keyvaluebackup), the advantage is that I manage it as I want but it seems to have the limit of 5MB per backup. however this limit does not seem to me to be imposed by android (I have not found anything about it in its base code).
3) currently I backup profile preferences (native), local state (native) and SharedPreferences (java), ie _everything_ that is managed by the settings. everything that is managed through sqllite is _not covered_.
4) sqllite dbs are not managed because they are open (therefore inaccessible) and above all because in general it is not enough to replace the file when restoring, because I would risk losing the current information present, since there is no management of when it is done the backup but above all the restore. theoretically I should study the structure of each db and make the inserts / updates necessary for correct management and keep them at each modification of the structure, unthinkable.
5) the most absurd thing in android is that the application has no management of when androidbackup does the backup or restore!
and this is the main problem because to restore we have to relaunch the application, since some settings require it (always incognito or javascript jit and all about://flags).
6) and finally, the backup is not currently encrypted and goes straight to google's servers (for those who have it active), although this seems to be editable with [backup_rules.xml](https://developer.android.com/guide/topics/data/autobackup#define-device-conditions) as also written in #1431.
seedvault seems to be a good alternative, but from what I understand it needs to be integrated into the android build so it is not possible to install it stand alone (I'm waiting for someone to prove me wrong)

however, point 5 is the most critical.
a possible alternative is to run `bmgr` [ref](https://developer.android.com/studio/command-line/bmgr) as it does for logcat, so you can manage when the user wants to backup and restore, you tell me.
other alternative is to throw everything and start over, emulating [BackupDataOutput](https://cs.android.com/android/platform/superproject/+/master:frameworks/base/core/java/android/app/backup/BackupDataOutput.java) to save the file when and where I want.
instead point 4 I would see it together with #244 

let me know

### DIFF
--- a/build/patches/Enable-android-backup-wip.patch
+++ b/build/patches/Enable-android-backup-wip.patch
@@ -1,0 +1,569 @@
+From: uazo <uazo@users.noreply.github.com>
+Date: Thu, 21 Oct 2021 07:06:08 +0000
+Subject: [wip] enable android backup
+
+---
+ chrome/android/chrome_public_apk_tmpl.gni     |   1 +
+ .../java/res/xml/about_chrome_preferences.xml |  11 ++
+ .../chrome/browser/ChromeBackupAgentImpl.java | 122 ++++++++++++++++--
+ .../about_settings/AboutChromeSettings.java   |  70 +++++++++-
+ chrome/browser/android/chrome_backup_agent.cc |  97 ++++++++++++++
+ .../strings/android_chrome_strings.grd        |  12 ++
+ components/prefs/json_pref_store.cc           |   5 +
+ components/prefs/json_pref_store.h            |   2 +
+ components/prefs/persistent_pref_store.h      |   3 +
+ components/prefs/pref_service.cc              |   4 +
+ components/prefs/pref_service.h               |   2 +
+ 11 files changed, 316 insertions(+), 13 deletions(-)
+
+diff --git a/chrome/android/chrome_public_apk_tmpl.gni b/chrome/android/chrome_public_apk_tmpl.gni
+--- a/chrome/android/chrome_public_apk_tmpl.gni
++++ b/chrome/android/chrome_public_apk_tmpl.gni
+@@ -22,6 +22,7 @@ default_chrome_public_jinja_variables = [
+   "channel=$android_channel",
+   "enable_vr=$enable_vr",
+   "include_arcore_manifest_flag=false",
++  "backup_key=unused" # see https://developer.android.com/guide/topics/data/keyvaluebackup#BackupManifest
+ ]
+ 
+ # Enable stack unwinding only on official build with specific channels. It is
+diff --git a/chrome/android/java/res/xml/about_chrome_preferences.xml b/chrome/android/java/res/xml/about_chrome_preferences.xml
+--- a/chrome/android/java/res/xml/about_chrome_preferences.xml
++++ b/chrome/android/java/res/xml/about_chrome_preferences.xml
+@@ -19,4 +19,15 @@
+         android:fragment="org.chromium.chrome.browser.about_settings.LegalInformationSettings"
+         android:key="legal_information"
+         android:title="@string/legal_information_title" />
++    <org.chromium.components.browser_ui.settings.ChromeSwitchPreference
++        android:key="allow_android_backup"
++        android:title="@string/allow_android_backup_title"
++        android:summary="@string/allow_android_backup_summary"
++        android:defaultValue="false" />
++    <Preference
++        android:key="backup_now"
++        android:title="@string/backup_now_title"/>
++    <Preference
++        android:key="restore_now"
++        android:title="@string/restore_now_title"/>
+ </PreferenceScreen>
+diff --git a/chrome/android/java/src/org/chromium/chrome/browser/ChromeBackupAgentImpl.java b/chrome/android/java/src/org/chromium/chrome/browser/ChromeBackupAgentImpl.java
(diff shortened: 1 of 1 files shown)


csagan5:
> 1. not a problem in itself, but the `ChromeBackupAgent` code seems unused to me, that is, I couldn't find anything that could enable it in the chromium base code.
>    in fact it is based on the presence of "`backup_key`" which however does not exist in any gn ([ref](https://source.chromium.org/chromium/chromium/src/+/main:chrome/android/java/AndroidManifest.xml;l=181))

Is `backup_key` a remnant or something that somehow is activated only during the official Chrome builds?

Also: have you read this? https://bugs.chromium.org/p/chromium/issues/detail?id=661188 The backup logic was disabled at some point

> 2\. the current code uses androidbackup in [key-value pairs mode](https://developer.android.com/guide/topics/data/keyvaluebackup), the advantage is that I manage it as I want but it seems to have the limit of 5MB per backup. however this limit does not seem to me to be imposed by android (I have not found anything about it in its base code).

See the previous issue, perhaps it was a limitation on older Android versions.



> * currently I backup profile preferences (native), local state (native) and SharedPreferences (java), ie _everything_ that is managed by the settings. everything that is managed through sqllite is _not covered_.
> * sqllite dbs are not managed because they are open (therefore inaccessible) and above all because in general it is not enough to replace the file when restoring, because I would risk losing the current information present, since there is no management of when it is done the backup but above all the restore. theoretically I should study the structure of each db and make the inserts / updates necessary for correct management and keep them at each modification of the structure, unthinkable.

I don't think we should cover the SQLite files; what is there anyways of user interest? cookies, bookmarks, saved passwords?



> 5\. the most absurd thing in android is that the application has no management of when androidbackup does the backup or restore!
>     and this is the main problem because to restore we have to relaunch the application, since some settings require it (always incognito or javascript jit and all about://flags).

There is no way to detect a **restore** happened? But perhaps it will automatically close it before restoring?



> 6\. and finally, the backup is not currently encrypted and goes straight to google's servers (for those who have it active), although this seems to be editable with [backup_rules.xml](https://developer.android.com/guide/topics/data/autobackup#define-device-conditions) as also written in [Enable encrypted Backups #1431](https://github.com/bromite/bromite/issues/1431).
>     seedvault seems to be a good alternative, but from what I understand it needs to be integrated into the android build so it is not possible to install it stand alone (I'm waiting for someone to prove me wrong)

So #1431 would be useless for all those that do not have a rooted phone/custom ROM?

As for the backup going to Google servers: is this the functionality used when switching to a new phone? It works differently than the sync profile.

> a possible alternative is to run `bmgr` [ref](https://developer.android.com/studio/command-line/bmgr) as it does for logcat, so you can manage when the user wants to backup and restore, you tell me.

Would this require a rooted device as well?

> other alternative is to throw everything and start over, emulating [BackupDataOutput](https://cs.android.com/android/platform/superproject/+/master:frameworks/base/core/java/android/app/backup/BackupDataOutput.java) to save the file when and where I want.
> instead point 4 I would see it together with #244

It depends from what we want to maintain; I would first study a bit further how backup used to work for Chromium on Android, to see if we can restore some functionality (and whether it is useful or not). In the end if we do not want to use the core Android functionality we could have some import/export of all settings and flags alone, in a file, but sooner or later we would start thinking about encrypting it etc which becomes scope creep.

csagan5:
> really anomalous, but there was no safer method?

Asking the user to enter a master password...to encrypt all the other passwords, would be safer. And it would still need key derivation.

> basically the passwords should never be saved using the chromium internal manager on a rooted android, correct?

There is no right answer to this. If your phone is rooted it does not mean that all apps can do anything they want. Most SU managers still require you to give permissions before root is granted to apps; some other security features are disabled when you root, but not all and might not be applicable for this specific attack scenario.

> > I do not consider this encryption with a hard-coded password valid from the perspective of securing user data
> 
> what if it was manageable by the user? provided that that value is then registered in a store that is really valid for security purposes. I wonder where the credit card details are.
> > chromium however has already included the library javax.crypto
> 
> what do you think of this?

This export feature can work without any form of encryption; it is up to the user to figure that out, meanwhile we can further think about the use-case.

> > I am talking about the Android sandbox: can other apps open files created by other apps, without any limitation?
> 
> as far as I understand, in a non-rooted system, if the file is on the internal memory, only if the application allows it through a saf provider. and also, before A11, if an application was installed on the external card (which is accessible by all applications), but I don't know if there is a possibility to install that way now.
> 
> more than the other applications, the problem is the browser itself, as he obviously has access to those files. an example is the page chrome://local-state/ if for some reason was readable by the sites. when i've done the crashui i saw that access is governed by [IsAccessAllowedAndroid](https://source.chromium.org/chromium/chromium/src/+/main:chrome/browser/net/chrome_network_delegate.cc;l=127), if there was a way to cheat that function you would have access to all chromium files.

There are many more barriers to break before sites can access a local file, it is not just governed by `IsAccessAllowedAndroid`.

PR #1533, opened 2021-11-02
Fix patches for v95

uazo:
there is only to regenerate the "Automated domain substitution"
if you have any doubts, just ask!


uazo:
if you want to see them before applying them use [this](https://uazo.github.io/superpatch/bromite/commit/1e63bb866ccf113c929fea71823fedcb6a23f3b2), so you tell me if it can be useful

csagan5:
Thanks for the PR; are these fixes for the build or is there also some functionality change?

uazo:
> are these fixes for the build or is there also some functionality change

i just re-enabled the build. a check must be made. 

> As for the missing PNGs: that might be due to the automatic patch apply that I used.

yes, they were just missing. what did you use?

uazo:
i found a crash.
if you open an offline page without having opened any tab it crashes.
this is because the incognito profile is created only after having opened at least one tab.
https://github.com/bromite/bromite/pull/1533/commits/d17ea911b1ff68f825e2eeaccf94cfb9bb989dd4 fix it

csagan5:
> > As for the missing PNGs: that might be due to the automatic patch apply that I used.
> 
> yes, they were just missing. what did you use?

I think it is the `patch` command that ignores binary blobs.

PR #1555, opened 2021-11-09
fix overlap in bookmarks patch editing

### LINKED ISSUE #103: Bromite patches
Hi @csagan5, I haven't been able to apply `Add-bookmark-import-export-actions.patch` since cr93, build fails with

```
../../chrome/android/java/src/org/chromium/chrome/browser/bookmarks/BookmarkBridge.java:776: error: method initialize in class DownloadLocationCustomView cannot be applied to given types;
                mCustomView.initialize(
                           ^
  required: int,long
  found: int,File,int,String
  reason: actual and formal argument lists differ in length
../../chrome/android/java/src/org/chromium/chrome/browser/bookmarks/BookmarkBridge.java:781: error: cannot find symbol
                Resources resources = mContext.getResources();
                                      ^
  symbol: variable mContext
../../chrome/android/java/src/org/chromium/chrome/browser/bookmarks/BookmarkBridge.java:1271: error: cannot find symbol
            ((Activity)window.getContext().get()).runOnUiThread(new Runnable() {
              ^
  symbol:   class Activity
  location: class BookmarkBridge
```

The last one is easy to solve but the first error I am not sure where you get the extra parameters from?

uazo:
fix https://github.com/ungoogled-software/ungoogled-chromium-android/issues/103

### DIFF
--- a/build/patches/Add-bookmark-import-export-actions.patch
+++ b/build/patches/Add-bookmark-import-export-actions.patch
@@ -14,7 +14,7 @@ Completely remove contacts picker permission from the file dialog
  .../browser/TabbedModeTabDelegateFactory.java |   5 +-
  .../browser/bookmarks/BookmarkActionBar.java  |  12 +
  .../browser/bookmarks/BookmarkActivity.java   |  30 ++
- .../browser/bookmarks/BookmarkBridge.java     | 277 +++++++++++++++++
+ .../browser/bookmarks/BookmarkBridge.java     | 278 +++++++++++++++++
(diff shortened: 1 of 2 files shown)


csagan5:
The end result is the same but the changes are better arranged by feature, right?

uazo:
> The end result is the same but the changes are better arranged by feature

yes, the error started with [92.0.4515.103](https://github.com/bromite/bromite/blob/92.0.4515.103/build/patches/Add-menu-item-to-bookmark-all-tabs.patch#L164)

PR #1558, opened 2021-11-09
permanently disabled the fetch of field trials

uazo:
i fixed the actual patch because it is not related to android code.
that is, the `disable_fieldtrial_testing_config` flag is active in other configurations but not in android.
fortunately, the download is not active because it was disabled by the FRE patch, I don't know if it is intentionally.
i also eliminated the fall-back to (luckily few) field trials in case the download of the experiments fails
I also removed the support for the `X-Client-Data` header which sends the configuration of the experiments in progress to the google servers only.
if you want to check, with `chrome://version/?show-variations-cmd` the active trials are displayed

### DIFF
--- a/build/patches/Disable-fetching-of-all-field-trials.patch
+++ b/build/patches/Disable-fetching-of-all-field-trials.patch
@@ -3,13 +3,129 @@ Date: Sun, 8 Jul 2018 18:16:34 +0200
 Subject: Disable fetching of all field trials
 
 ---
- .../variations/service/variations_service.cc       | 14 ++------------
- 1 file changed, 2 insertions(+), 12 deletions(-)
+ .../chrome_metrics_services_manager_client.cc |  2 ++
+ components/ukm/ukm_recorder_impl.cc           |  3 +++
+ .../firstrun/VariationsSeedBridge.java        |  2 +-
+ .../firstrun/VariationsSeedFetcher.java       | 23 ++++---------------
+ .../variations/net/omnibox_http_headers.cc    |  2 +-
+ .../variations/net/variations_http_headers.cc |  3 +++
+ .../variations/service/variations_service.cc  | 15 +++---------
+ 7 files changed, 17 insertions(+), 33 deletions(-)
 
+diff --git a/chrome/browser/metrics/chrome_metrics_services_manager_client.cc b/chrome/browser/metrics/chrome_metrics_services_manager_client.cc
+--- a/chrome/browser/metrics/chrome_metrics_services_manager_client.cc
++++ b/chrome/browser/metrics/chrome_metrics_services_manager_client.cc
+@@ -174,6 +174,8 @@ ChromeMetricsServicesManagerClient::~ChromeMetricsServicesManagerClient() {}
+ void ChromeMetricsServicesManagerClient::CreateFallbackSamplingTrial(
+     version_info::Channel channel,
+     base::FeatureList* feature_list) {
++  if ((true))
++    return;
+   // The trial name must be kept in sync with the server config controlling
+   // sampling. If they don't match, then clients will be shuffled into different
+   // groups when the server config takes over from the fallback trial.
+diff --git a/components/ukm/ukm_recorder_impl.cc b/components/ukm/ukm_recorder_impl.cc
+--- a/components/ukm/ukm_recorder_impl.cc
++++ b/components/ukm/ukm_recorder_impl.cc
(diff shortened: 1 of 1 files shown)


uazo:
ah, one more thing, [here](https://source.chromium.org/chromium/chromium/src/+/main:components/feature_engagement) you will find some more details.
I am becoming more and more convinced that google uses the backup of applications to acquire this information as well.

build/patches/Disable-fetching-of-all-field-trials.patch
```diff
@@ -24,11 +140,12 @@ diff --git a/components/variations/service/variations_service.cc b/components/va
 -  }
 -#endif
 -  return true;
++  // set to false event if not used in android
```
csagan5, line 143:
```suggestion
+  // set to false even if not used on Android
```

csagan5:
> i fixed the actual patch because it is not related to android code.
> that is, the `disable_fieldtrial_testing_config` flag is active in other configurations but not in android.

So https://github.com/bromite/bromite/blob/master/build/GN_ARGS#L9 is not effective on Android? It means Bromite has the built-in fieldtrial testing config flags enabled by default?

> fortunately, the download is not active because it was disabled by the FRE patch, I don't know if it is intentionally.

Can you refer to the actual change that makes the download not active in the first run experience patch?

> I also removed the support for the `X-Client-Data` header which sends the configuration of the experiments in progress to the google servers only.

`X-Client-Data` sending is already disabled because no domain has any special treatment, see `ungoogled-chromium-Disable-Google-host-detection.patch`

> i also eliminated the fall-back to (luckily few) field trials in case the download of the experiments fails

The net effect of this PR is the above and that the built-in field trials (currently active) will be disabled, correct?

uazo:
> It means Bromite has the built-in fieldtrial testing config flags enabled by default?

Code is here:
```
  bool used_testing_config = false;
#if BUILDFLAG(FIELDTRIAL_TESTING_ENABLED)                                           <--- THIS IS FALSE
  if (!command_line->HasSwitch(switches::kDisableFieldTrialTestingConfig) &&
      !command_line->HasSwitch(::switches::kForceFieldTrials) &&
      !command_line->HasSwitch(switches::kVariationsServerURL)) {
    // Note that passing base::Unretained(this) below is safe because the
    // callback is executed synchronously.
    AssociateDefaultFieldTrialConfig(
        base::BindRepeating(&VariationsFieldTrialCreator::OverrideUIString,
                            base::Unretained(this)),
        GetPlatform(), client_->GetCurrentFormFactor(), feature_list.get());
    used_testing_config = true;
  }
#endif  // BUILDFLAG(FIELDTRIAL_TESTING_ENABLED)
  bool used_seed = false;
  if (!used_testing_config) {                                                       <--- THIS IS TRUE
    used_seed = CreateTrialsFromSeed(low_entropy_provider.get(),
                                     feature_list.get(), safe_seed_manager);
  }
```
I did not find any other references to the flag.
what changes the "entropy" from high to low. if you want to check, maybe I'm wrong.

> Can you refer to the actual change that makes the download not active in the first run experience patch?

Yes.

> X-Client-Data sending is already disabled because no domain has any special treatment,

the code foresees the use also if the initiator is the user, unfortunately I have not marked the line, I try to find it.

> The net effect of this PR is the above and that the built-in field trials (currently active) will be disabled, correct?

completely deactivated, obviously always unless my mistakes

csagan5:
> I did not find any other references to the flag.
> what changes the "entropy" from high to low. if you want to check, maybe I'm wrong.

But we also have a patch to disable the trials seed, that is effective, as far as I know.



> > The net effect of this PR is the above and that the built-in field trials (currently active) will be disabled, correct?
> 
> completely deactivated, obviously always unless my mistakes

Ok, what I am trying to understand is whether there is something undesired already happening or we are just fortifying the patches against future undesired behaviours.

uazo:
> But we also have a patch to disable the trials seed

which? I only saw this one that I edited.

> what I am trying to understand is whether there is something undesired already happening

no, all right

> or we are just fortifying the patches against future undesired behaviours.

Yes, it is.
I checked the code to see if there were any other places where trials were being modified and found these things fixed with this patch.

csagan5:
> > But we also have a patch to disable the trials seed
> 
> which? I only saw this one that I edited.

`Disable-seed-based-field-trials.patch`

csagan5:
> > But we also have a patch to disable the trials seed
> 
> which? I only saw this one that I edited.

The other patch is `Disable-seed-based-field-trials.patch`.

PR #1579, opened 2021-11-14
Re-introduce patch to add option to use home page as NTP

csagan5:
This does not work for a circular dependency and the methods in `TabCreator` would have to be implemented as abstract. There are at last 2 implementations to cover after that.


csagan5:
Superseded by #1586

PR #1582, opened 2021-11-15
Permanently Disable Floc

uazo:
Permanently removes floc support, disabling the download of lsh clusters,  the marking the history navigation and the javascript api and permission policies.

I also added a way to the disabling of blink features through the DisabledForBromite tag, so that it doesn't happen that we forget the flags, try to look, it is a hypothesis of course.

theoretically it is not really necessary everything I have done, in fact I am a bit in doubt, because, apart from the download of the clusters (which they use to estimate the amount of users in the calculation of the hashes, using those that are synchronized with their servers , despite privacy), the rest is already disabled.

also marking the navigation as floc enabled is theoretically already disabled, because the `kAdTagging` flag is disabled (by the way, why?)

however I have seen that, unlike the other features related to privacy sandbox, this one is quite stable so maybe it won't give us problems in the future. in case, I can simplify it.

### DIFF
--- a/build/patches/Disable-floc.patch
+++ b/build/patches/Disable-floc.patch
@@ -1,0 +1,479 @@
+From: uazo <uazo@users.noreply.github.com>
+Date: Sat, 13 Nov 2021 18:02:15 +0000
+Subject: Disable Floc
+
+Permanently removes floc support, disabling the download of lsh clusters, 
+the marking the history navigation and the javascript api and permission policies.
+Also added the disabling of blink features through the DisabledForBromite tag
+---
+ .../browser/chrome_content_browser_client.cc  |  4 ++
+ .../floc_component_installer.cc               | 10 +---
+ .../floc_eligibility_observer.cc              | 58 +------------------
+ .../floc_id_provider_impl.cc                  |  4 +-
+ .../privacy_sandbox_settings.cc               | 14 +++--
+ .../federated_learning/features/features.cc   | 12 ++--
+ components/federated_learning/floc_id.cc      | 19 +++---
+ .../history/core/browser/history_backend.cc   | 13 +----
+ .../history/core/browser/history_service.cc   |  9 +--
+ components/history/core/browser/url_row.h     |  1 -
+ third_party/blink/common/features.cc          |  4 +-
+ .../permissions_policy_feature.mojom          |  5 --
+ .../blink/renderer/core/dom/document.cc       |  3 +-
+ .../blink/renderer/core/dom/document.idl      |  1 -
+ .../renderer/core/dom/interest_cohort.idl     |  4 --
+ .../permissions_policy_features.json5         |  6 --
+ .../platform/runtime_enabled_features.json5   | 12 +++-
+ 17 files changed, 45 insertions(+), 134 deletions(-)
+
+diff --git a/chrome/browser/chrome_content_browser_client.cc b/chrome/browser/chrome_content_browser_client.cc
+--- a/chrome/browser/chrome_content_browser_client.cc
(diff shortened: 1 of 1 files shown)


csagan5:
> the `kAdTagging` flag is disabled (by the way, why?)

Because it is used only for this:
```
const char kPresetPerformanceTestingDryRunOnAllSites[] =
    "performance_testing_dryrun_on_all_sites";
```

Obviously there is no interest in Bromite to run performance testing for any third-party.

csagan5:
> theoretically it is not really necessary everything I have done, in fact I am a bit in doubt,
> however I have seen that, unlike the other features related to privacy sandbox, this one is quite stable so maybe it won't give us problems in the future. in case, I can simplify it.

I am concerned that this patch (like some others recently merged) will make porting to v96 and so on even more difficult/time-consuming. Also: one important aspect of security is small patches, the more changes the more chances to introduce bugs and possibly security bugs.

uazo:
> Because it is used only for this:

no it is not true.
there is a whole internal logic that, in some way, tries to understand if a page contains an ad and if it blocks it automatically.
at first glance it also seems well done, because it navigates the javascript stack in search of something (which I have not yet understood well).
if you agree it should be evaluated better.

> I am concerned that this patch (like some others recently merged) will make porting to v96 and so on even more difficult/time-consuming.

that's right, it's my own concern. the fact is: do we trust upstream? that is, if there is a flag that says to disable, should we trust it? it's all there, I think.

> the more changes the more chances to introduce bugs and possibly security bugs.

uhm .. I wouldn't generalize, it depends on the change.

csagan5:
> > the more changes the more chances to introduce bugs and possibly security bugs.
> 
> uhm .. I wouldn't generalize, it depends on the change.

This can be generalized very well: bugs are introduced when code is changed, and when code is poorly understood bugs are introduced in bigger numbers.

> > Because it is used only for this:
> 
> no it is not true. there is a whole internal logic that, in some way, tries to understand if a page contains an ad and if it blocks it automatically. at first glance it also seems well done, because it navigates the javascript stack in search of something (which I have not yet understood well). if you agree it should be evaluated better.

Where is this logic? This is the description of the feature:
```
// Enables the tagging of ad frames and resource requests by using the
// subresource_filter component in dry-run mode.
extern const base::Feature kAdTagging;
```

Feel free to explore how it works, but obviously I expect that _not all ads_ are blocked automatically.

uazo:
> Where is this logic? 

https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/frame/ad_tracker.cc

csagan5:
@uazo any idea why the feature description is so poor?

uazo:
I don't know, maybe because it's more of a blink feature than a chromium feature.

csagan5:
I will also re-enable `kAdTagging`.

uazo:
> I will also re-enable kAdTagging.

did you see what it does? it's safe?

csagan5:
> > I will also re-enable kAdTagging.
> 
> did you see what it does? it's safe?

It looks safe, I think it covers a hole with CNAME aliases, see https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/platform/loader/fetch/resource_loader.cc;drc=429093c183d9f48446bc5d2dae196fa6d907c9e1;l=1622

Not sure if it will be effective in Bromite (in case it needs additional resources), but we can try.

uazo:
> It looks safe, I think it covers a hole with CNAME aliases, see

that code however is under a flag `kSendCnameAliasesToSubresourceFilterFromRenderer` disabled by default

PR #1583, opened 2021-11-15
Disable conversion measurement api

### LINKED ISSUE #1531: Disable "Conversion Measurement API"

Yes

No

I think that patch is incomplete: there are other types of reports not covered by that patch like [QueueAccessReport](https://source.chromium.org/chromium/chromium/src/+/main:content/browser/net/cross_origin_opener_policy_reporter.cc;l=222;drc=7af7400d03766f76cdc34b9ffee27178dc9430c6;bpv=1;bpt=1?q=cross-origin-opener-policy-report-only
) or QueueNavigationReport

Fix directly [this](https://source.chromium.org/chromium/chromium/src/+/main:services/network/network_context.cc;l=1007;drc=8cc4402c875055fe7989c15f7dfdd04a46c753fc;bpv=1;bpt=1) or the other side of the mojo call

uazo:
Disable `Conversion Measurement API` by disabling the flag and removing support for the AttributionReporting provider. it also removes the handling of attributions via intents between apps.
This patch enforces the deactivation by preventing the report from being sent and being saved to disk, although it is currently in uncalled code.

For this patch I preferred the "soft" approach for two reasons. The first is that the deletion is intrusive ([this](https://gist.github.com/uazo/2ac3c8d6640a300c4e764bfe082f9a32) is the patch that completely removes it), the second is that I have seen that it is still under development and in any case the flag actually seems to disable most of the code.
what was left I took off.

consider that it is currently active in bromite.

fix #1531

### DIFF
--- a/build/patches/Disable-conversion-measurement-api.patch
+++ b/build/patches/Disable-conversion-measurement-api.patch
@@ -1,0 +1,333 @@
+From: uazo <uazo@users.noreply.github.com>
+Date: Mon, 15 Nov 2021 09:43:29 +0000
+Subject: Disable conversion measurement api
+
+Disable Conversion Measurement API by disabling the flag and removing 
+support for the AttributionReporting provider. it also removes 
+the handling of attributions via intents between apps.
+This patch enforces the deactivation by preventing the report from 
+being sent and being saved to disk, although it is currently in uncalled code.
+---
+ chrome/android/java/AndroidManifest.xml         |  8 --------
+ .../AttributionIntentHandlerFactory.java        |  9 +--------
+ .../flags/android/chrome_feature_list.cc        |  2 +-
+ .../embedder_support/origin_trials/features.cc  |  3 ++-
+ .../render_view_context_menu_base.cc            |  3 ---
+ .../browser/android/navigation_handle_proxy.cc  | 10 ----------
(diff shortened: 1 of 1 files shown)


build/patches/Disable-conversion-measurement-api.patch
```diff
@@ -0,0 +1,333 @@
...
+       name: "ConversionMeasurement",
+-      origin_trial_feature_name: "ConversionMeasurement",
+-      origin_trial_allows_third_party: true,
+-      status: "experimental",
++      depends_on: ["DisabledForBromite"],
+     },
+     {
+       name: "ConversionMeasurementEventSources",
+-      status: "test",
++      depends_on: ["DisabledForBromite"],
```
csagan5, line 327: If you start using this, shouldn't the other patches also use this approach?

csagan5 (approved):

build/patches/Disable-conversion-measurement-api.patch
uazo: in fact, I put the statement in the other patch (https://github.com/bromite/bromite/pull/1582)...
if you don't merge that it won't compile.
maybe it is the case to put it alone.

beware that once again I have not added it to the patch list

PR #1586, opened 2021-11-16
Update Add-option-to-use-home-page-as-NTP.patch for v95

uazo:
patch fixed, I added a button to set, if the user wants, the blank page.
I didn't use `UrlConstants.CHROME_BLANK_URL` because it's deprecated and removed a month ago.

### DIFF
--- a/build/patches/Add-option-to-use-home-page-as-NTP.patch
+++ b/build/patches/Add-option-to-use-home-page-as-NTP.patch
@@ -1,26 +1,26 @@
 From: csagan5 <32685696+csagan5@users.noreply.github.com>
-Date: Tue, 16 Nov 2021 11:48:52 +0000
+Date: Sat, 20 Nov 2021 15:36:54 +0000
 Subject: Add option to use home page as NTP
 
 And allow use about:blank as default homepage
 ---
- .../java/res/xml/homepage_preferences.xml     |  9 ++++++
- .../browser/homepage/HomepageManager.java     | 16 +++++++++++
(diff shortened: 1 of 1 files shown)


csagan5:
> patch fixed, I added a button to set, if the user wants, the blank page.

I think this is not needed because if you set your home page as blank and then use the home as NTP, you achieved the same result.

uazo:
> I think this is not needed because if you set your home page as blank and then use the home as NTP, you achieved the same result.

yes, that's right, I had noticed. then I saw this [assert](https://source.chromium.org/chromium/chromium/src/+/main:chrome/android/java/src/org/chromium/chrome/browser/metrics/LaunchMetrics.java;l=106;drc=429093c183d9f48446bc5d2dae196fa6d907c9e1) and I got the doubt, in fact I removed it, and it still seems to work, although it would be better defined.

csagan5:
That is about metrics, not important.

csagan5 (changes requested):
Since it is possible to use `about:blank` via the home page it would be best that we make the patch smaller/simpler and remove that (duplicate) capability.

uazo:
>  would be best that we make the patch smaller/simpler

it's already quite small and simple, what do you mean?

> and remove that (duplicate) capability.

sorry I don't understand the request, can you detail it?

csagan5:
> > and remove that (duplicate) capability.
> 
> sorry I don't understand the request, can you detail it?

Please remove `NTP_SET_ABOUT_BLANK`

csagan5 (approved):

PR #1590, opened 2021-11-17
Try to fix #1570 and add a flag for incognito mode

uazo:
fix #1570 as shown in https://github.com/bromite/bromite/issues/1570#issuecomment-971505527

added flag to enable native autofill for incognito mode (default false)

PR #1592, opened 2021-11-18
Reactivation of browser autofill

### LINKED ISSUE #1534: Auto-fill pop-up menu doesn't appear when selecting or typing in an entry field
### Bromite version

Version: 94.0.4606.109
Arch: arm64
Android version: 12 beta 5
Device model: Pixel 3a

### Flags

No

Yes

No

Yes

No

### Is the bug a crash?

No

### Describe the bug

auto-fill pop-up menu doesn't appear when selecting or typing in an entry field

Steps to reproduce the bug:
1. go to https://fill.dev/form/identity-simple
2. select an entry field or start typing
3. the auto-fill pop-up menu does not appear
4. if you long-press the field, a pop-up menu will show 'paste' along with a three-dot button. clicking the button brings up 'auto-fill'. clicking it has no effect

### Expected behavior

the corresponding auto-fill information appears in pop-up

### Screenshots

Screenshot_20211104-230545
- long-press entry field result

Screenshot_20211104-230551
- click on three-dot menu button result

Screenshot_20211104-230700
- correct behavior in chrome

uazo:
I modified the patch to reactivate the browser autofill management together with the native one.
Some functions are in fact managed by the internal chromium manager.

Fix for #1534

### DIFF
--- a/build/patches/Enable-native-Android-autofill.patch
+++ b/build/patches/Enable-native-Android-autofill.patch
@@ -6,25 +6,46 @@ There are 2 different types of autofill: one managed via GCM and the
 native Android one that uses the provider assigned by the user
 (which can be any user installed app like Bitwarden for example).
 In chromium GCM is active while in the WebView the latter.
-This patch uses the WebView code to activate native Android autofill.
+This patch uses WebView code to enable native Android autofill 
+along with browser-managed autofill.
 
 See also: https://github.com/bromite/bromite/issues/547
 ---
+ android_webview/browser/aw_contents.cc        |  3 +-
  chrome/android/BUILD.gn                       |  1 +
- .../settings/PasswordSettings.java            | 58 ++++++++++++++++++-
- .../chromium/chrome/browser/tab/TabImpl.java  | 43 ++++++++++++++
- .../browser/tab/TabViewAndroidDelegate.java   | 14 +++++
- chrome/browser/BUILD.gn                       |  8 +++
- chrome/browser/android/tab_android.cc         | 26 +++++++++
+ .../settings/PasswordSettings.java            | 81 ++++++++++++++++++-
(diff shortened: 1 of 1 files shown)


csagan5:
I think the PR does not describe everything which is happening here: there are 2 new switches introduced?

uazo:
you are right, because there are also changes related to #1590

csagan5 (changes requested):
Please update the PR to describe what changes are being made; the patch description should also mention that.

csagan5:
The conflict cannot be fixed by me:
```
remote: Resolving deltas: 100% (4/4), completed with 4 local objects.
remote: error: GH006: Protected branch update failed for refs/heads/enable-browser-autofill.
remote: error: At least 1 approving review is required by reviewers with write access.
```

However I have merged your patch and applied the changes I requested (to the patch description, PR description still does not mention the new toggle).

uazo:
> remote: error: At least 1 approving review is required by reviewers with write access.

I'm sorry, but I just don't know how to fix it.
it would be easier perhaps if you allowed me to create branches directly in bromite.

csagan5:
It is not a big deal, only some merges are missing.

> I'm sorry, but I just don't know how to fix it.

This error explains it: `At least 1 approving review is required by reviewers with write access.` You need to remove branch protection from branches which are not `master`.

If you fork from scratch it will also have the correct settings (or it used to have them).

> it would be easier perhaps if you allowed me to create branches directly in bromite.

That is not possible, I prefer to keep the security very strict.

uazo:
> This error explains it: At least 1 approving review is required by reviewers with write access. You need to remove branch protection from branches which are not master

I am about to insert a new action on my fork to automatically check the git apply, maybe I will ask for help to configure the permissions correctly and then remove that protection.
the alternative is that you accept superpowers in my fork, I think I can trust :)

> That is not possible, I prefer to keep the security very strict.

my own motivation

PR #1606, opened 2021-11-20
Always ask for permissions

### LINKED ISSUE #1549: Unable to use camera or microphone
### Bromite version

Version: `95.0.4638.78`
Arch: `arm`
Android version: `12`
Device model: `Pixel 5a`

### Flags

None changed

No

Yes

No

Yes

No

### Is the bug a crash?

No

```
<paste here logcat dump individuated with 'adb logcat | grep chromium'>
```

### Describe the bug

When I navigate to a website(such as https://webrtc.github.io/samples/src/content/devices/input-output/ or https://doxy.me/precall-test ) that uses the phones camera or microphone, neither work. I have disabled and re-enabled all Bromite permissions in androids permission manager. I have tried to clear the cache and data for the app. I have reinstalled the app. None of these steps have solved this issue. Curiously enough, in Android 12, they added a new notification that is a green dot in the upper right of the notification bar that shows up when a app is accessing the camera or microphone. When I go to those website and allow camera and microphone, the green dot appears, but neither the front or back camera show video and no audio is recorded and played back.

1. Go to https://webrtc.github.io/samples/src/content/devices/input-output/ or https://doxy.me/precall-test
2. Allow access to camera and microphone

### Expected behavior

Expect to see video from devices camera and audio from the microphone.

uazo:
Always ask for permissions for location, microphone and camera in every session and reactivation of the kUnifiedAutoplay flag.
The patch initiates the request to the user at the first use of the android location, microphone and camera permissions, even with the relative content setting active.

fix #1549


build/patches/Always-ask-for-permissions.patch
```diff
@@ -0,0 +1,291 @@
...
+diff --git a/chrome/android/java/src/org/chromium/chrome/browser/infobar/PermissionUpdateInfoBarDelegate.java b/chrome/android/java/src/org/chromium/chrome/browser/infobar/PermissionUpdateInfoBarDelegate.java
+--- a/chrome/android/java/src/org/chromium/chrome/browser/infobar/PermissionUpdateInfoBarDelegate.java
++++ b/chrome/android/java/src/org/chromium/chrome/browser/infobar/PermissionUpdateInfoBarDelegate.java
+@@ -23,6 +23,11 @@ import java.util.Collections;
+ import java.util.HashSet;
+ import java.util.Set;
+ 
++import org.chromium.components.permissions.AndroidPermissionRequester;
++
++// remove this comment when moved to
```
csagan5, line 54: This discussion should be in the PR description not in the comments.

When would the move happen? v96?

build/patches/Always-ask-for-permissions.patch
```diff
@@ -0,0 +1,291 @@
...
+         if (mNativePtr != 0) {
+diff --git a/chrome/browser/permissions/permission_update_infobar_delegate_android.cc b/chrome/browser/permissions/permission_update_infobar_delegate_android.cc
+--- a/chrome/browser/permissions/permission_update_infobar_delegate_android.cc
++++ b/chrome/browser/permissions/permission_update_infobar_delegate_android.cc
+@@ -53,6 +53,9 @@ infobars::InfoBar* PermissionUpdateInfoBarDelegate::Create(
+     permissions::AppendOptionalAndroidPermissionsForContentSetting(
+         content_settings_type, &optional_permissions);
+ 
++    // in bromite there are other contentsettings defined as permission
++    // so in debug please don't crash
```
csagan5, line 80: Can we cover them in the relative patches instead?

build/patches/Always-ask-for-permissions.patch
```diff
@@ -0,0 +1,291 @@
...
+                     }
+                 }
+ 
+@@ -198,6 +207,13 @@ public class AndroidPermissionRequester {
+         return true;
+     }
+ 
++    public static void showMissingPermissionDialogUnchecked(Activity activity, @StringRes int messageId,
++            Runnable onPositiveButtonClicked, Runnable onCancelled) {
++        // permitted without any check for requestFileAccessPermissionHelper.FileAccessPermissionHelper
```
csagan5, line 191: Where is the filter for that specific `FileAccessPermissionHelper` applied?

build/patches/Disable-unified-autoplay-feature.patch
```diff
@@ -1,23 +0,0 @@
...
-
-diff --git a/media/base/media_switches.cc b/media/base/media_switches.cc
---- a/media/base/media_switches.cc
-+++ b/media/base/media_switches.cc
-@@ -480,7 +480,7 @@ const base::Feature kUseR16Texture{"use-r16-texture",
- // Enables the Unified Autoplay policy by overriding the platform's default
- // autoplay policy.
- const base::Feature kUnifiedAutoplay{"UnifiedAutoplay",
--                                     base::FEATURE_ENABLED_BY_DEFAULT};
-+                                     base::FEATURE_DISABLED_BY_DEFAULT};
```
csagan5, line 17: Can you remind me what the unified autoplay does exactly?

build/patches/Always-ask-for-permissions.patch
uazo: > This discussion should be in the PR description not in the comments.

ok

> When would the move happen? v96?

I don't know, I still haven't figured out (well) how to see it (I know, you already explained it to me :)

build/patches/Always-ask-for-permissions.patch
uazo: > Can we cover them in the relative patches instead?

yes, I marked it.
however, I am waiting for a bug to be fixed in the content settings patches.

build/patches/Always-ask-for-permissions.patch
uazo: I do not understand.

that function is for the download manager to request access to the storage, I excluded it from the checks.

build/patches/Disable-unified-autoplay-feature.patch
uazo: set the policy `kDocumentUserActivationRequired`
```
// The document needs to have received a user activation or received one
// before navigating.
```
that is, any click on the document

instead of `kUserGestureRequired`
```
// A local user gesture on the element is required.
```
that is, a click on the element.

csagan5 (changes requested):
Please remove discussions from the comments

build/patches/Always-ask-for-permissions.patch
csagan5: Fixed by upstream (issue number = ?) or by another PR (PR number = ?)?

build/patches/Always-ask-for-permissions.patch
```diff
@@ -0,0 +1,291 @@
...
+-        NOTREACHED();
++        //NOTREACHED();
+       }
+     } else if (message_id == IDS_INFOBAR_MISSING_CAMERA_PERMISSION_TEXT) {
+-      DCHECK(content_settings_type == ContentSettingsType::MEDIASTREAM_MIC);
++      //DCHECK(content_settings_type == ContentSettingsType::MEDIASTREAM_MIC);
+       message_id = IDS_INFOBAR_MISSING_MICROPHONE_CAMERA_PERMISSIONS_TEXT;
+     } else if (message_id == IDS_INFOBAR_MISSING_MICROPHONE_PERMISSION_TEXT) {
+-      DCHECK(content_settings_type == ContentSettingsType::MEDIASTREAM_CAMERA);
++      //DCHECK(content_settings_type == ContentSettingsType::MEDIASTREAM_CAMERA);
```
csagan5, line 98: These changes should also be removed from here

build/patches/Always-ask-for-permissions.patch
csagan5: Ok; so I think the comment should be something like:
```
// do not re-request permissions for file access
```

build/patches/Disable-unified-autoplay-feature.patch
csagan5: Ok; and you want to enable it because the local user gesture will be replaced by the permissions request instead.

build/patches/Always-ask-for-permissions.patch
uazo: I meant "the first future bromite bug affecting one of those patches"

build/patches/Always-ask-for-permissions.patch
uazo: do you intend not to comment or not to remove?
I put them because without those, I would not be able to use the debug version with active dchecks, it would often crash. there are already other parts that make me crash bromite in debug: every new version I have to remove them, I would not like to have others.

build/patches/Always-ask-for-permissions.patch
csagan5: I do not understand what has to be moved and who decides when

build/patches/Always-ask-for-permissions.patch
csagan5: So there is currently no bug but you would like these to be removed just in case? If the bugs happen they need to be fixed, we should not remove these.

I understand you wrote that the DCHECKs are already failing, then we need to fix them - not ignore all.

build/patches/Always-ask-for-permissions.patch
csagan5: > there are already other parts that make me crash bromite in debug: every new version I have to remove them, I would not like to have others.

The solution is to make the DCHECKs not fail, not to ignore them. For example the `NOTREACHED()`: we cannot just ignore that.

build/patches/Always-ask-for-permissions.patch
uazo: > not ignore all.

uhm... and [this](https://github.com/bromite/bromite/pull/731#issuecomment-687851540)?

csagan5: That also needs to be fixed.

This PR cannot be merged with these commented DCHECKs.

build/patches/Always-ask-for-permissions.patch
uazo: fixed in https://github.com/bromite/bromite/pull/1606/commits/e5a236efbc5c2fa74667089fa3d4c508563fc4aa

`PermissionUpdateInfoBarDelegate.java` was removed in v96

build/patches/Always-ask-for-permissions.patch
uazo: fixed in https://github.com/bromite/bromite/pull/1606/commits/e5a236efbc5c2fa74667089fa3d4c508563fc4aa

that code was removed in v96

build/patches/Always-ask-for-permissions.patch
uazo: fixed in https://github.com/bromite/bromite/pull/1606/commits/e5a236efbc5c2fa74667089fa3d4c508563fc4aa

uazo:
you can find everything in

https://github.com/uazo/bromite/blob/master-v97/build/patches/Always-ask-for-permissions.patch

PR #1620, opened 2021-11-27
Patches for v96

uazo:
@csagan5 these are the patches for v96
they build and seem to work.
there is no `Automated-domain-substitution.patch`, I haven't learned how to generate it yet ..


csagan5:
@uazo I have already prepared the patches for v96, they are not released yet

csagan5:
> there is no `Automated-domain-substitution.patch`, I haven't learned how to generate it yet ..

This uses ungoogled-chromium's scripts, I could share this part as well, it is not very sophisticated.

csagan5:
I will commit my patches for v96 later; I skipped the user scripts one and the timezone one.

The timezone patch has really little use, I do not think we should have added it in first place; but regardless, what do you think of merging it with the user-agent one?

uazo:
> The timezone patch has really little use,

I think so too

>  I skipped the user scripts one and the timezone one.

those here should go

> but regardless, what do you think of merging it with the user-agent one?

I increasingly like the idea that patches remain separate by functionality, why would you want to merge them?

csagan5:
> I increasingly like the idea that patches remain separate by functionality, why would you want to merge them?

To reduce work when porting patches.

> > The timezone patch has really little use,
> 
> I think so too

We could drop it.

uazo:
> To reduce work when porting patches.

i'm trying to solve that problem, but i'm going to trial because i still haven't found a good idea to follow.

> We could drop it.

I'd wait a bit, I'd like to understand first how the various javascript fingerprinting tools use that data.
it is also true that we can reintroduce it later, but honestly I see more problems with other patches.

csagan5:
> > We could drop it.
> 
> I'd wait a bit, I'd like to understand first how the various javascript fingerprinting tools use that data. it is also true that we can reintroduce it later, but honestly I see more problems with other patches.

That is clear to me, and was also before the introduction of the patch: timezone corresponds to very few bits, 2-3 max, and over the rest of fingerprinting data it is meaningless.

The only purpose of this patch is to provide some sense of uniqueness to whoever does not understand how uniqueness/fingerprinting works (snake oil).

Which other patches are you referring to?

uazo:
> timezone corresponds to very few bits, 2-3 max, and over the rest of fingerprinting data it is meaningless.

interesting, source? I would like to understand a little more too ...
for now I just understand that I have to read the specs and the PING Reviews, because it is always reporting the privacy effects.
the problem seems to me that there are few new implementations that do not affect the possibility of tracking, and that following all the new features is complex and should be organized.
Plus, I don't want to sound opinionated, after all, who am I to say delete a feature if for example mozilla also says it's useful?
I'll give you an example https://w3c.github.io/hr-time/#sec-privacy-security on which some changes are being proposed. for me all `Performance` should be removed (or submitted to the user's choice), but instead it is supported by everyone.


> Which other patches are you referring to?

they made me waste more time `Disable safe browsing` and `Remove signin and data saver integrations` in the v96 and
`Remove binary blob integrations` in the v95.
however, relative "more time", because it took me less than a day to port to the new versions (among other things, doing something else). it's more of the hassle of not being able to do it automatically ...

on these three I thought I'd try to create some no-op classes to replace the current ones, so as not to change all that code, I would like to try it sooner or later

csagan5:
> > timezone corresponds to very few bits, 2-3 max, and over the rest of fingerprinting data it is meaningless.
> 
> interesting, source? I would like to understand a little more too ...

Last time I dived into this topic I collected some interesting sources here: https://github.com/bromite/bromite/wiki/Fingerprinting
But nowadays I am sure there are better ones.

To explain my point above about the little bits that timezone covers: let's make an example that tries to model all the fingerprinting bits that client-side can extract from browsers:



420 is all the bits we know about (User Agent, Canvas, Audio etc everything which is known, including timezone) while `x` is all the bits that can be extracted but we have not yet tackled down (it is a cat and mouse game after all).
The real number is not 420, I have not counted how many bits are known but for timezone you can estimate it because there are a certain amount of maximum timezone values possible (let's say 378 according to [Wikipedia](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones)). This value can be expressed with 9 bits:

 378">

So we have that the timezone influence on the total amount of fingerprinting data is:



My point is that this value is very small and negligible, and since we do not know the actual value of `x` it tends to be even smaller.

> Plus, I don't want to sound opinionated, after all, who am I to say delete a feature if for example mozilla also says it's useful?
> I'll give you an example https://w3c.github.io/hr-time/#sec-privacy-security on which some changes are being proposed. for me all `Performance` should be removed (or submitted to the user's choice), but instead it is supported by everyone.

This is an opinionated project; it has a practical use abut it is also open source and we do it for learning and fun. If opinions vary too much from the current mission/ideas it can always be forked and made it something different.

I have been thinking about making something like this: https://mozilla.github.io/standards-positions/
(we can follow Mozilla in most cases and use a different one in cases like performance as you pointed out).

> > Which other patches are you referring to?
> 
> they made me waste more time `Disable safe browsing` and `Remove signin and data saver integrations` in the v96 and `Remove binary blob integrations` in the v95. however, relative "more time", because it took me less than a day to port to the new versions (among other things, doing something else). it's more of the hassle of not being able to do it automatically ...
> 
Sure; but it used to be even worse before those last 2 patches were merged from a variety of different ones; now it is much leaner compared to the past.

uazo:
> Last time I dived into this topic I collected some interesting sources here: https://github.com/bromite/bromite/wiki/Fingerprinting

interesting, I hadn't seen it. I will study it

> To explain my point above about the little bits that timezone covers

thanks, i will study this too.

> I have been thinking about making something like this

I like it and it's a good idea, we should understand how long it takes to fill it with content

> now it is much leaner compared to the past.

I guess

csagan5:
@uazo before closing this PR can you please check if there is anything I missed? I have already checked the diff of the patches.

uazo:
`Add-flag-to-disable-vibration.patch`: both you and I have put the flag between two comments, but it doesn't matter

`Disable-safe-browsing.patch`: we use different approaches, but I guess they do the same thing.
for `manage_sync_preferences.xml` I was in doubt because it is used by the safe browsing code rather than `Remove-signin-and-data-saver-integrations.patch` but that's okay anyway

`Restore-Search-Ready-Omnibox-flag.patch`: there is something that may not be right in `about_flags.cc` try to see with https://uazo.github.io/superpatch/bromite/commit/master

`Revert-the-removal-of-an-option-to-block-autoplay.patch` and `Site-setting-for-images.patch`: I have seen some `index` that maybe there shouldn't be (`index d6cba78972382..ea78c679219ac 100644` and `index ea78c679219ac..36bfd0553d944 100644`)

`ungoogled-chromium-Disable-domain-reliability.patch` and `ungoogled-chromium-Disable-Google-host-detection.patch`: I think you have reversed the position in the patches of `HasGoogleSearchQueryParam` but it is the same

csagan5:
> `Add-flag-to-disable-vibration.patch`: both you and I have put the flag between two comments, but it doesn't matter

Fixed.



> `Disable-safe-browsing.patch`: we use different approaches, but I guess they do the same thing.
> for `manage_sync_preferences.xml` I was in doubt because it is used by the safe browsing code rather than `Remove-signin-and-data-saver-integrations.patch` but that's okay anyway

Fixed.



> `Restore-Search-Ready-Omnibox-flag.patch`: there is something that may not be right in `about_flags.cc` try to see with https://uazo.github.io/superpatch/bromite/commit/master

I have put it there on purpose, it should conflict on every release but at least will not move around with automatic fuzzying.

> `Revert-the-removal-of-an-option-to-block-autoplay.patch` and `Site-setting-for-images.patch`: I have seen some `index` that maybe there shouldn't be (`index d6cba78972382..ea78c679219ac 100644` and `index ea78c679219ac..36bfd0553d944 100644`)

The index lines are optional in all cases except the binary patches; will filter them out in future.

> `ungoogled-chromium-Disable-domain-reliability.patch` and `ungoogled-chromium-Disable-Google-host-detection.patch`: I think you have reversed the position in the patches of `HasGoogleSearchQueryParam` but it is the same

Fixed.

> ah the `Never-send-any-crash-upload-data.patch` I assure you it is useless, that code is not used in android

It is not included anymore in v96.

csagan5:
> it's more of the hassle of not being able to do it automatically ...

For this I have thought of a new patch format: the patch description would contain some encoded information on what to do when the patch fails applying. I have not yet made a PoC for this.

uazo:
> I have not yet made a PoC for this.

I'm curious to see your idea!
speaking of metadata, I am trying using kythe (do you remember?), but after some hours of generation I blocked it because the index file had reached 200gb with only contained 1/5 of the chromium classes... it's a great shame, I thought it could be something to exploit, I'll try a few terabyte hd and try again

csagan5:
I remember you mentioned that tool, looks interesting; we could also setup something like https://source.chromium.org/chromium for Bromite, but it would be huge.

> > I have not yet made a PoC for this.
> 
> I'm curious to see your idea!

I will let you know.

uazo:
>  we could also setup something like

I thought so too, for now it's science fiction.

> for Bromite

and to read the brave code more easily :)

uazo:
> for Bromite
> and to read the brave code more easily :)

the other use we could make of it is to extract all new `base::Features` automatically being scattered throughout the code

csagan5:
Yes I also would like to track all features across releases, it is an old TODO item of mine.

uazo:
> the other use we could make of it is to extract all new base::Features automatically

look, I did it!
[base_feature_flags.xlsx](https://github.com/bromite/bromite/files/7677553/base_feature_flags.xlsx)

I used clangd rather than kythe, because the resources for the latter are too high.
rather complex, I had to modify ninja to keep low the amount of data.

PATCH: 
https://github.com/uazo/bromite-buildtools/blob/master/ninja-one-target-for-compdb.patch

GENERATOR: 
https://github.com/uazo/bromite-buildtools/blob/master/.github/workflows/build_bromite.yaml#L332
https://github.com/uazo/bromite-buildtools/blob/master/.github/workflows/build_bromite.yaml#L354

you can try with the results:
```
docker run -ti --rm -p 50051:50051 uazo/bromite-remote-index:5b6a7442e0c41d682efd0670755240f08563225a
```

to extract it I used this https://github.com/frabert/ClangQL which allows me to browse the index via sql (unfortunately only windows, but maybe it can be compiled for linux too)

```
.load clangql
CREATE VIRTUAL TABLE llvm_symbols USING clangql (symbols, <ip_address>:50051);
CREATE VIRTUAL TABLE llvm_base_of USING clangql (base_of, <ip_address>:50051);
CREATE VIRTUAL TABLE llvm_overridden_by USING clangql (overridden_by, <ip_address>:50051);
CREATE VIRTUAL TABLE llvm_refs USING clangql (refs, <ip_address>:50051);

SELECT * FROM llvm_symbols WHERE Name = "Feature" and scope = "base::";
SELECT * FROM llvm_symbols WHERE type = "c:$@N@base@S@Feature";
```

what can we do with this list?

csagan5:
> look, I did it!
> [base_feature_flags.xlsx](https://github.com/bromite/bromite/files/7677553/base_feature_flags.xlsx)

Very clever use of `ninja`, congratulations!

> what can we do with this list?

If we had infinite time: review each feature. But more realistically for every new major version we should check what was introduced and what was dropped to see if they are in line with Bromite goals.

uazo:
> But more realistically for every new major version we should check what was introduced and what was dropped to see if they are in line with Bromite goals.

perfect, for the next ones I will generate it manually, then I will think about making it automatic.
by the way, if you need help with v97 I'm available.

PR #1628, opened 2021-12-01
Make image shields more useful

BraINstinct0:
Images used as shields are expected to work as hyperlinks to relevant urls.
They weren't.
Now they should be.
Alternative: Use shields.io, they provide high-res, auto-updated shield images.
Try the below:
[GitHub release (latest SemVer)](https://github.com/bromite/bromite/releases/latest) [GitHub license](https://github.com/bromite/bromite/blob/master/LICENSE)

Downside is, without adding Shields.io app to the project, you might hit GitHub rate limits. By adding it it can be mitigated freely.
Note that Shields.io itself is also an opensource project without any tracking, so that won't be much of a problem.


csagan5:
Since it can be done without using shields.io, why would one want to add an extra dependency?

Thanks for your PR, I will merge it as-is.

PR #1650, opened 2021-12-13
Fix crash when cookies preference is default

### LINKED ISSUE #1355: Crash when clicking permissions in certain websites
### Bromite version

Version: `92.0.4515.134`
Arch: `arm`
Android version: `10`
Device model: `Redmi 7a`

### Flags

1. Chrome-sharing-hub: Disabled
2. Omibox-switch-to-tab-suggestions: Enabled
3. Reading-list: Enabled
4. Tab-group-UI-improvements: Enabled

No

Yes

No
Yes

No

### Is the bug a crash?

Yes

### Describe the bug

Basically clicking the permissions settings on certain websites cause the app to hang and crash to homescreen

Steps to reproduce the bug:
1. Go to forum.palemoon.com or msn.com
2. Click on the lock icon
3. click on permissions settings

### Expected behavior

Bromite shows the pop-up to toggle settings like JavaScript, block ads, etc

### Actual behaviour

The app hangs and exits to home screen after a while

uazo:
fixes #1355

### DIFF
--- a/build/patches/Show-site-settings-for-cookies-javascript-and-ads.patch
+++ b/build/patches/Show-site-settings-for-cookies-javascript-and-ads.patch
@@ -3,6 +3,7 @@ Date: Sat, 28 Dec 2019 10:23:04 +0100
 Subject: Show site settings for cookies, javascript and ads
 
 Avoid displaying info about intrusive ads
+Fix crash when cookies preference is default
 ---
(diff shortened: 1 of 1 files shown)


uazo:
if you want, you can find the updated patch for v97 in

https://github.com/uazo/bromite/blob/master-v97/build/patches/Show-site-settings-for-cookies-javascript-and-ads.patch

as soon as possible I will make a new pull request.

PR #1651, opened 2021-12-13
Fix always incognito mode

### LINKED ISSUE #1496: Site setting permissions reset on exit in always incognito
it's a bug report @uazo told me to create a report of what i initially tought was a feature or so. 

### Bromite version

Version: 94.0.4606.94
Arch: arm64
Android version: 11
Device model: HD1903

### Flags

Webgl disabled - on
Adaptive button in top toolbar - always new tab page
Isolate additional orogins - on & empty
Strict-origin isolation - on
TLS Post-Quantum Confidentiality enabled
Tab Groups, TG UI, and TG Continuation all disabled. 
Incognito screenshots - enabled

No

Yes

No

Yes

No

### Is the bug a crash?

No

### Describe the bug

If you run Bromite with always incognito mode, and any site setting disbaled by default, going to a random website, and whitelisting the disabled setting does not presist after a restart of the browser.

Steps to reproduce the bug:
Enable always incognito under privacy settings.
Disable JS, Cookies or Images by default in Site Settings
Go to github.com
Open the quick site settings, by tapping the incognito sign next to the url, and select permissions and whitelist the site. 
Close Bromite
Go back to github.com
The blacklisted parameter should be disabled, our whitelist entry is earased.

### Expected behavior

I expect the changed setting to presist after restarting bromite. 

### Screenshots

N/A

uazo:
this is the most complex, fixes #1496 

I moved the preference from java to native, at the first start it brings the current value, but unfortunately I was not able to completely eliminate the use of SharedPreferences because there are points (when managing intents) where I need the value but the native is not yet initialized.
The native flag was needed for the content settings, which can now be changed.

I changed the incognito history patch in the same way.

### DIFF
--- a/build/patches/Add-an-always-incognito-mode.patch
+++ b/build/patches/Add-an-always-incognito-mode.patch
@@ -6,42 +6,34 @@ More specifically, add a preference that causes all new tabs and all
 clicked links to launch as incognito.
 Make sure initial incognito status is correctly recognized.
 Enable incognito custom tabs and fix crashes for incognito/custom tab intents (credits to @uazo)
+Added preference in native
+Modification of content settings allowed
 ---
- chrome/android/chrome_java_sources.gni        |  1 +
  .../java/res/xml/privacy_preferences.xml      |  5 ++
- .../AlwaysIncognitoLinkInterceptor.java       | 80 +++++++++++++++++++
- .../chrome/browser/ChromeTabbedActivity.java  |  6 +-
- .../chrome/browser/app/ChromeActivity.java    |  4 +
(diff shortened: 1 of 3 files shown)


uazo:
please, disregard this (I no longer have the button to lock the merge)
I have found a cleaner method that I like best

csagan5:
@uazo you mean that this should be closed? Is there a new PR open already?

uazo:
> you mean that this should be closed?

yes, I close it, these changes do not satisfy me.
i can't find a clean way to use flags before native activation. i think i will try with switches.

>  Is there a new PR open already?

no not yet.

PR #1705, opened 2022-01-11
Patches for v97

uazo:
@csagan5, I hope everything is fine.

here you can find the patches for v97, if you need them.

I simply point out that I had to add two more:

`Fix-StartupTabPreloader-crash.patch` to fix an assert at startup due to the removal of metrics
`Fix-build-error-for-v97.patch` because in my environment I detect an error due to the use of libatomic that is not needed (because the compiler has built-in support).


uazo:
> Why bromite 97 is not already compiled

no obligation, ever. I just hope it's not something serious, given the moment.
I close this because these are already old. If you want, you can find the latest in [master-v97](https://github.com/uazo/bromite/tree/master-v97/build/patches)

coxackie:
no one has heard from @csagan5 ? This is a bit worrying...

uazo:
> no one has heard from @csagan5 ?

I do not.

Issam28:
Yes this is getting a bit worrying... I hope he is doing fine.

csagan5:
> I close this because these are already old. If you want, you can find the latest in [master-v97](https://github.com/uazo/bromite/tree/master-v97/build/patches)

I am refactoring the build process, will release the v97 patches after that is complete.

csagan5:
I have used your version 97 of these patches:
* `Remove-binary-blob-integrations.patch`, although I see you removed some files from it?
* `Bromite-auto-updater.patch`

We can continue discussion here: https://github.com/bromite/bromite/pull/1795

uazo:
> We can continue discussion here: https://github.com/bromite/bromite/pull/1795

I cannot write or do any reviews, unfortunately github does not allow me ...
this is what maybe you could check, I hope not to waste your time.

1. you probably have to change `Add-menu-item-to-bookmark-all-tabs.patch` like that
```
+    // Used to bookmark all tabs in a specific folder, created if not existing
+    public BookmarkId addToTabsCollection(Context context, Tab tab) {
+        BookmarkId parent = getTabsCollectionFolderId();
+        BookmarkId existingId = getUserBookmarkIdForTab(tab);
+        if (existingId != null && existingId.getId() != BookmarkId.INVALID_ID) {
+            BookmarkId existingBookmarkId = new BookmarkId(existingId.getId(), BookmarkType.NORMAL);
+            BookmarkItem existingBookmark = getBookmarkById(existingBookmarkId);
```
strange that you can compile

2. in `Add-option-to-not-persist-tabs-across-sessions.patch` I think you have the wrong position of
```
+        ChromeSwitchPreference closeTabsOnExitPref =
+                (ChromeSwitchPreference) findPreference(PREF_CLOSE_TABS_ON_EXIT);
+        closeTabsOnExitPref.setOnPreferenceChangeListener(this);
+        closeTabsOnExitPref.setManagedPreferenceDelegate(mManagedPreferenceDelegate);
```
it should not be in `createManagedPreferenceDelegate()` but in `updatePreferences()`

3. try to verify `Allow-building-without-enable_reporting.patch`, does not seem correct:
```
+#if BUILDFLAG(ENABLE_REPORTING)
     base::DictionaryValue body_to_pass;
     for (const auto& pair : body) {
       body_to_pass.SetString(pair.first, pair.second);
@@ -122,6 +123,7 @@ void CrossOriginEmbedderPolicyReporter::QueueAndNotify(
       storage_partition->GetNetworkContext()->QueueReport(
           kType, *endpoint, context_url_, reporting_source_,
           network_isolation_key_,
+#endif  <----------------------------------------------------------- one line below
           /*user_agent=*/absl::nullopt, std::move(body_to_pass));
     }
```

4. in `Disable-safe-browsing.patch` why did you change the `kEnableCrowdDenyTriggering` flag to false? that is, what is it for?

5. in `Disable-text-fragments-by-default.patch` I think the position of this is wrong
```
@@ -4079,6 +4083,7 @@ void Document::UpdateBaseURL() {
     DCHECK(!elem_sheet_->Contents()->RuleCount());
     elem_sheet_ = CSSStyleSheet::CreateInline(*this, base_url_);
   }
+  }
 
   if (!EqualIgnoringFragmentIdentifier(old_base_url, base_url_)) {
     // Base URL change changes any relative visited links.
```

6. `Experimental-user-scripts-support.patch` you should get my version, I realize it is quite different from yours because I added the flag for the evals. eventually you could remove it temporarily and then we talk about it in another pull

7. in `JIT-less-toggle.patch` I think the position of this is wrong, should be before the if
```
@@ -268,6 +274,11 @@ public class PrivacySettings
         Preference privacySandboxPreference = findPreference(PREF_PRIVACY_SANDBOX);
         if (privacySandboxPreference != null) {
+        ChromeSwitchPreference forceNoJit =
+                (ChromeSwitchPreference) findPreference(PREF_FORCE_NO_JIT);
+        forceNoJit.setOnPreferenceChangeListener(this);
+        forceNoJit.setManagedPreferenceDelegate(mManagedPreferenceDelegate);
+
             privacySandboxPreference.setSummary(
                     PrivacySandboxSettingsFragment.getStatusString(getContext()));
```

8. some difference between `Skip-the-first-run-and-metrics.patch` and `Remove-signin-and-data-saver-integrations.patch`, but the meaning is the same, we could unify them so we don't think about it anymore

9. in `User-agent-customization.patch` it seems to me that this is incorrect
```
--- a/components/embedder_support/user_agent_utils.cc
+++ b/components/embedder_support/user_agent_utils.cc
@@ -275,6 +275,13 @@ blink::UserAgentBrandList GenerateBrandVersionList(
 }
 
 blink::UserAgentBrandVersion GetGreasedUserAgentBrandVersion(
+  base::CommandLine* command_line = base::CommandLine::ForCurrentProcess();
+  if (command_line->HasSwitch(kUserAgent)) {
+    std::string ua = command_line->GetSwitchValueASCII(kUserAgent);
+
+    return metadata;
+  }
+
```

csagan5:
> > We can continue discussion here: #1795
> 
> I cannot write or do any reviews, unfortunately github does not allow me ... this is what maybe you could check, I hope not to waste your time.

It is due to how GitHub permissions work: it is not possible to make someone a collaborator without giving dangerous permissions as well (deletion). I have now unlocked #1795.

I will reply there.

PR #1751, opened 2022-02-02
Import 0082-Hide-currently-no-op-preload-settings.patch from Vanadium

ASTechnical:
Patch for Hiding the “Preload Pages for faster browsing and searching” Setting.


csagan5:
This should work when enabled, it is not a no-op.

Issue #1749, opened 2022-02-02
Enable trivial auto var init


BayLee4:
### Is your feature request related to privacy?

Not really, it's mostly related to security.

### Is there a patch available for this feature somewhere?

Yes : https://github.com/GrapheneOS/Vanadium/tree/12/patches

### Describe the solution you would like

Merge the patches using the Clang compile option `-ftrivial-auto-var-init=zero` as it prevents memory corruption vulnerabilities related to uninitialized memory.

[enable-ftrivial-auto-var-init-zero.patch](https://github.com/GrapheneOS/Vanadium/blob/12/patches/0005-enable-ftrivial-auto-var-init-zero.patch)
[disable-broken-warning-for-auto-var-init.patch](https://github.com/GrapheneOS/Vanadium/blob/12/patches/0006-disable-broken-warning-for-auto-var-init.patch)

Merge the patch enabling by default the `Strict Origin Isolation` flag, it should have no impact for devices with low memory as (afaik) Site Isolation is already disabled for them, and it increase the security of devices with sufficient memory. Enabling it would also help with #1720.

[Enable-strict-origin-isolation-by-default.patch](https://github.com/GrapheneOS/Vanadium/blob/12/patches/0080-Enable-strict-origin-isolation-by-default.patch)

### Describe alternatives you have considered

• `-ftrivial-auto-var-init=zero` related patches : No simple alternative, as it's a compile time flag.
• `Strict Origin Isolation` related patch : Keep the flag disabled, and let users manually enabling it in `chrome://flags` if they want to.

ASTechnical:
You can open a Pull Request

csagan5:
Fixed in `97.0.4692.106`.

PR #1752, opened 2022-02-04
Master v98

uazo:
this is my version of the patches for v98
there are no noticeable differences in the patches, you can check each one.
i started from my v97 so it could be complex because you will notice too the bug fixes that I gradually fixed.

this is the list:

## changed patchs

**Add an always-incognito mode**
fixes #1496 
You will also find the preference in native (prefs :: kAlwaysIncognitoEnabled) and a code cleanup in general eliminating the superfluous parts. In addition I have added the possibility for the user to enable autocomplete in the search if he wants to.

**Add exit menu item**
If the user closes the menu, the incognito notification disappears.
don't fix #1503 completely but only the exit button

**Add history support in incognito mode**
there are no changes but I adjusted it based on the cleanup done for the always incognito mode patch

**Block gateway attacks via websockets**
fixes #1693

**Disable third-party origin trials**
I found another static method related to trial tokens, which I deleted

**Enable share intent**
fixes #1568

**Experimental user scripts support**
fixes #792 and I added the flag to enable evals in scripts

**Override UA client hint for model**
now completely disable client hints

**Show site settings for cookies, javascript and ads**
fixes #1355

## new patchs
**Add more private google search engine**
a new google search engine without parameters. I added an extra random delay every time when you send a key to the server

**Add an option that allows changing the number of days to keep navigation history**
and eventually completely disable the recording of the history

**Always ask for permissions**
fixes #1549

**Disable all predictors code**
fixes #1632

**Disable URL-keyed metrics and reporting**
Eliminate UKMs completely

**Enable Chrome Certificate Transparency and disable reporting**
fixes #1554

**Enable StrictOriginIsolation and SitePerProcess**
need for #1720

**Move top toolbar to bottom**
version for 98

**Remove fetching google fonts from gms**
Remove the request for fonts via intent a gms

**Remove HashAffiliationFetcher**
In bromite it is already disabled, but this ensures it

**Implement UI for JIT site settings**
fixes #1720

PR #1755, opened 2022-02-04
Enable trivial auto var init

BayLee4:
Build with '-ftrivial-auto-var-init=zero' ~~and enable 'Strict Origin Isolation' by default~~.

Fix for #1749


ASTechnical:
No need to add enable Strict Origin Isolation in this patch, coz there is already pull request that does that see #1752

BayLee4:
@ASTechnical Thanks, I did not see it

csagan5:
Have you tried building with these patches and checking if the browser is functional?

BayLee4:
Fuck I messed with git.

BayLee4:
@csagan5 No I did not tested, but honestly there's no way it could break anything

csagan5:
@BayLee4 all changes must be tested; I will do some checks before release.

Thanks for your PR!

PR #1758, opened 2022-02-06
Migrate to using GitHub issue forms

TacoTheDank:
While updating my Bromite instance, I noticed that almost one-third of all issues opened are ones that have ignored the issue template. This makes me sad because this means a lot of people cannot read. I don't like when people cannot read.

Thus, I have decided to migrate / rewrite the existing templates into issue forms (a new GitHub feature). This forces the user to interact with an actual template and not a bunch of text they can just delete. This would also mean you will no longer need to use the bot to automatically close issues that don't follow the template (or only partially), as now users won't have a choice. As a result, this will lower the amount of issue spam and other garbage that people often create, letting you get to the issues that actually matter. This also helps the user by making the process much more UI-friendly.

I have mostly kept the important stuff from the previous templates. However, I have also taken some liberties and cleaned up a bunch of redundancies, grammar, and potential confusion that they had as well.

You can see a live demo on my fork: https://github.com/TacoTheDank/bromite/issues/new/choose

Tell me what you think! Feel free to suggest any changes.

Related info:
- https://github.blog/changelog/2021-06-23-issues-forms-beta-for-public-repositories/
- https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository


ASTechnical:
Awesome

csagan5:
Thanks @TacoTheDank, I wasn't aware of this new feature. Does it also work from mobile? Issue templates are notoriously not shown to mobile users.

I will merge your PR and make some changes.

TacoTheDank:
Oof I had completely missed that for some reason. Thanks for fixing the multiples thing.

It works for me on mobile, so it should be good.

I noticed that [issues are still being closed](https://github.com/bromite/bromite/issues/1794) by the bot, even after these changes were made. The bot should either be turned off or adjusted to account for the new issue templates.

csagan5:
The bot is working correctly; it will not be turned off.

In that case the submitter did not read carefully and did not update Bromite before submitting the bug.

TacoTheDank:
Oh, I see, ok 👍

git-em:
In the latest issue closed by the bot (https://github.com/bromite/bromite/issues/1804) the links provided in the [bot's comment](https://github.com/bromite/bromite/issues/1804#issuecomment-1044223840) are still pointing to the old `.md` templates, instead of the new `.yml` files.

csagan5:
It should use the correct URLs now.

PR #1795, opened 2022-02-13
Patches for v97

csagan5:
New patches for v97, builds not yet finished.


csagan5:
@uazo these are my patches for v97; I have also added your patch to fix the libatomic linking error
```
FAILED: chrome_crashpad_handler exe.unstripped/chrome_crashpad_handler exe.unstripped/chrome_crashpad_handler.map.gz 
...
ld.lld: error: unable to find library -latomic
clang++: error: linker command failed with exit code 1 (use -v to see invocation)
```

In [this comment](https://github.com/bromite/bromite/pull/1752#issuecomment-1030658761) I am asking what are the patches for v96 that were not included: please submit them separately because it is not possible to review patch changes together with the whole bulk of patch updates.

Over time we have gone from you stating "I am submitting these bulk patch updates for my own training" to "I require that you merge this ASAP, with my new patches included".

Let's go a few step backs here: if you submit a PR and it is not in draft mode then you are asking to merge it. That is great and I find your patches very helpful in some cases (maintenance of your own contributed patches or other patches that are difficult to port), however you cannot expect these huge PRs to be merged: they need to be split:
* one bulk PR that only updates the patches from one major version to another and makes no other changes at all; if you have to make some changes because of necessity, **document them**
* individual PRs (which can be on top of your bulk PR) for the bugs fixed or features introduced

I understand this is more work, but it is the only way we can review the changes and the only way we can also update a CHANGELOG.

Each and every of your major version update PRs has not documented the extra changes (in some cases minor and in some cases substantial). For example if one file you deem not necessary to include anymore in the patch, it has to be documented.
If in another case you find that it is possible to exclude a file from compiling and thus the patch hunks are not needed, it has to be documented.

For v98 I will try to go through the fixes in #1752 once I get there but for v99 and forward we should follow the above.

csagan5:
> `Fix-StartupTabPreloader-crash.patch` to fix an assert at startup due to the removal of metrics

Which metrics? I think this patch should be merged with the patch that removes such metrics.

csagan5:
> 1. you probably have to change `Add-menu-item-to-bookmark-all-tabs.patch` like that

Already done.

> 3\. in `Add-option-to-not-persist-tabs-across-sessions.patch` I think you have the wrong position of
> 
> it should not be in `createManagedPreferenceDelegate()` but in `updatePreferences()`

Great, thanks; will fix this.

> 3\. try to verify `Allow-building-without-enable_reporting.patch`, does not seem correct:

I have used your version since then, it should be correct now.

> 4\. in `Disable-safe-browsing.patch` why did you change the `kEnableCrowdDenyTriggering` flag to false? that is, what is it for?

It is necessary because in the original patch this was disabled, in the new version the flag has moved thus it has to be disabled also there.
It is not a new feature, you can read here: https://source.chromium.org/chromium/chromium/src/+/main:chrome/browser/permissions/quiet_notification_permission_ui_config.h;l=36?q=kEnableCrowdDenyTriggering&sq=&ss=chromium%2Fchromium%2Fsrc

It should not trigger but better to disable it to avoid runtime bugs and similar.

> 7\. in `JIT-less-toggle.patch` I think the position of this is wrong, should be before the if

This also confused me, but the later patch `Disable privacy sandbox` removes the surrounding block so it ends up being correct. It could be solved by moving the order of patches.

> 9\. in `User-agent-customization.patch` it seems to me that this is incorrect

I have used your version since then; what is a "greased" user agent anyways?

uazo:
```
// Generate a pseudo-random permutation of the following brand/version pairs:
//   1. The base project (i.e. Chromium)
//   2. The browser brand, if available
//   3. A randomized string containing GREASE characters to ensure proper
//      header parsing, along with an arbitrarily low version to ensure proper
//      version checking.
```
ah, I don't know, maybe your version is better, I'll check.

uazo:
> Perhaps we need to hide this hint as well (or keep it).

I would disable the all clienthints, what I like least is [Accept-CH](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-CH) and in 98 it is active.

`The Accept-CH header may be set by a server to specify which headers a client should include in subsequent requests.`

I have already checked, it is per session until the browser is restarted but I do not like it because it can make the site understand that I have visited it other times (in incognito browsing if I close all the tabs without restarting I clear the cache).

uazo:
> Doesn't the site have already the IP address for that?

that information can be is not so useful, if the user never activates the wifi and go through the cellular network which is under a proxy of provider (as in my case, my provider, among other things, have proxy often located in france)

csagan5:
> in incognito browsing if I close all the tabs without restarting I clear the cache but not the clienthints

I think we need more information about this, and to check what upstream is doing about it. Perhaps there are already countermeasures or some being studied.

Please see my question [here](https://github.com/bromite/bromite/pull/1795#issuecomment-1038459807) and this one also:

> In [this comment](https://github.com/bromite/bromite/pull/1752#issuecomment-1030658761) I am asking what are the patches for v96 that were not included

uazo:
>> Fix-StartupTabPreloader-crash.patch to fix an assert at startup due to the removal of metrics
>
> Which metrics? I think this patch should be merged with the patch that removes such metrics.

the crash is due to the assert since `onFirstNavigationStart()` [is never called](https://source.chromium.org/chromium/chromium/src/+/main:chrome/android/java/src/org/chromium/chrome/browser/init/StartupTabPreloader.java;l=159;drc=a875629c29157e817f135dbd9acef79727303a34;bpv=1;bpt=1)
```
java.lang.AssertionError
   at org.chromium.chrome.browser.init.StartupTabPreloader.onFirstNavigationCommit(StartupTabPreloader.java:173)
   at org.chromium.chrome.browser.metrics.ActivityTabStartupMetricsTracker.registerFinishNavigation(ActivityTabStartupMetricsTracker.java:320)
   at org.chromium.chrome.browser.metrics.ActivityTabStartupMetricsTracker.access$300(ActivityTabStartupMetricsTracker.java:27)
   at org.chromium.chrome.browser.metrics.ActivityTabStartupMetricsTracker$1.onDidFinishNavigation(ActivityTabStartupMetricsTracker.java:178)
   at org.chromium.chrome.browser.tab.TabWebContentsObserver$Observer.didFinishNavigation(TabWebContentsObserver.java:323)
   at org.chromium.content.browser.webcontents.WebContentsObserverProxy.didFinishNavigation(WebContentsObserverProxy.java:122)
   at android.os.MessageQueue.nativePollOnce(Native Method)
   at android.os.MessageQueue.next(MessageQueue.java:326)
   at android.os.Looper.loop(Looper.java:160)
   at android.app.ActivityThread.main(ActivityThread.java:6669)
   at java.lang.reflect.Method.invoke(Native Method)
   at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493)
   at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:858)
```

the cause is the inhibition of the metrics, which does not start `onNewNavigation` [java](https://source.chromium.org/chromium/chromium/src/+/main:chrome/android/java/src/org/chromium/chrome/browser/metrics/PageLoadMetrics.java;l=171;bpv=1;bpt=1?q=onNewNavigation) [native](https://source.chromium.org/chromium/chromium/src/+/main:chrome/browser/page_load_metrics/observers/android_page_load_metrics_observer.cc;l=148;drc=a875629c29157e817f135dbd9acef79727303a34;bpv=1;bpt=1)

uazo:
>  I am asking what are the patches for v96 that were not included

I didn't answer you because I don't understand the question.
everything between v97 and v98 could be merged, but being already v98 with the development environment I can't easily get you the correct version:

image

that was why I was asking you when v98 would be released, if before the stable v99 so that I could get you the correct version or if I should have directly made you the version for v99, but I have explained myself badly (or you have misunderstood) ...
so what do we do?

csagan5:
> that was why I was asking you when v98 would be released, if before the stable v99 so that I could get you the correct version or if I should have directly made you the version for v99, but I have explained myself badly (or you have misunderstood) ...
> so what do we do?

v98 will be released after v97, that is the answer. We do not skip any major version. Feel free to wait then; v99 will also be released after v98.

csagan5:
> > I am asking what are the patches for v96 that were not included
> 
> I didn't answer you because I don't understand the question.

You can also answer "I don't understand the question", which is better than no answer (I am trying to not leave anything behind from v96 or v97, since you have put effort in developing fixes/patches!)

csagan5:
> > > Fix-StartupTabPreloader-crash.patch to fix an assert at startup due to the removal of metrics
> > 
> > 
> > Which metrics? I think this patch should be merged with the patch that removes such metrics.
> 
> the crash is due to the assert since `onFirstNavigationStart()` [is never called](https://source.chromium.org/chromium/chromium/src/+/main:chrome/android/java/src/org/chromium/chrome/browser/init/StartupTabPreloader.java;l=159;drc=a875629c29157e817f135dbd9acef79727303a34;bpv=1;bpt=1)
> 
> ```
> java.lang.AssertionError
>    at org.chromium.chrome.browser.init.StartupTabPreloader.onFirstNavigationCommit(StartupTabPreloader.java:173)
>    at org.chromium.chrome.browser.metrics.ActivityTabStartupMetricsTracker.registerFinishNavigation(ActivityTabStartupMetricsTracker.java:320)
>    at org.chromium.chrome.browser.metrics.ActivityTabStartupMetricsTracker.access$300(ActivityTabStartupMetricsTracker.java:27)
>    at org.chromium.chrome.browser.metrics.ActivityTabStartupMetricsTracker$1.onDidFinishNavigation(ActivityTabStartupMetricsTracker.java:178)
>    at org.chromium.chrome.browser.tab.TabWebContentsObserver$Observer.didFinishNavigation(TabWebContentsObserver.java:323)
>    at org.chromium.content.browser.webcontents.WebContentsObserverProxy.didFinishNavigation(WebContentsObserverProxy.java:122)
>    at android.os.MessageQueue.nativePollOnce(Native Method)
>    at android.os.MessageQueue.next(MessageQueue.java:326)
>    at android.os.Looper.loop(Looper.java:160)
>    at android.app.ActivityThread.main(ActivityThread.java:6669)
>    at java.lang.reflect.Method.invoke(Native Method)
>    at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493)
>    at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:858)
> ```
> 
> the cause is the inhibition of the metrics, which does not start `onNewNavigation` [java](https://source.chromium.org/chromium/chromium/src/+/main:chrome/android/java/src/org/chromium/chrome/browser/metrics/PageLoadMetrics.java;l=171;bpv=1;bpt=1?q=onNewNavigation) [native](https://source.chromium.org/chromium/chromium/src/+/main:chrome/browser/page_load_metrics/observers/android_page_load_metrics_observer.cc;l=148;drc=a875629c29157e817f135dbd9acef79727303a34;bpv=1;bpt=1)

Still cannot pinpoint which patch causes these navigation events to be skipped :thinking: 
I have been looking into `Disable-safe-browsing.patch`, `Skip-the-first-run-and-metrics.patch` and `Remove-signin-and-data-saver-integrations.patch`.

uazo:
> Still cannot pinpoint which patch causes these navigation events to be skipped

it's probably the [floc ](https://github.com/bromite/bromite/blob/master/build/patches/Disable-FLoC.patch#L135)

uazo:
> Still cannot pinpoint which patch causes these navigation events to be skipped

found!

https://github.com/bromite/bromite/blob/master/build/patches/Bromite-AdBlockUpdaterService.patch#L577

there has always been ..

csagan5:
Thanks for finding this; I had already checked the throttles removal code in the safe browsing patch but didn't remember about this one.

I will move this block together with your patch as a separate one.

uazo:
> I will move this block together with your patch as a separate one.

possibly you decide whether to disable [this flag](https://github.com/uazo/bromite/blob/master-v98/build/patches/Disable-ukm.patch#L30) as well

csagan5:
> > I will move this block together with your patch as a separate one.
> 
> possibly you decide whether to disable [this flag](https://github.com/uazo/bromite/blob/master-v98/build/patches/Disable-ukm.patch#L30) as well

Sure; but are those metrics currently being sent?

uazo:
> but are those metrics currently being sent?

no, follow [this](https://source.chromium.org/chromium/chromium/src/+/main:components/ukm/ukm_reporting_service.cc;l=50;bpv=1;bpt=1)
but we can completely remove the code from
```
components/ukm/ukm_reporting_service.cc
components/embedder_support/android/metrics/android_metrics_service_client.cc
```
that I would like more

csagan5:
> > but are those metrics currently being sent?
> 
> no, follow [this](https://source.chromium.org/chromium/chromium/src/+/main:components/ukm/ukm_reporting_service.cc;l=50;bpv=1;bpt=1) but we can completely remove the code from
> 
I am asking because each patch should document whether it is effective or defensive.

> ```
> components/ukm/ukm_reporting_service.cc
> components/embedder_support/android/metrics/android_metrics_service_client.cc
> ```
> 
> that I would like more

Ok; for now I have moved all metrics-related changes (except for NTP) in a new stand-alone patch.

PR #1806, opened 2022-02-19
Patches for v98

csagan5:
Patches ported to v98, with very few changes/additions:
* re-introduce flag for text fragments
* re-introduce content feature flag to disable field trials
* disable idle detection
* merge pre-existing client hints patch into one and disable critical client hints
* refactor `DISALLOW_COPY_EXISTING` as http://google.github.io/styleguide/cppguide.html#Copyable_Movable_Types suggests (I checked #1752 about this, it was not documented and also done incorrectly)

Nothing else; will add more fixes as the build progresses; @uazo will also start looking at your work in #1752.


uazo:
ah oops, you didn't ask me to check... sorry ..

in any case I would say perfect!
as soon as you publish them I compile and then I fill the repo with my pull requests :)

csagan5:
> ah oops, you didn't ask me to check... sorry ..

It's ok, they are not ready yet, I am ironing out some build bugs.

> in any case I would say perfect!
> as soon as you publish them I compile and then I fill the repo with my pull requests :)

Thanks! That will help.

I noticed in the user agent patch you added an `if ((true)) return;` that I do not have, I wonder what that is for.

uazo:
> I noticed in the user agent patch you added an if ((true)) return; that I do not have, I wonder what that is for

because what the chromium code does is already done by the patch, plus the handling of the sticky mode

csagan5:
> > I noticed in the user agent patch you added an if ((true)) return; that I do not have, I wonder what that is for
> 
> because what the chromium code does is already done by the patch, plus the handling of the sticky mode

I do not understand what the change is for; is it an optimisation or is it functional?

PR #1821, opened 2022-02-22
Remove duplicate architectures

TacoTheDank:
These are one and the same.


csagan5:
They certainly are not.

csagan5:
I thought you meant that `arm` and `arm64` are the same, now I get what you mean; even if `arm` and `arm64` are aliases of `armeabi-v7a` and `arm64-v8a` it is more convenient to keep them around because they are easier to type.

PR #1826, opened 2022-02-23
Minor changes to v98 patches

uazo:
no functionality changed.

small changes:
- elimination of excess blank lines
- moving the position of `g_browser_process->adblock_updater();` in `Bromite AdBlockUpdaterService` 
- fix overlap between `Disable privacy sandbox` and `JIT-less toggle`
- changed to `raw_ptr<BookmarkPermanentNode> tabs_collection_node_ = nullptr`
- removed last `$google_play_services_package` in `Remove binary blob integrations`

### DIFF
--- a/build/patches/Add-a-proxy-configuration-page.patch
+++ b/build/patches/Add-a-proxy-configuration-page.patch
@@ -70,7 +70,7 @@ diff --git a/chrome/android/java/res/xml/privacy_preferences.xml b/chrome/androi
 diff --git a/chrome/android/java/src/org/chromium/chrome/browser/privacy/settings/PrivacySettings.java b/chrome/android/java/src/org/chromium/chrome/browser/privacy/settings/PrivacySettings.java
 --- a/chrome/android/java/src/org/chromium/chrome/browser/privacy/settings/PrivacySettings.java
 +++ b/chrome/android/java/src/org/chromium/chrome/browser/privacy/settings/PrivacySettings.java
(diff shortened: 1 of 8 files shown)


csagan5:
> * changed to `raw_ptr<BookmarkPermanentNode> tabs_collection_node_ = nullptr`

Reason? It is not safe to use those if there is no logic to correctly free the objects from memory.

csagan5:
> > It is not safe to use those if there is no logic to correctly free the objects from memory.
> 
> from what I understand instead the use of raw_ptr is [better](https://chromium.googlesource.com/chromium/src/+/ddc017f9569973a731a574be4199d8400616f5a5/base/memory/raw_ptr.md)

My bad, I thought none of those benefits were available for `raw_ptr`.

> we have no memory leak

We will still have memory leaks if the freed object contains pointers which are not `raw_ptr`, right?
Is this the case here?

csagan5:
I will merge this but please let me know about the question above, memory leaks are serious issues.

uazo:
> We will still have memory leaks if the freed object contains pointers which are not raw_ptr, right?
Is this the case here?

no, not necessarily. it depends on what is connected to that pointer, if it is a C++ object there is basically no problem, as long as the references are correctly moved. in case a class can be copied, the suitable measures are already foreseen in the chromium code.
`raw_ptr` resets the value of the pointer to NULL when the "c++ garbage collector" is correctly managed otherwise even the use of that is useless.

PR #1828, opened 2022-02-23
fix Block gateway attacks via websockets

uazo:
I added the missing points to the patch.

I don't understand exactly how it worked for me before, but I double checked and now it works

### DIFF
--- a/build/patches/Block-gateway-attacks-via-websockets.patch
+++ b/build/patches/Block-gateway-attacks-via-websockets.patch
@@ -4,15 +4,15 @@ Subject: Block gateway attacks via websockets
 
 ---
  .../renderer/core/loader/base_fetch_context.h |  1 +
- .../core/loader/frame_fetch_context.cc        | 20 ++++++++++++++
+ .../core/loader/frame_fetch_context.cc        | 20 +++++++++++++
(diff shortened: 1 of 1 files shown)


csagan5:
@uazo the function was removed completely here: https://github.com/chromium/chromium/commit/10b906d8372148cdbae36b2d958447c2064cdf07
The new upstream tracking issue is at https://bugs.chromium.org/p/chromium/issues/detail?id=1295244

It mentions that check can be bypassed using a service like `sslip.io`; is that true? I would like to understand how.

Meanwhile for v100 I will revert that commit.

uazo:
because network_`utils::IsReservedIPAddress()` reads the ip from the url without going through the dns, so while "192.168.1.1.sslip.io" returns `IPAddressSpace::kPublic` in reality, pointing to 192.168.1.1, it is effectively private

csagan5:
I understand, thanks. I will give a look before or after next release.

csagan5:
The upstream tracking bug to fix this properly is here: https://bugs.chromium.org/p/chromium/issues/detail?id=590714

It is a lot of work; for now I will mention this issue in the patch description.

PR #1829, opened 2022-02-23
Multiple fixes to user script patch

### LINKED ISSUE #792: User scripts component

No, but it could become

No.

as some users have said, I also find privacy pop-ups boring, especially google ones, since in my phone they are quite lagged

initially I thought about reusing the Inject-scripts-for-AMP-tracking-ads-and-video.patch but then I saw that there is the possibility to activate the support for userscripts, which is somehow better because it leaves no trace in the dom, as opposed to that patch.

as a personal test (and also a bit of a challenge) I tried to reactivate user scripts in bromite (https://www.chromium.org/developers/design-documents/user-scripts) taking them from the extensions and moving them to a new one component, managing to make them work (https://github.com/uazo/bromite/blob/gitpages/userscripts_v2.7z my work)
Obviously, as my first test, there is no ui (and by the way it could be quite long to implement it), we could think to proceed in steps.

I wanted to understand if there was interest and, if so, if we wanted to discuss how to implement it, also considering the security aspects, updates and possible integration with chromium developer tools.

current state:
- [x] base code + reading from folder
  - [x] add details for unmatch header
  - [x] bug DOCUMENT_START not work
- [x] preferences (enable/disable)
  - [x] add link to help page (actually https://github.com/bromite/bromite/wiki/UserScripts)
- [x] ui management (enable/disable/remove script) 
  - [x] consider changing proguard settings for R (see https://stackoverflow.com/a/2889844)
  - [x] add icons (enable/disable)
  - [x] disable add button if feature is disabled
  - [x] bug enable/disable need restart
- [x] add script from file dialog
  - [x] remove excessive permissions prompt from filedialog (camera + audio)
  - [x] check how to transform a content:// (URI) to file:// (PATH)
  - [x] get real script filename from a content://
  - [x] when importing from file, skip non ascii chars used in description to prevent [FATAL:web_string.cc(112)] Check failed: ContainsOnlyASCII()
  - [x] reload list after add and show toast with error (check how works CallByNative)
  - [x] check if a file is real user script (es refuse txt without javascript)
- [x] understand if installation from download makes sense (entry point https://source.chromium.org/chromium/chromium/src/+/master:chrome/android/java/src/org/chromium/chrome/browser/download/DownloadUtils.java;l=360), actually we have to download the file
  - [x] check how to request permission from the user
  - [x] not working
- [x] add feature flag
- [x] consider if we want to add a way to prevent crash in the startup
- [ ] code cleaning
  - [x] remove unused code
  - [x] remove unused BrowserContextHandle from bridge
  - [ ] try to simply rendering code
  - [x] change all ifdef removing extension word
  - [x] add a flag for logging
  - [x] add file name to pickle (file) so we can log script name at injection time
  - [x] check compiling webview
- [ ] ui management (2)
  - [ ] script details
  - [ ] disable in incognito
  - [x] view source script
  - [ ] check if we can use an "interstitial" page to install a userscript
  - [ ] localize error strings
- [ ] security verification
  - [ ] consider whether to use a sha256 to verify a possible modification of the file without the user's knowledge (if we give the possibility to choose the scripts folder)
  - [x] check if we can deny network access to injected scripts (check ScriptContext)
  - [x] permit only content:// from pickerui
  - [x] block install intent from outside --> no, ask user
- [ ] understand if it makes sense to use zip files (for adding css directly, userscripts.js doesn't support them directly) code https://source.chromium.org/chromium/chromium/src/+/master:components/services/unzip/public/cpp/unzip.cc
- [ ] understand if self-updating makes sense
  - [ ] check how the source can be verified (script is unsigned by default)
  - [ ] or enable include metadata support
- [ ] add management in devtools (like snippets, code https://source.chromium.org/chromium/chromium/src/+/master:third_party/devtools-frontend/src/front_end/snippets/ScriptSnippetFileSystem.js)
  - [ ] add flag for "fall-back" in devtools (code https://source.chromium.org/chromium/chromium/src/+/master:chrome/browser/android/devtools_server.cc)
- [ ] add test cases
- [x] try with v88
- [x] try with v89
- [x] try with v90
- [x] try with v91
- [x] bug  [FATAL:utf_string_conversions.cc(338)] Check failed: IsStringASCII(utf16). Tampermonkey • Userscript Sources
- [x] handling unicode scripts

uazo:
- added support for unicode scripts
- allow the use of multiple scripts with the same url patterns 
- added more detailed log 
- disabled access to native pages

fix for #792

### DIFF
--- a/build/patches/Experimental-user-scripts-support.patch
+++ b/build/patches/Experimental-user-scripts-support.patch
@@ -95,7 +95,7 @@ See also: components/user_scripts/README.md
  components/user_scripts/common/url_pattern.h  | 302 +++++++
  .../user_scripts/common/url_pattern_set.cc    | 335 ++++++++
  .../user_scripts/common/url_pattern_set.h     | 160 ++++
- components/user_scripts/common/user_script.cc | 325 +++++++
+ components/user_scripts/common/user_script.cc | 329 +++++++
  components/user_scripts/common/user_script.h  | 403 +++++++++
  .../common/user_scripts_features.cc           |  32 +
  .../common/user_scripts_features.h            |  34 +
@@ -133,7 +133,7 @@ See also: components/user_scripts/README.md
  .../renderer/web_ui_injection_host.h          |  27 +
  .../strings/userscripts_strings.grdp          |  55 ++
  tools/gritsettings/resource_ids.spec          |   6 +
- 111 files changed, 9569 insertions(+), 2 deletions(-)
+ 111 files changed, 9573 insertions(+), 2 deletions(-)
  create mode 100644 components/user_scripts/README.md
  create mode 100755 components/user_scripts/android/BUILD.gn
  create mode 100644 components/user_scripts/android/java/res/layout/accept_script_item.xml
@@ -262,7 +262,7 @@ diff --git a/chrome/android/java/res/xml/main_preferences.xml b/chrome/android/j
 diff --git a/chrome/android/java/src/org/chromium/chrome/browser/download/DownloadUtils.java b/chrome/android/java/src/org/chromium/chrome/browser/download/DownloadUtils.java
 --- a/chrome/android/java/src/org/chromium/chrome/browser/download/DownloadUtils.java
 +++ b/chrome/android/java/src/org/chromium/chrome/browser/download/DownloadUtils.java
-@@ -69,6 +69,7 @@ import org.chromium.content_public.browser.BrowserStartupController;
+@@ -68,6 +68,7 @@ import org.chromium.content_public.browser.BrowserStartupController;
  import org.chromium.content_public.browser.LoadUrlParams;
  import org.chromium.ui.base.DeviceFormFactor;
  import org.chromium.ui.widget.Toast;
@@ -6407,7 +6407,7 @@ diff --git a/components/user_scripts/common/user_script.cc b/components/user_scr
 new file mode 100755
 --- /dev/null
 +++ b/components/user_scripts/common/user_script.cc
-@@ -0,0 +1,325 @@
+@@ -0,0 +1,329 @@
 +// Copyright 2013 The Chromium Authors. All rights reserved.
 +// Use of this source code is governed by a BSD-style license that can be
 +// found in the LICENSE file.
@@ -6515,7 +6515,11 @@ new file mode 100755
 +}
 +
 +bool UserScript::MatchesURL(const GURL& url) const {
-+  // exclude schemas not valid
++  // Since the injecton is also provided for native pages,
++  // we must verify that the render process does not include
++  // scripts in the schema that are not allowed
++
++  // we allow only URLPattern::SCHEME_HTTP(S)
 +  URLPattern pattern(kValidUserScriptSchemes);
 +  pattern.Parse(url.spec());
 +  if (!pattern.IsValidScheme(pattern.scheme()))
--- a/build/patches/Experimental-user-scripts-support.patch
+++ b/build/patches/Experimental-user-scripts-support.patch
@@ -72,7 +72,7 @@ See also: components/user_scripts/README.md
  .../user-script-ui/user-scripts-ui.js         |   9 +
  .../browser/ui/user_scripts_ui.cc             | 147 ++++
  .../user_scripts/browser/ui/user_scripts_ui.h |  37 +
- .../browser/user_script_loader.cc             | 721 ++++++++++++++++
+ .../browser/user_script_loader.cc             | 714 ++++++++++++++++
  .../user_scripts/browser/user_script_loader.h | 169 ++++
  .../browser/user_script_pref_info.cc          |  34 +
  .../browser/user_script_pref_info.h           |  72 ++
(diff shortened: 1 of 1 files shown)


build/patches/Experimental-user-scripts-support.patch
```diff
@@ -3096,13 +3087,12 @@ new file mode 100755
...
 +
++    script->set_name(user_script_path.BaseName().value());
 +    script->set_match_origin_as_fallback(MatchOriginAsFallbackBehavior::kNever);
 +
-+    // remove unicode chars and set content into File
-+    stripUnicode(content);
 +    std::unique_ptr<UserScript::File> file(new UserScript::File());
 +    file->set_content(content);
-+    file->set_url(GURL(/*script_key*/ "script.js")); // name doesn't matter
++    file->set_url(GURL(base::StrCat({"https://userscripts/file/", script->name(), ".js"})));
```
csagan5, line 3095: What is this change for?

csagan5:
> fix handling of multiple scripts with more detailed log

for cases where an user script contains multiple scripts? Is that even possible?

> not allowed access to native pages

Can you explain what this is about?

JustAnUserLol:
> Can you explain what this is about?

I think he by native pages he mean chrome://version,chrome://flags other native flags.

build/patches/Experimental-user-scripts-support.patch
```diff
@@ -6525,6 +6515,12 @@ new file mode 100755
 +}
 +
 +bool UserScript::MatchesURL(const GURL& url) const {
++  // exclude schemas not valid
++  URLPattern pattern(kValidUserScriptSchemes);
++  pattern.Parse(url.spec());
++  if (!pattern.IsValidScheme(pattern.scheme()))
```
uazo, line 6521: > > not allowed access to native pages
>
> Can you explain what this is about?

[this condition](https://github.com/bromite/bromite/blob/e74523cd70d21875d5e2ec7412b0662708ca4cff/build/patches/Experimental-user-scripts-support.patch#L6451-L6457) was not controlled sufficiently.
with these lines I fix it

uazo:
> for cases where an user script contains multiple scripts? Is that even possible?

no, in case the user has more than one userscript for the same url pattern

uazo:
> I think by native pages he mean chrome://version,chrome://flags, other native flags.

user scripts can be injected in all natives that have webcontents, because also the android ui is considered native but userscripts cannot be injected in that. to understand better, in the desktop version all the settings interface, in bromite, for example, only the proxy setting. interstitials are another example of a native with webcontent

build/patches/Experimental-user-scripts-support.patch
csagan5: Can you please remove the commented out code?

> user scripts can be injected in all natives that have webcontents, because also the android ui is considered native but userscripts cannot be injected in that. to understand better, in the desktop version all the settings interface, in bromite, for example, only the proxy setting. interstitials are another example of a native with webcontent

Is this change closing a security problem? If yes, what is the impact on current and previous versions of Bromite?

build/patches/Experimental-user-scripts-support.patch
csagan5: Can you please add this to the PR description? So that I can document it in the changelog.

Can `script->name()` possibly contain dots or `/` characters?

uazo: > Can you please add this to the PR description?

fixed.

> Can script->name() possibly contain dots or / characters?

I don't know, i'll try.

build/patches/Experimental-user-scripts-support.patch
csagan5: > > Can you please remove the commented out code?

> it's incorrect? or don't like comments in the patches? I need them to remind me of the changes after some time

You can never tell whether it was commented and forgot to uncomment by mistake. If you need to remind of the change it's better to add a human-readable comment.

> > Is this change closing a security problem?
> 
> if you want to call it that

Yes, we will call everything for what it is.

> > If yes, what is the impact on current and previous versions of Bromite?
> 
> theoretically nothing, but potentially.
the scripts are isolated so they cannot communicate with the other javascript objects exposed to the native (who are also isolated), but they are injected into an environment that is less protected (and that allows more) than normal (but in the desktop, extensions are allowed access to native pages).
> 
> and as always, it depends on the script, and in any case in bromite it is not needed

Sorry, I didn't get an answer from this. How can this be exploited?

csagan5: You can leave the code commented but then add a human-readable comment on top explaining why they are left there?

build/patches/Experimental-user-scripts-support.patch
uazo: tried, the point can be present but no problem emerges.
`/` cannot be present because between the `FilePath.BaseName()` separators used to retrieve the name

build/patches/Experimental-user-scripts-support.patch
csagan5: Can you please explain the potential way to exploit this in current Bromite?

uazo: for example in the proxy patch, a bridge to the native with `window.ProxyConfigView` is added in `proxy_config.js`.
If a userscript had the ability to use those methods, they would have access to the configuration (should the user enter that page). But userscripts is isolated and therefore does not have access to functions added by other isolated scripts, at least in theory.

csagan5: > But userscripts is isolated and therefore does not have access to functions added by other isolated scripts, at least in theory.

In theory or in practice?

uazo: unfortunately I don't know, the code isolation in v8/blink is a bit complex to read, an exploit should be tried or I can try to check in tests suite too

Issue #1781, opened 2022-02-09
To consider removing NavigatorUAData.getHighEntropyValues()


uazo:
### Is your feature request related to privacy?

Yes

### Is there a patch available for this feature somewhere?

No, but it's trivial

### Describe the solution you would like

since even with ua reduction it is possible to obtain the browser version (see https://developer.mozilla.org/en-US/docs/Web/API/NavigatorUAData/getHighEntropyValues) is perhaps better completely remove the method from the idl

### Describe alternatives you have considered

removing the method could lead to unmanaged errors in scripts. While it may be fair from a privacy standpoint, we may break some sites.
the alternative is to delete only the [uaFullVersion](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/frame/navigator_ua_data.cc;l=173;drc=404e8d654e8b26336d2cb103b9c21faecbf7f73a?q=NavigatorUAData::getHighEntropyValues&ss=chromium%2Fchromium%2Fsrc
)

csagan5:
> since even with ua reduction it is possible to obtain the browser version (see https://developer.mozilla.org/en-US/docs/Web/API/NavigatorUAData/getHighEntropyValues)

Hiding the browser version is not a goal of Bromite and it is not possible anyways because of the adblock filters.



> the alternative is to delete only the [uaFullVersion](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/frame/navigator_ua_data.cc;l=173;drc=404e8d654e8b26336d2cb103b9c21faecbf7f73a?q=NavigatorUAData::getHighEntropyValues&ss=chromium%2Fchromium%2Fsrc)

The client hint for model was disabled here: https://github.com/bromite/bromite/blob/master/CHANGELOG.md#9304577110

So we could also disable this one, or provide a dummy version.

uazo:
> So we could also disable this one, or provide a dummy version.

perfect, I'll do so.

> Hiding the browser version is not a goal of Bromite

ok

> is not possible anyways because of the adblock filters.

can you explain to me why?

csagan5:
> > is not possible anyways because of the adblock filters.
> 
> can you explain to me why?

It is possible to verify which ad filters are present by checking what resources do not get loaded. This allows to detect Bromite; it is not a secret and was discussed in multiple places; it is also the reason why Bromite does not try to look like any other browser.

uazo:
> This allows to detect Bromite;

[...]

PR #1830, opened 2022-02-23
Proposal: complete disabling of client hints

uazo:
I make you the proposal to completely disable client hints, since they would make it possible to make the purposes of the ua reduction ineffective.
in particular, the management of accept ch in the always incognito mode, which by keeping the requests of the sites in the content settings, would allow to know if a user has already visited the site in the same session.
currently the management of client hints is under development and therefore it could be possible that the patch will need to be revised over time.

### DIFF
--- a/build/patches/Client-hints-overrides.patch
+++ b/build/patches/Client-hints-overrides.patch
@@ -6,11 +6,30 @@ Disable critical client hints
 Hard-code model to SAMSUNG SM-G960U
 use Google Chrome branding for client hints
 ---
- components/embedder_support/user_agent_utils.cc | 4 +---
- content/common/user_agent.cc                    | 7 +------
- content/public/common/content_features.cc       | 2 +-
- 3 files changed, 3 insertions(+), 10 deletions(-)
+ chrome/browser/prefs/browser_prefs.cc                      | 2 +-
+ components/embedder_support/user_agent_utils.cc            | 4 +---
+ content/browser/client_hints/client_hints.cc               | 1 +
+ content/common/user_agent.cc                               | 7 +------
+ content/public/common/content_features.cc                  | 2 +-
+ services/network/public/cpp/client_hints.cc                | 1 +
+ services/network/public/cpp/features.cc                    | 2 +-
+ third_party/blink/common/client_hints/client_hints.cc      | 1 +
+ .../blink/common/client_hints/enabled_client_hints.cc      | 4 +++-
+ third_party/blink/common/features.cc                       | 4 ++--
+ 10 files changed, 13 insertions(+), 15 deletions(-)
 
+diff --git a/chrome/browser/prefs/browser_prefs.cc b/chrome/browser/prefs/browser_prefs.cc
+--- a/chrome/browser/prefs/browser_prefs.cc
++++ b/chrome/browser/prefs/browser_prefs.cc
+@@ -716,7 +716,7 @@ void RegisterLocalStatePrefsForMigration(PrefRegistrySimple* registry) {
+   registry->RegisterIntegerPref(kStabilityDebuggerPresent, 0);
+   registry->RegisterIntegerPref(kStabilityDebuggerNotPresent, 0);
+ 
+-  registry->RegisterBooleanPref(kUserAgentClientHintsEnabled, true);
++  registry->RegisterBooleanPref(kUserAgentClientHintsEnabled, false);
+ 
+   registry->RegisterBooleanPref(kCloudPolicyOverridesPlatformPolicy, false);
+ 
 diff --git a/components/embedder_support/user_agent_utils.cc b/components/embedder_support/user_agent_utils.cc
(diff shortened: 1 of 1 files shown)


csagan5:
Related: #1781

In current patch we have:
* hard-coded fixed model
* critical client hints disabled
* brand set to Google Chrome

As explained there in #1781 I expect major breakage of websites by disabling the functionality like this, I would rather have `NotAllowedError` implemented after upstream starts using it or provide fake values.

> since they would make it possible to make the purposes of the ua reduction ineffective.

Can you explain how?

> in particular, the management of accept ch in the always incognito mode, which by keeping the requests of the sites in the content settings, would allow to know if a user has already visited the site in the same session.

Also this, not clear to me.

uazo:
> critical client hints disabled

in my opinion all hints must be disabled, not just the management of the critical ones. it is true that they can potentially retrieve the same information via javascript (and the client-hints, with javascript disabled, are disabled), but the scripts can intercept them with the adblock while the headers cannot be blocked.

>> since they would make it possible to make the purposes of the ua reduction ineffective.
>>
> Can you explain how?
```
Rather than broadcasting this data to everyone, all the time, user agents can make 
reasonable decisions about how to respond to given sites' requests for more granular 
data, reducing the [passive fingerprinting](https://w3c.github.io/fingerprinting-guidance/#dfn-passive-fingerprinting) 
surface area exposed to the network
```
[source](https://wicg.github.io/ua-client-hints/)
since the ua is reduced, I do not pass information, but by activating the client hints I could instead do it when a site request it, why should I?

>> in particular, the management of accept ch in the always incognito mode, which by keeping the requests of the sites in the content settings, would allow to know if a user has already visited the site in the same session.
>>
> Also this, not clear to me.

normally not all client hints are sended but only the [this list](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/common/client_hints/client_hints.cc;drc=f89a27501d4dcedca8af38dd350279eaab971c62;bpv=1;bpt=1;l=119) but the site can request them through [critical ones](https://github.com/WICG/client-hints-infrastructure/blob/main/reliability.md) and [accept-ch](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-CH)
the site request is persisted in a [session content settings](https://source.chromium.org/chromium/chromium/src/+/main:components/client_hints/browser/client_hints.cc;l=193-198;bpv=1;bpt=0) or [always persisted](https://source.chromium.org/chromium/chromium/src/+/main:components/client_hints/browser/client_hints.cc;l=188-191;bpv=1;bpt=0). the first (default) is destroyed only when the profile is destroyed (in incognito) or on restart (in normal mode).
so a site can know if there is a subsequent visit by verifying if it is necessary to resend accept-ch or not, why do I have to tell them?

csagan5:
> > critical client hints disabled
> 
> in my opinion all hints must be disabled, not just the management of the critical ones. it is true that they can potentially retrieve the same information via javascript (and the client-hints, with javascript disabled, are disabled), but the scripts can intercept them with the adblock while the headers cannot be blocked.

The correct way to disable them is to throw a `NotAllowedError`, see
> As explained there in https://github.com/bromite/bromite/issues/1781 I expect major breakage of websites by disabling the functionality like this, I would rather have NotAllowedError implemented after upstream starts using it or provide fake values.

> [source](https://wicg.github.io/ua-client-hints/)
> since the ua is reduced, I do not pass information, but by activating the client hints I could instead do it when a site request it, why should I?

Because the same information is provided? e.g. still reduced UA (correct term is "greased"?)

> > > in particular, the management of accept ch in the always incognito mode, which by keeping the requests of the sites in the content settings, would allow to know if a user has already visited the site in the same session.
> > 
> > 
> > Also this, not clear to me.
> 
> normally not all client hints are sent but only the [this list](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/common/client_hints/client_hints.cc;drc=f89a27501d4dcedca8af38dd350279eaab971c62;bpv=1;bpt=1;l=119) but the site can request them through [critical ones](https://github.com/WICG/client-hints-infrastructure/blob/main/reliability.md) and [accept-ch](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-CH) the site request is persisted in a [session content settings](https://source.chromium.org/chromium/chromium/src/+/main:components/client_hints/browser/client_hints.cc;l=193-198;bpv=1;bpt=0) or [always persisted](https://source.chromium.org/chromium/chromium/src/+/main:components/client_hints/browser/client_hints.cc;l=188-191;bpv=1;bpt=0). the first (default) is destroyed only when the profile is destroyed (in incognito, by closing all tabs) or on restart (in normal mode). so a site can know if there is a subsequent visit by verifying if it is necessary to resend accept-ch or not, why do I have to tell them?

Is there already an open upstream Chromium issue about this?

In any case: until `NotAllowedError` is implemented by upstream it's better to provide fake values to avoid client-side breakage.

PR #1831, opened 2022-02-23
Fix another way to activate origin trials

uazo:
based on the research I have done, unfortunately it is still possible to activate origin trials.
in particular, the static method `TrialTokenValidator::IsTrialPossibleOnOrigin()` must be disabled.
also the assumption that I had made about the source is not correct, so it is necessary to inhibit also `ProcessHttpEquivOriginTrial`

### DIFF
--- a/build/patches/Disable-third-party-origin-trials.patch
+++ b/build/patches/Disable-third-party-origin-trials.patch
@@ -3,16 +3,16 @@ Date: Thu, 4 Nov 2021 09:19:24 +0100
 Subject: Disable third-party origin trials
 
 ---
- .../origin_trials/origin_trial_policy_impl.cc   |  7 ++++---
- .../origin_trials/origin_trial_policy_impl.h    |  2 +-
- .../browser/renderer_host/navigation_request.cc |  1 -
- content/renderer/render_frame_impl.cc           | 17 +----------------
- .../shell/common/shell_origin_trial_policy.cc   | 10 ++++------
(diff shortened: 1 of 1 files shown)


build/patches/Disable-third-party-origin-trials.patch
```diff
@@ -255,19 +256,5 @@ diff --git a/third_party/blink/renderer/core/origin_trials/origin_trial_context.
...
  bool OriginTrialContext::IsFeatureEnabled(OriginTrialFeature feature) const {
-diff --git a/third_party/blink/renderer/platform/runtime_enabled_features.json5 b/third_party/blink/renderer/platform/runtime_enabled_features.json5
---- a/third_party/blink/renderer/platform/runtime_enabled_features.json5
-+++ b/third_party/blink/renderer/platform/runtime_enabled_features.json5
-@@ -2272,6 +2272,10 @@
-       name: "TextFragmentTapOpensContextMenu",
-       status: {"Android": "stable"},
-     },
-+    {
-+      name: "ThirdPartyOriginTrials",
```
csagan5, line 266: If we remove this then `RuntimeEnabledFeatures::ThirdPartyOriginTrialsEnabled()` will not exist anymore; I think we should keep this, not just for the compilation error but for visibility.

uazo: > not just for the compilation error

but that's what I want, I want a build error.
the alternative is to periodically check that patch to see if the flag is otherwise used, instead so it is the compiler that warns us.

Issue #1819, opened 2022-02-21
Disable JIT by default


JustAnUserLol:
### Preliminary checklist

- [X] I have read the [README](https://github.com/bromite/bromite/blob/master/README.md)
- [X] I have read the [FAQs](https://github.com/bromite/bromite/blob/master/FAQ.md).
- [X] I have searched [existing issues](https://github.com/bromite/bromite/issues) for my feature request. This is a new issue (NOT a duplicate) and is not related to another issue.

### Is your feature request related to privacy?

Yes

### Is there a patch available for this feature somewhere?

No

### Describe the solution you would like

Disable JIT by default, Disabling JIT kills half of the bugs in V8 and Eliminates Significant attack surface.

### Describe alternatives you have considered

Let the users manually disable JIT in settings (Which is already implemented).

themoonmeetsthesun:
Doesn't it impact performance?

JustAnUserLol:
> Doesn't it impact performance?

Only on Javascript heavy sites

themoonmeetsthesun:
So wouldn't the login pages of various services become slow when they rely a lot on javascript?

themoonmeetsthesun:
Also is there a need to do this? Cause the setting is already there. I don't understand your issue.

uazo:
I have developed a patch that allows you to insert a new site setting for javascript jit, @csagan5 are you interested? would allow you to remove the global flag and leave the user the choice of being able to activate or deactivate it.

JustAnUserLol:
Fixed by #1832

csagan5:
Fixed in `99.0.4844.55`.

PR #1832, opened 2022-02-23
Implement jit toggle patch

### LINKED ISSUE #1720: Per Site JIT Toggle
Somewhat

No

Due to disable JIT option enabled all the time, some sites for ex : protonmail won't let me login until I disable the disable JIT option, however if you provided the enable jit per-site toggle it would be helpful, as I have to restart the browser twice just to login in such sites.
Thanks :)

None

uazo:
here is a possible patch for activating the site setting for jit.
to make it work it is necessary to activate `kStrictOriginIsolation` and `kSitePerProcess` otherwise the setting would only work with the scheme and eTLD+1.
I tried for almost a month on my 2gb memory cell phone, not detecting any problems.

fix for #1720

### DIFF
--- a/build/patches/Enable-StrictOriginIsolation-and-SitePerProcess-flags.patch
+++ b/build/patches/Enable-StrictOriginIsolation-and-SitePerProcess-flags.patch
@@ -1,0 +1,65 @@
+From: uazo <uazo@users.noreply.github.com>
+Date: Sat, 29 Jan 2022 15:25:19 +0000
+Subject: Enable StrictOriginIsolation and SitePerProcess
+
+---
+ chrome/browser/chrome_content_browser_client.cc    | 4 ++--
+ components/site_isolation/site_isolation_policy.cc | 2 ++
+ content/public/common/content_features.cc          | 2 +-
+ 3 files changed, 5 insertions(+), 3 deletions(-)
+
+diff --git a/chrome/browser/chrome_content_browser_client.cc b/chrome/browser/chrome_content_browser_client.cc
+--- a/chrome/browser/chrome_content_browser_client.cc
++++ b/chrome/browser/chrome_content_browser_client.cc
+@@ -1271,7 +1271,7 @@ void ChromeContentBrowserClient::RegisterLocalStatePrefs(
+   registry->RegisterFilePathPref(prefs::kDiskCacheDir, base::FilePath());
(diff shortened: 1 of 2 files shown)


csagan5:
Can you please adjust this to cover also #1819?

uazo:
> Can you please adjust this to cover also https://github.com/bromite/bromite/issues/1819?

already disabled by default.

https://github.com/bromite/bromite/blob/e59fe6f0ab2ad0bb09659803dec0090991624b31/build/patches/Site-setting-for-javascript-jit.patch#L398

csagan5:
In this case can you please document the fact in the patch description? And other relevant information that you think is useful.

JustAnUserLol:
what's the need of enabling `kStrictOriginIsolation` since `kStrictSiteIsolation` already isolates the site and `kStrictOriginIsolation` only isolates specific origins.

uazo:
kStrictOriginIsolation is better than kStrictOriginIsolation.

```
kStrictSiteIsolation
    "Security mode that enables site isolation for all sites (SitePerProcess). "
    "In this mode, each renderer process will contain pages from at most one "
    "site, using out-of-process iframes when needed.
```
```
kStrictOriginIsolation
    "Experimental security mode that strengthens the site isolation policy. "
    "Controls whether site isolation should use origins instead of scheme and "
    "eTLD+1.
```
so the first isolates by schema and eTLD+1 while the last narrows it down to just for origin, so the first generates a rendering process for each "eTLD+1" while the second shares it for each "eTLD" [more details](https://bugs.chromium.org/p/chromium/issues/detail?id=1259920)
the activation of kStrictSiteIsolation is also mandatory for the new flag set as default to work
vanadium has it, bromite should have it too
https://github.com/GrapheneOS/Vanadium/blob/12/patches/0050-enable-strict-site-isolation-by-default-on-Android.patch
https://github.com/GrapheneOS/Vanadium/blob/12/patches/0080-Enable-strict-origin-isolation-by-default.patch

csagan5:
Thanks!

JustAnUserLol:
Strict Site Isolation is not enabled by default on `99.0.4844.55`.

uazo:
`Strict Site Isolation` is superseded by `Strict Origin Isolation`, see my comment above "kStrictOriginIsolation is better than kStrictSiteIsolation"

PR #1869, opened 2022-03-10
Add Disable-ua-full-version.patch

### LINKED ISSUE #1781: To consider removing NavigatorUAData.getHighEntropyValues()

uazo:
As agreed, the patch that modifies `getHighEntropyValues()` returning only the major version.

fixes #1781

### DIFF
--- a/build/patches/Disable-ua-full-version.patch
+++ b/build/patches/Disable-ua-full-version.patch
@@ -1,0 +1,64 @@
+From: uazo <uazo@users.noreply.github.com>
+Date: Wed, 16 Feb 2022 14:28:58 +0000
+Subject: Disable ua full version in getHighEntropyValues()
+
+getHighEntropyValues returns only the major version
+---
+ .../renderer/core/frame/navigator_ua_data.cc  | 22 +++++++++++++++++--
+ 1 file changed, 20 insertions(+), 2 deletions(-)
+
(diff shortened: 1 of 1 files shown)


csagan5:
Thanks for the PR!

Isn't this function similar/same to this one? https://source.chromium.org/chromium/chromium/src/+/main:components/embedder_support/user_agent_utils.cc;l=460;drc=51117f6b525f9d16ed5c32d2e5133335b4134ccf?q=greasedu&ss=chromium%2Fchromium%2Fsrc

uazo:
> Isn't this function similar/same to this one?

yes, and perhaps that is even more beautiful!

uazo:
I change it that way

PR #1901, opened 2022-03-18
Remove autofill and payments

csagan5:
Attempt to remove autofill assistant completely.


uazo:
have you tried if you reduce the size and build time?

uazo:
silly, I launch the compilation and then I tell you ..
https://github.com/uazo/bromite-buildtools/actions/runs/2009484479

csagan5:
It will reduce the APK size and the build time, but it will be moot but it will also increase the patchwork: it is touching too many files.

Also, my patch is incomplete, I gave up as I saw that there are too many changes.

PR #1958, opened 2022-04-08
Add lifetime options support for permissions

uazo:
I'm sorry, but my patches seem to get more and more complex along the time...
I try to explain the applied logics, ask me if I'm not clear.

this patch takes advantage of the possibility to indicate the session mode for content-settings: in fact in chromium it is possible to specify the constraint `content_settings::SessionModel` as `UserSession` when setting the value, even choosing a possible expiration time.
The logics I have seen with the ClientHints but it seems that chromium exploits this possibility when it is necessary to maintain a specific value by origin.
In any case, all content settings of this type are not saved on disk, allowing you to reset the status each time the application is restarted.

The patch acts on these points:
- components/content_settings
defines a new `content_settings::LifetimeMode` enum to specify the user's choice (Always, OnlyThisTime, UntilOriginClosed, UntilBrowserClosed).
the enumeration is also generated for java by adding in `content_settings_enums_javagen` (gn). enumeration is mainly used in `content_settings_utils.cc` to create a `specific content_settings::ContentSettingConstraints` for use in `SetContentSettingDefaultScope()` by `PermissionContextBase::UpdateContentSetting`.
in fact the chromium code does not provide a specific property to define a choice, which must instead be encoded through the ContentSettingConstraints: such mechanism is already in use in other parts of the code.
Therefore, `content_settings::GetConstraintSessionExpiration()` and `content_settings::IsConstraintSessionExpiration()` manage the lifetime modes of the session content-settings.
The patch also adds the session pattern to the `ContentSettingPatternSource`, so that it is mainly available for the ui.

- components/permissions
actually adds lifetime support to permissions.
most of the changes are caused by the fact that it is necessary to report the value selected by the user from the java ui managed by `components/browser_ui`, up to `PermissionContextBase::UpdateContentSetting()`, without necessarily having to modify all requests that are not related to the geolocation / camera / microphone.
the mechanism I used was adding a new `PermissionRequest::PermissionDecidedCallbackWithLifetime` used by an overload of `PermissionContextBase::CreatePermissionRequest` so that only the code that interests me is modified: in fact the options are present only for the specific content-settings (see PermissionDialogModel.java).
for other permissions, the patch behaves as before (see `PermissionDialogDelegate::Accept`)
for geolocation it was necessary to act directly in the specific context, because, unlike the microphone / camera, although the class always derives from PermissionContextBase, the value in the content-setting is inserted in its specific method (`FinishNotifyPermissionSet`).

- components/page_info
here the changes make it possible to see in the summary of the page_info the text "(only this session)" (aka `page_info_android_permission_session_permission`) through adding a new property "is_user_session" in `PageInfoPermissionEntry` (java)

- components/browser_ui
the changes affect the settings ui, showing "(only this session)" in the specific content-setting.
the same view is used both in the settings and in the page_info.

for the management of `UntilOriginClosed`, I recovered the logics used by the flag `kOneTimeGeolocationPermission` active only in the desktop (files last_tab_standing_tracker_ *). basically that class manages a list of the active origins and allows you to perform operations when all the tabs relating to that origin are been closed: at that moment I delete the session content settings of type `UntilOriginClosed`.

finally, turn `kUnifiedAutoplay` back on.

this is the result:
image
image

Fix for #1549

### DIFF
--- a/build/patches/Add-lifetime-options-permissions.patch
+++ b/build/patches/Add-lifetime-options-permissions.patch
@@ -1,0 +1,1415 @@
+From: uazo <uazo@users.noreply.github.com>
+Date: Fri, 8 Apr 2022 11:04:04 +0000
+Subject: Add lifetime options support for permissions
+
+Add three options to the permissions selection panel 
+(geolocation, camera and microphone):
+- Just this once
+- Until the page is closed
+- Until the browser is closed
+
+see also: https://github.com/bromite/bromite/issues/1549
+---
+ .../permissions/last_tab_standing_tracker.cc  | 33 +++++++-
+ .../permissions/last_tab_standing_tracker.h   |  4 +-
+ .../last_tab_standing_tracker_factory.cc      |  2 +-
+ ...hrome_browser_main_extra_parts_profiles.cc |  5 +-
+ chrome/browser/ui/tab_helpers.cc              |  1 +
+ .../site_settings/PermissionInfo.java         | 14 +++-
+ .../site_settings/SingleWebsiteSettings.java  | 10 +++
+ .../WebsitePreferenceBridge.java              |  6 +-
+ .../android/website_preference_bridge.cc      |  7 +-
+ .../strings/android/browser_ui_strings.grd    |  5 ++
+ components/content_settings/android/BUILD.gn  |  1 +
+ .../core/browser/content_settings_utils.cc    | 28 +++++++
+ .../core/browser/content_settings_utils.h     |  6 ++
+ .../core/browser/host_content_settings_map.cc |  2 +-
+ .../core/common/content_settings.cc           |  7 +-
+ .../core/common/content_settings.h            |  4 +-
+ .../common/content_settings_constraints.h     | 11 +++
+ .../page_info/PageInfoController.java         |  4 +-
+ .../PermissionParamsListBuilder.java          | 13 ++-
+ .../android/page_info_controller_android.cc   | 10 ++-
+ components/page_info/page_info.cc             |  2 +
+ components/page_info/page_info.h              |  1 +
+ .../permissions/PermissionDialogDelegate.java | 13 +++
+ .../permissions/PermissionDialogModel.java    | 81 ++++++++++++++++++-
+ .../android/permission_dialog_delegate.cc     | 19 +++++
+ .../android/permission_dialog_delegate.h      |  1 +
+ .../android/permission_prompt_android.cc      |  8 ++
+ .../android/permission_prompt_android.h       |  2 +
+ .../android/permissions_android_strings.grd   | 17 ++++
+ .../geolocation_permission_context_android.cc | 35 ++++++--
+ .../geolocation_permission_context_android.h  | 13 ++-
+ .../permissions/permission_context_base.cc    | 50 ++++++++++--
+ .../permissions/permission_context_base.h     | 24 +++++-
+ components/permissions/permission_prompt.h    |  3 +-
+ components/permissions/permission_request.cc  | 34 +++++++-
+ components/permissions/permission_request.h   | 17 +++-
+ .../permissions/permission_request_manager.cc | 38 ++++++---
+ .../permissions/permission_request_manager.h  | 11 ++-
+ media/base/media_switches.cc                  |  2 +-
+ 39 files changed, 481 insertions(+), 63 deletions(-)
+
+diff --git a/chrome/browser/permissions/last_tab_standing_tracker.cc b/chrome/browser/permissions/last_tab_standing_tracker.cc
+--- a/chrome/browser/permissions/last_tab_standing_tracker.cc
++++ b/chrome/browser/permissions/last_tab_standing_tracker.cc
+@@ -7,9 +7,35 @@
+ #include "base/observer_list.h"
+ #include "chrome/browser/profiles/profile.h"
+ #include "url/gurl.h"
++#include "components/content_settings/core/browser/host_content_settings_map.h"
++#include "components/content_settings/core/common/content_settings_utils.h"
++#include "components/permissions/permissions_client.h"
+ 
+-LastTabStandingTracker::LastTabStandingTracker() = default;
++namespace {
++  // Remove all sessions content setting by origin and type
++  void RemoveSessionSettings(HostContentSettingsMap* content_settings,
++                             const url::Origin& origin,
++                             ContentSettingsType type) {
++    ContentSettingsForOneType session_settings;
++    content_settings->GetSettingsForOneType(
++        type, &session_settings,
++        content_settings::SessionModel::UserSession);
+ 
++    GURL url = origin.GetURL();
++    for (ContentSettingPatternSource& entry : session_settings) {
++      if (content_settings::IsConstraintSessionExpiration(entry,
++              content_settings::LifetimeMode::UntilOriginClosed) &&
++          entry.primary_pattern.Matches(url)) {
++        content_settings->SetWebsiteSettingCustomScope(
++            entry.primary_pattern, entry.secondary_pattern,
++            type, base::Value());
++      }
++    }
++  }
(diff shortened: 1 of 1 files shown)


build/patches/Add-lifetime-options-permissions.patch
```diff
@@ -0,0 +1,1415 @@
...
+ }
+diff --git a/components/browser_ui/strings/android/browser_ui_strings.grd b/components/browser_ui/strings/android/browser_ui_strings.grd
+--- a/components/browser_ui/strings/android/browser_ui_strings.grd
++++ b/components/browser_ui/strings/android/browser_ui_strings.grd
+@@ -581,6 +581,11 @@
+       <message name="IDS_PAGE_INFO_URL_TRUNCATED" desc="Accessibility announcement when the URL in PageInfo switches from full to truncated display">
+         URL truncated
+       </message>
++      <message name="IDS_PAGE_INFO_ANDROID_PERMISSION_SESSION_PERMISSION"
++               desc="The label used in the About Page dialog to indicate an session permission">
```
csagan5, line 317:
```suggestion
+               desc="The label used in the About Page dialog to indicate a session permission">
```

build/patches/Add-lifetime-options-permissions.patch
```diff
@@ -0,0 +1,1415 @@
...
+--- a/components/content_settings/core/browser/content_settings_utils.cc
++++ b/components/content_settings/core/browser/content_settings_utils.cc
+@@ -187,4 +187,32 @@ base::Time GetConstraintExpiration(const base::TimeDelta duration) {
+   return base::Time::Now() + duration;
+ }
+ 
++ContentSettingConstraints GetConstraintSessionExpiration(LifetimeMode lifetime_mode) {
++  if (lifetime_mode == LifetimeMode::OnlyThisTime) {
++    // note: this content settings will be discarded immediately
++    return {base::Time() + base::Hours(1), content_settings::SessionModel::UserSession};
```
csagan5, line 345: If it is discarded immediately then why the +1 hour here?

build/patches/Add-lifetime-options-permissions.patch
```diff
@@ -0,0 +1,1415 @@
...
+   // Calls RequestFinished on a request and all its duplicates.
+diff --git a/media/base/media_switches.cc b/media/base/media_switches.cc
+--- a/media/base/media_switches.cc
++++ b/media/base/media_switches.cc
+@@ -469,7 +469,7 @@ const base::Feature kUseR16Texture{"use-r16-texture",
+ // Enables the Unified Autoplay policy by overriding the platform's default
+ // autoplay policy.
+ const base::Feature kUnifiedAutoplay{"UnifiedAutoplay",
+-                                     base::FEATURE_DISABLED_BY_DEFAULT};
++                                     base::FEATURE_ENABLED_BY_DEFAULT};
```
csagan5, line 1410: Rather than this change here, the patch that disables `kUnifiedAutoplay` should be removed.

build/patches/Add-lifetime-options-permissions.patch
csagan5: Please add a comment here and where it is evaluated about this `1h` being a magic value used to differentiate it from the others.

build/patches/Add-lifetime-options-permissions.patch
uazo: sure, it is that in my repo I tend not to modify the official patches anymore so as not to mess up. 
if you accept it we'll fix it.

build/patches/Add-lifetime-options-permissions.patch
uazo: in my opinion it is not needed, because the logic is only present in that file, that is, it is an opaque value for everyone

csagan5:
> In any case, all content settings of this type are not saved on disk

Except when `Forever` is selected?

> The patch acts on these points:

Can you please add a summary of this information in the patch description itself?

> * for geolocation it was necessary to act directly in the specific context, because, unlike the microphone / camera, although the class always derives from PermissionContextBase, the value in the content-setting is inserted in its specific method (`FinishNotifyPermissionSet`).

Any idea why it is like that?

> finally, turn `kUnifiedAutoplay` back on.

Can you remind me (& others) why we could not turn it on before?

> my patches seem to get more and more complex along the time

It was unavoidable for the support of this; congratulations, it seems like very good work.

build/patches/Add-lifetime-options-permissions.patch
csagan5: It is necessary because it is a magic number and it is not made as a separate constant.

There should be a comment here:
```
} else if (source.expiration == (base::Time() + base::Hours(1))) {
    // 1h is used as a magic constant to identify the one-time lifetime mode
```
And also here:
```
// note: this content settings will be discarded immediately
// the 1h value is used as a magic constant here
```

Without such comments reader can think it is valid for 1h.

build/patches/Add-lifetime-options-permissions.patch
csagan5: I would like to know if there are any remaining downsides in `kUInifiedAutoplay` being enabled (see question in main discussion).

uazo:
> Except when Forever is selected?

sure, in that case the code ends exactly as before

> Can you please add a summary of this information in the patch description itself?

ci posso provare ;)

> Any idea why it is like that?

notifies other components of the permission and waits for a response (by a callback)

> Can you remind me (& others) why we could not turn it on before?

the problem is that some contexts on the javascript side do not have an interface and therefore it is not possible for the user to activate the function via touch.
with this patch the problem is overcome because, if the user wants, he can choose a session option and therefore he will always be prompted for activation.

> It was unavoidable for the support of this; congratulations, it seems like very good work.

grazie!

csagan5:
> > Can you please add a summary of this information in the patch description itself?
> 
> ci posso provare ;)

You already have it in the PR description, you could copy/paste/adjust that :)

build/patches/Add-lifetime-options-permissions.patch
uazo: fixed in https://github.com/bromite/bromite/pull/1958/commits/334744dfe5cfc2ebcfb78243978e0af531f85f31

csagan5 (approved):

csagan5:
There is a change that does not compile:
```
../../chrome/browser/permissions/last_tab_standing_tracker.cc:84:5: error: [chromium-style] auto variable type must not deduce to a raw pointer type.
    auto content_settings =
    ^~~~
    auto*
1 error generated.
```
Will try to fix it.

csagan5:
On a related note: are these changes tested?

uazo:
> On a related note: are these changes tested?

manual tests, it would be too much to think that I develop without compiling!

>  auto variable type must not deduce to a raw pointer type.

i think we compile with different versions of clang.
I use a particular flag of [gn](https://github.com/uazo/bromite-buildtools/blob/master/images/bromite-build/build_args.gni#L17) for [kythe](https://source.chromium.org/chromium/chromium/src/+/main:tools/mb/mb_config_expectations/chromium.infra.codesearch.json;l=5?q=clang_use_chrome_plugins%20&ss=chromium%2Fchromium%2Fsrc), I don't know if that's why.
some time ago i saw that cfi requires an older version of clang, but i didn't elaborate on the matter, and in any case, in the production build I use bromite flags.
this is the [corresponding one](https://github.com/uazo/bromite-buildtools/actions/runs/2115506013), and as you can see it compiles without problems.

csagan5:
> this is the [corresponding one](https://github.com/uazo/bromite-buildtools/actions/runs/2115506013), and as you can see it compiles without problems.

It compiles without problems because your run is skipping this: https://chromium.googlesource.com/website.git/+/2c8422cfa53e5bcd666636f284b9ff2bf82273aa/site/developers/coding-style/chromium-style-checker-errors/index.md

PR #1964, opened 2022-04-11
Don't save WebsiteSettings if it's not a ContentSetting

uazo:
Modified the patch not to save WebsiteSettings other than ContentSettings in always incognito mode.

Fix for #1942

### DIFF
--- a/build/patches/Add-an-always-incognito-mode.patch
+++ b/build/patches/Add-an-always-incognito-mode.patch
@@ -27,7 +27,7 @@ Use a native flag (credits to @uazo)
  .../webapps/WebappIntentDataProvider.java     | 14 +++++
  .../chrome_autocomplete_provider_client.cc    |  7 +++
  .../chrome_autocomplete_provider_client.h     |  1 +
- .../host_content_settings_map_factory.cc      | 16 +++++-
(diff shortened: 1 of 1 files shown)


build/patches/Add-an-always-incognito-mode.patch
```diff
@@ -760,6 +764,108 @@ diff --git a/chrome/common/pref_names.h b/chrome/common/pref_names.h
...
+                          bool is_off_the_record,
++                         bool always_incognito_enabled,
+                          bool store_last_modified,
+                          bool restore_session);
+ 
+@@ -309,6 +310,7 @@ class HostContentSettingsMap : public content_settings::Observer,
+ 
+   // Whether this settings map is for an incognito or guest session.
+   bool IsOffTheRecord() const { return is_off_the_record_; }
++  bool IsAlwaysIncognito() const { return always_incognito_enabled_; }
```
csagan5, line 856: Unused?

uazo: ops... yes.
I used it to check with log(), then I forgot to take it off.

csagan5: Ok, I will remove it.

PR #1974, opened 2022-04-13
Partition dns requests by top-frame nik

uazo:
As written in [bug 1082311](https://bugs.chromium.org/p/chromium/issues/detail?id=1082311) currently chromium does not provide isolation of DOH requests.
This would easily allow doh servers to reconstruct user navigation by _device_, especially if TLS Resumption is enabled.

The patch fixes the problem by associating the doh request to the top frame nik, but in any case separating it with an additional layer because it is marked as dns request (`IsDns` in the `NetworkIsolationKey`).
The patch adds two new flags: `kSplitDnsByNetworkIsolationKey`, active by default, and `kLogSplitDnsByNetworkIsolationKey` which allows check functionality (via a "DNS:" filter in logcat).

actually the patch multiplies the https ports open to the doh server, which you can see with applications like PCAPdroid, but the management doesn't seem to cause any problems, especially if you are using doh servers that have a rather low server side timeout (like opendns).

### DIFF
--- a/build/patches/Partition-dns-requests-by-topframe-nik.patch
+++ b/build/patches/Partition-dns-requests-by-topframe-nik.patch
@@ -1,0 +1,472 @@
+From: uazo <uazo@users.noreply.github.com>
+Date: Wed, 23 Mar 2022 08:37:47 +0000
+Subject: Partition dns requests by top-frame nik
+
+---
+ chrome/browser/about_flags.cc        | 12 ++++++++++++
+ chrome/browser/flag_descriptions.cc  |  9 +++++++++
+ chrome/browser/flag_descriptions.h   |  6 ++++++
+ net/base/features.cc                 |  6 ++++++
+ net/base/features.h                  |  6 ++++++
+ net/base/isolation_info.cc           | 24 +++++++++++++++++++++++-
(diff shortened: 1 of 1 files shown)


csagan5 (approved):

thestinger:
We're not planning on using Chromium's DoH implementation much ourselves since there's the OS DoT implementation and usually people want to use it for every app or no apps.

Maybe you can get this landed upstream? It would be ideal if we could do that with some of our patches. Ideally the configuration default changes could also be changed into some kind of build configuration. It would be so much easier than way.

uazo:
> Maybe you can get this landed upstream? It would be ideal if we could do that with some of our patches. 

For me it would be the umpteenth challenge, but unfortunately I have often seen that they do not accept patches without also test cases, which I personally am not currently capable of doing.
so for now I can't do it, even because I have the problem of my bad english :(

obviously @csagan5 feel free to do what you see fit

csagan5:
@uazo I am thinking to change the wording a bit to mention "DoH" rather than "DNS" in some instances because the improvement does not really apply to non-DoH traffic.

As for upstreaming: yes, that would indeed be the best outcome.

PR #1975, opened 2022-04-13
Move top toolbar to bottom

uazo:
The patch allows you to move the navigation bar down, via a new flag in the accessibility ui. unlike what is present in other browsers, this actually moves the bar below, without creating other elements.
Although it is large patch, it is actually quite simple because it acts on the offsets of the compositor, which builds the scene of the interface, regardless of the content.
The patch also colors the current bottom bar as the topbar to make the interface pleasant and fixes the bug [1285311](https://bugs.chromium.org/p/chromium/issues/detail?id=1285311) that I opened.
The patch is also active in tablet mode.

feel free to ask me all the questions you want.

### DIFF
--- a/build/patches/Move-toptoolbar-to-bottom.patch
+++ b/build/patches/Move-toptoolbar-to-bottom.patch
@@ -1,0 +1,2172 @@
+From: uazo <uazo@users.noreply.github.com>
+Date: Tue, 18 Jan 2022 07:43:32 +0000
+Subject: Move top toolbar to bottom
+
+Adds an accessibility flag that allows the top toolbar to be moved to the bottom.
+The patch also includes tablet mode.
+---
+ cc/base/features.cc                           |  3 ++
+ cc/base/features.h                            |  1 +
+ cc/input/browser_controls_offset_manager.cc   |  6 +++
+ cc/trees/layer_tree_host_impl.cc              |  3 ++
+ .../tab_management/TabGroupUiCoordinator.java |  7 ++-
+ .../tab_management/TabGroupUiMediator.java    | 50 +++++++++++++++++-
+ .../tab_management/TabGroupUiProperties.java  |  6 ++-
+ .../tab_management/TabGroupUiToolbarView.java | 18 +++++++
+ .../tab_management/TabGroupUiViewBinder.java  |  3 ++
+ .../tab_management/TabSwitcherMediator.java   | 15 ++++++
+ .../ChromeAccessibilitySettingsDelegate.java  | 52 +++++++++++++++++++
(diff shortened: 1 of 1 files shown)


csagan5:
Thanks, I went quickly through it, seems fine. Will give it a try later.

build/patches/Move-toptoolbar-to-bottom.patch
```diff
@@ -0,0 +1,2172 @@
...
+     }
+ 
++    /**
++     * Allows the modification of the flag value on the java side.
++     * Currently only the feature flag with on / off values is managed.
++     *
++     * @param featureName the feature name from ChromeFeatureList.
++     * @param flagName the flag name name from about_flags.cc.
++     */
++    public static void setFlagEnabled(String featureName, String flagName, Boolean newValue) {
```
csagan5, line 1460: Isn't this like `setForTesting`?

build/patches/Move-toptoolbar-to-bottom.patch
```diff
@@ -0,0 +1,2172 @@
...
+         String getAdBlockFiltersURL();
+diff --git a/chrome/browser/flags/android/java/src/org/chromium/chrome/browser/flags/ChromeFeatureList.java b/chrome/browser/flags/android/java/src/org/chromium/chrome/browser/flags/ChromeFeatureList.java
+--- a/chrome/browser/flags/android/java/src/org/chromium/chrome/browser/flags/ChromeFeatureList.java
++++ b/chrome/browser/flags/android/java/src/org/chromium/chrome/browser/flags/ChromeFeatureList.java
+@@ -403,6 +403,8 @@ public abstract class ChromeFeatureList {
+     public static final String MOBILE_IDENTITY_CONSISTENCY_PROMOS =
+             "MobileIdentityConsistencyPromos";
+     public static final String MODAL_PERMISSION_DIALOG_VIEW = "ModalPermissionDialogView";
++    public static final String MOVE_TOP_TOOLBAR_TO_BOTTOM =
++            "MoveTopToolbarToBottom";
```
csagan5, line 1489: It could fit in one line?

csagan5 (approved):

csagan5:
I will add it to the roster and possibly release it with next v100; I cannot see anything dangerous but I also think there might be more bugs related to the offsets.

Nice work :+1:

csagan5:
@uazo there is a bug: a white area appears just below the top navigation bar on the NTP, and below them there are the Bookmarks and Downloads small text and icons and the NTP icons.

It looks like this patch is interfering with https://github.com/bromite/bromite/blob/master/build/patches/Restore-Simplified-NTP-launch.patch

uazo:
> there is a bug

ok I check.

Issue #2011, opened 2022-04-22
Bottom Toolbar & 3-dot Menu


js290:
### Preliminary checklist

- [X] I have read the [README](https://github.com/bromite/bromite/blob/master/README.md).
- [X] I have searched the existing issues for my problem. This is a new ticket, NOT a duplicate or related to another open issue.
- [X] I have read the [FAQs](https://github.com/bromite/bromite/blob/master/FAQ.md).
- [X] I have updated Bromite to the latest version. The bug is reproducible on this latest version.

### Can the bug be reproduced with corresponding Chromium version?

No

### Bromite version

100.0.4896.135

### Device architecture

arm

### Android version

7.0

### Device model

Moto X Pure Edition

### Changed flags

move-top-toolbar-to-bottom

### Is this bug about the SystemWebView?

No

### Is this bug happening in an incognito tab?

Yes

### Is this bug caused by the adblocker?

No

### Is this bug a crash?

No

### Describe the bug

When toolbar is moved to the bottom, the 3-dot (ellipse?) menu does not display all menu options.  It shows the basic navigation icons, but doesn't show other options, most importantly Settings.  

Also, setting the flag to Default and relaunching does not move the toolbar back to the top; had to set to Disable (see also #1996).

### Steps to reproduce the bug

1. Settings -> Accessibility -> Move toolbar to bottom
2. Relaunch
3. Select 3-dot menu at bottom toolbar, observe missing menu options

### Expected behavior

The 3-dot menu should fully display regardless of where the toolbar is.

### Screenshots
![Screenshot_20220423-090138](https://user-images.githubusercontent.com/4715607/164909272-96306929-a869-4117-befb-c657f0e7e078.png)

kokukyocho:
Hello @js290!

I am a bot :robot:

Thanks for submitting this issue!
I noticed that it is missing the template, please edit the issue to match the template for either a [bug](https://github.com/bromite/bromite/blob/master/.github/ISSUE_TEMPLATE/bug_report.yml) or a [feature request](https://github.com/bromite/bromite/blob/master/.github/ISSUE_TEMPLATE/feature_request.yml); the issue will be automatically re-opened afterwards.

If this is not an issue nor a feature request and you want to start a discussion then you can use [GitHub Discussions](https://github.com/bromite/bromite/discussions).

**Tip**: if you use GitHub in Desktop mode the template will be proposed to you when submitting issues.

uazo:
could you please post some screenshots?
[...]

PR #2029, opened 2022-04-29
Fix Move-navigation-bar-to-bottom.patch

### LINKED ISSUE #1995: Unwanted grey line when toolbar is at bottom
### Preliminary checklist

No

### Bromite version

100.0.4896.135

### Device architecture

arm64-v8a

### Android version

10

### Device model

Huawei Nova CDY-NX9B

### Changed flags

Screenshot_20220419_123637

No

No

No

### Is this bug a crash?

No

### Describe the bug

A thin grey line appears where the bottom of the toolbar used to be before "move top toolbar to bottom" flag is set. 

Line is present in tab grid and in websites when scrolling up, at the same time when toolbar appears after being hidden.

Tab Grid:
1. Go to tab grid.
2. Line appears.

Website:
1. Go to https://www.iana.org/domains/reserved.
2. Scroll down until toolbar is hidden.
3. Scroll up untill toolbar appears.
4. Line appears.

### Expected behavior

Line should not be present when toolbar is at the bottom.

### Screenshots

Screenshot_20220419_121340_org bromite bromite_1

uazo:
the bug of #2011 is present in all versions because the offset of the menu is not correctly calculated and is off-screen, but only in A7 is it taken into account.
in the other android versions it automatically adjusts the offsets to make the menu be always visible.
in addition, and only in A7, the menu is not drawn correctly if the bar is docked at the bottom, for this reason I reduce the size so that it is pleasant but still functional.

Fixes #1995 #2011

### DIFF
--- a/build/patches/Move-navigation-bar-to-bottom.patch
+++ b/build/patches/Move-navigation-bar-to-bottom.patch
@@ -1328,7 +1328,7 @@ diff --git a/chrome/android/java/src/org/chromium/chrome/browser/ui/BottomContai
 diff --git a/chrome/browser/about_flags.cc b/chrome/browser/about_flags.cc
 --- a/chrome/browser/about_flags.cc
 +++ b/chrome/browser/about_flags.cc
-@@ -6590,6 +6590,11 @@ const FeatureEntry kFeatureEntries[] = {
+@@ -6587,6 +6587,11 @@ const FeatureEntry kFeatureEntries[] = {
       flag_descriptions::kImpulseScrollAnimationsDescription, kOsAll,
       FEATURE_VALUE_TYPE(features::kImpulseScrollAnimations)},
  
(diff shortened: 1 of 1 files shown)


csagan5:
I assume no side effects for non-A7 owners?

uazo:
> I assume no side effects for non-A7 owners?

no effect, it is under [version checking](https://github.com/bromite/bromite/pull/2029/files#diff-349cd662c22e8f16ced3c50d2b682062d5a3f6ccac2f0bd4d088df2cbafb0f4bR1596).

csagan5:
> > I assume no side effects for non-A7 owners?
> 
> no effect, it is under [version checking](https://github.com/bromite/bromite/pull/2029/files#diff-349cd662c22e8f16ced3c50d2b682062d5a3f6ccac2f0bd4d088df2cbafb0f4bR1596).

In this case I will change the comment to mention Android Nougat.

PR #2037, opened 2022-05-02
Multiple fixes on Always Incognito Mode

### LINKED ISSUE #1991: Reset permissions
### Preliminary checklist

Yes

Not in browsers as I have seen but Android does have this on its app permissions 

Reset site settings permission after sometime to prevent js allowing fingerprinting or show ads if it was turned off for a while

After javascript or Ads are turned off for one time it needs to be reverted. Some may find it annoying to turn it off again and again so its better to have a option to set a time like 3days or a week . So  if those websites ,where its turned off, isnt opened or accessed within that timeframe will revert permissions to what they did set on settings->site settings

uazo:
1. added specific menu accessible from privacy settings with options:
- Enable always incognito (default false)
- Enable history support for incognito mode (default false)
- Allow the saving of the site settings for incognito mode (default false)

I've also added a link for online help to https://github.com/bromite/bromite/wiki/AlwaysIncognito

2. merged [patch](https://github.com/uazo/bromite/blob/cfdc726cee3d7539d8e59b0bfd30f539f3a3db0f/build/patches/contrib/Add-history-support-in-incognito-mode-addon.patch) which disable the support for automatic saving of pages for offline mode in always incognito mode, currently active by mistake with the saving of the history

3. added new `kOfflinePagesAutoSaveFeature` flag (`offline-pages-auto-save` from about_flags) which allows to disable the option also for normal mode, useful for speeding up the return from Android Recents with heavy pages (or with heavy userscripts, see https://github.com/bromite/bromite/issues/792#issuecomment-1062648912 @xarantolus)

4. removed test code, currently useless

5. fix incorrect prefs registration position (caused by rebase v101)

6. Removed useless `IsAlwaysIncognito()` in `HostContentSettingsMap`

Fix for #1991

### DIFF
--- a/build/patches/Add-an-always-incognito-mode.patch
+++ b/build/patches/Add-an-always-incognito-mode.patch
@@ -13,81 +13,88 @@ flag turned on.
 IncognitoPlaceholder is also deactivated, both in the phone and tablet version.
 The relative tests are also present.
 
+based on the original work by Ryan Archer <ryan.bradley.archer@gmail.com>
 See also: https://github.com/bromite/bromite/pull/1427
 ---
- chrome/android/chrome_java_sources.gni        |   1 +
- .../chrome_junit_test_java_sources.gni        |   4 +
- chrome/android/chrome_test_java_sources.gni   |   6 +
- .../java/res/xml/privacy_preferences.xml      |  10 +
- .../AlwaysIncognitoLinkInterceptor.java       |  53 +++
+ chrome/android/chrome_java_resources.gni      |   1 +
+ chrome/android/chrome_java_sources.gni        |   2 +
+ .../java/res/xml/incognito_preferences.xml    |  37 ++++
+ .../java/res/xml/privacy_preferences.xml      |   5 +
+ .../AlwaysIncognitoLinkInterceptor.java       |  53 ++++++
  .../chrome/browser/ChromeTabbedActivity.java  |   6 +-
  .../chrome/browser/app/ChromeActivity.java    |   4 +
- .../AppMenuPropertiesDelegateImpl.java        |  33 +-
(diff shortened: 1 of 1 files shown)


csagan5:
> 2\. merged [patch](https://github.com/uazo/bromite/blob/master-v101/build/patches/contrib/Add-history-support-in-incognito-mode-addon.patch)

Link is broken.

> * removed test code, currently useless

Sad to see them go, I think the idea was right.

> 3\. added new `kOfflinePagesAutoSaveFeature` flag (`offline-pages-auto-save` from about_flags) which allows to disable the option also for normal mode

I am thinking that perhaps this should be off by default, the user should ask to save, not automatically.

Can you also move "Open external links in incognito" in this new page? I will merge the patch for now so please make sure to get the updated version before that.

uazo:
> > 2. merged [patch](https://github.com/uazo/bromite/blob/master-v101/build/patches/contrib/Add-history-support-in-incognito-mode-addon.patch)
> 
> Link is broken.

updated

> > * removed test code, currently useless
> 
> Sad to see them go, I think the idea was right.

yes, but i think it's useless to keep it aligned in rebases if i don't use it.
i like the method used by @arthuredelstein more, which i will try sooner or later, maybe upgrading webdriver if necessary

> > 3. added new `kOfflinePagesAutoSaveFeature` flag (`offline-pages-auto-save` from about_flags) which allows to disable the option also for normal mode
> 
> I am thinking that perhaps this should be off by default, the user should ask to save, not automatically.

from my point of view there is no problem, I always use the always incognito mode and I have never found that need.

> Can you also move "Open external links in incognito" in this new page? I will merge the patch for now so please make sure to get the updated version before that.

Sure.

PR #2039, opened 2022-05-02
Add a flag to site engagement feature (default disabled)

### LINKED ISSUE #2022: Flag/setting to disable site engagement (affects which NTP tiles are displayed)
### Preliminary checklist

Yes

It used to be possible to hide all NTP tiles, although the site engagement was still calculated: https://github.com/bromite/bromite/discussions/1979#discussioncomment-2562394

Disable site engagement so that only bookmarked sites will appear as NTP tiles.

Completely hide the NTP by using a home page as NTP (feature reintroduced in v96, #1586)

uazo:
Fix for #2022

### DIFF
--- a/build/patches/Add-flag-to-disable-site-engagement.patch
+++ b/build/patches/Add-flag-to-disable-site-engagement.patch
@@ -1,0 +1,182 @@
+From: uazo <uazo@users.noreply.github.com>
+Date: Mon, 2 May 2022 11:48:03 +0000
+Subject: Add a flag to site engagement
+
+---
+ chrome/browser/about_flags.cc                 |  6 ++++
+ chrome/browser/flag_descriptions.cc           |  5 +++
+ chrome/browser/flag_descriptions.h            |  3 ++
+ .../content/site_engagement_score.cc          |  5 +++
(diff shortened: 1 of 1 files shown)


csagan5:
This is not really disabling it, it is resetting it after each commit.

uazo:
> This is not really disabling it, it is resetting it after each commit.

what do you mean then by disable?
currently the score is never saved, so it is always empty and all the related code is skipped, so it's basically disabled

csagan5:
> currently the score is never saved

Currently the score is saved, this PR wants to reset it on every commit.

Is this `score_dict_.reset();` necessary? That is what I mean. There is no other way for the user to reset it (other than reinstalling)?

uazo:
> Is this score_dict_.reset(); necessary?

yes, to respect the `std::move(score_dict_)` otherwise skipped

> There is no other way for the user to reset it (other than reinstalling)?

do you intend the current score? by deleting the site settings from "Clear browsing data"

uazo:
> this PR wants to reset it on every commit.

I did not understand then.
This PR disables the feature, but does not delete the current scores.
deleting the current score would be like deleting the data in startup, and at that point why not delete the cookies, as well as the point (in the code) is the same?

PR #2045, opened 2022-05-03
Add webgl site setting

uazo:
this pull contains two different patches:

- Content-settings-infrastructure.patch
it is a "small" utility that allows me to have less effort in adding a new site setting in chromium, considering that every time there are always the same points to touch.
the advantage of using it may be even more evident if we changed all the other patches to use this one.
explanations are in the included readme

- Add-webgl-site-setting.patch
trivially it adds a site setting related to webgl default disabled. in the patch there is actually only the implementation, since the ui is handled by the patch from before.

you can see the various changes with https://uazo.github.io/superpatch/bromite/pull/2045

### DIFF
--- a/build/patches/Add-webgl-site-setting.patch
+++ b/build/patches/Add-webgl-site-setting.patch
@@ -1,0 +1,492 @@
+From: uazo <uazo@users.noreply.github.com>
+Date: Tue, 3 May 2022 14:44:11 +0000
+Subject: Add webgl site settings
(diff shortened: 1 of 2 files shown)


csagan5:
Nice work! I will remove the flag to disable WebGL since there is a site setting now.

> * the advantage of using it may be even more evident if we changed all the other patches to use this one.

Please go ahead.

Issue #2051, opened 2022-05-05
Literally none of the Ads are being blocked


pixincreate:
### Preliminary checklist

- [X] I have read the [README](https://github.com/bromite/bromite/blob/master/README.md).
- [X] I have searched the existing issues for my problem. This is a new ticket, NOT a duplicate or related to another open issue.
- [X] I have read the [FAQs](https://github.com/bromite/bromite/blob/master/FAQ.md).
- [X] I have updated Bromite to the latest version. The bug is reproducible on this latest version.

### Can the bug be reproduced with corresponding Chromium version?

No

### Bromite version

101.0.4951.53

### Device architecture

arm64

### Android version

12

### Device model

Nokia 6.1 Plus (LineageOS 19.1 Official)

### Changed flags

No-Flags-Changed

### Is this bug about the SystemWebView?

No

### Is this bug happening in an incognito tab?

Yes

### Is this bug caused by the adblocker?

Yes

### Is this bug a crash?

No

### Describe the bug

None of the Ads are being blocked on the latest version of Bromite. The same is even with Incognito. I got a page full of Ads on a single page itself. Screenshots attached below for reference. 

### Steps to reproduce the bug

1. Turn on the phone
2. Select Bromite
3. Type a website that supports Ads. As an example, type car on Google.com
4. You'll see a ton of Ads. 

### Expected behavior

Ads are expected to be blocked on Bromite

### Screenshots

![Screenshot_20220505-063836_Bromite](https://user-images.githubusercontent.com/69745008/166851017-d86d55d6-8767-4927-9fa4-206d06fb327d.png)
![Screenshot_20220505-063845_Bromite](https://user-images.githubusercontent.com/69745008/166851021-3e210ca0-471b-492c-bff6-21be6f3d6b77.png)
![Screenshot_20220505-063957_Bromite](https://user-images.githubusercontent.com/69745008/166851022-1d7f5006-c5f4-4f3f-92cf-286256962833.png)

Some more?

csagan5:
Please report logs with `adb logcat | grep AdBlock`

pixincreate:
I think I need Linux to be installed on my PC or atleast WSL.

pixincreate:
How do I get the logs from a windows PC?

I tried the same on Windows, it's unsupported. Did the same on git bash. It's doing nothing.
```
adb logcat | grep AdBlock > sdcard/log.txt
```
This should get save the logs into the los.txt file, but it's just an empty file.

[...]

PR #2053, opened 2022-05-06
Add Android 12L (12.1) option to version list and update AdBlock section with URL

pixincreate:
See #2051 for more reference.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -142,7 +142,8 @@ body:
     id: adblock
     attributes:
       label: Is this bug caused by the adblocker?
-      description: You can test this by temporarily allowing ads for the site.
+      description: |
+        You can test this by temporarily allowing ads for the site.
         1. No.
         2. Yes, I have attached the AdBlock  URL and Logs generated from `chrome://crashes`.
       placeholder: |
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -73,6 +73,7 @@ body:
       multiple: false
       description: What version of Android are you running?
       options:
+        - 12.1
         - 12
         - 11
         - 10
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -73,6 +73,7 @@ body:
       multiple: false
       description: What version of Android are you running?
       options:
+        - 12.1
         - 12
```
csagan5, line 77: This would need to become `12.0`; I have already changed it.

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -136,19 +137,21 @@ body:
...
       label: Is this bug caused by the adblocker?
-      description: You can test this by temporarily allowing ads for the site.
-      multiple: false
-      options:
-        - "Yes"
-        - "No"
+      description: |
+        You can test this by temporarily allowing ads for the site.
+        1. No.
+        2. Yes, I have attached the AdBlock  URL and Logs generated from `chrome://crashes`.
```
csagan5, line 148: Your issue was related to the adblocker but not caused by it; this question does not need these changes as it is only about bugs caused by the adblocker.

csagan5:
Thanks for your PR; I have added the new Android version but will not change the question about adblocker bugs.

.github/ISSUE_TEMPLATE/bug_report.yml
pixincreate: @uazo had recommended to ask about the AdBlock filters URL for `.dat` file when the issue is with AdBlock, that's the reason I changed that.

uazo: > had recommended to ask about the AdBlock filter URL

I would propose a separate mandatory field only if the issue is related to the adblock

pixincreate: Sure then, I'll mark this as resolved.
Thank you.

PR #2054, opened 2022-05-06
Add webRTC site settings

### LINKED ISSUE #1965: webRTC user-facing configuration
### Preliminary checklist

No

No, but @uazo mentioned availability [here](https://github.com/ungoogled-software/ungoogled-chromium-android/discussions/116#discussioncomment-2394669)

User-facing global settings for all the internally-supported webRTC options (see https://github.com/bromite/bromite/wiki/WebRTC#configuration-internals); this is already available in the desktop version of Chromium.

The default in Bromite cannot be changed because it would leak information by default.

uazo:
I'm sorry, normally I do the bugs first and then the new stuff, but this was really just a simple copy / paste...
I just added `kDefault` because trying the others anyway webRTC doesn't work!

Fix for #1965

### DIFF
--- a/build/patches/Add-webrtc-site-setting.patch
+++ b/build/patches/Add-webrtc-site-setting.patch
@@ -2,7 +2,7 @@ From: uazo <uazo@users.noreply.github.com>
 Date: Fri, 6 May 2022 14:27:17 +0000
 Subject: Add webRTC site settings
 
-Require: Content-settings-infrastructure.patch
+Requires patch: Content-settings-infrastructure.patch
 ---
  .../browser_ui/site_settings/android/BUILD.gn |  3 +
  .../BromiteCustomContentSettingImpl.java      |  1 +
--- a/build/patches/Add-webrtc-site-setting.patch
+++ b/build/patches/Add-webrtc-site-setting.patch
@@ -1,0 +1,381 @@
+From: uazo <uazo@users.noreply.github.com>
+Date: Fri, 6 May 2022 14:27:17 +0000
+Subject: Add webRTC site settings
+
+Require: Content-settings-infrastructure.patch
+---
+ .../browser_ui/site_settings/android/BUILD.gn |  3 +
+ .../BromiteCustomContentSettingImpl.java      |  1 +
+ .../BromiteWebRTCContentSetting.java          | 85 +++++++++++++++++++
+ .../site_settings/SiteSettingsCategory.java   |  5 +-
+ .../strings/android/browser_ui_strings.grd    |  1 +
+ .../browser_ui/strings/android/webrtc.grdp    | 18 ++++
+ components/components_strings.grd             |  1 +
+ .../core/browser/content_settings_registry.cc | 14 +++
+ .../core/browser/content_settings_utils.cc    |  2 +
+ .../core/common/content_settings.cc           |  3 +-
+ .../core/common/content_settings.h            |  1 +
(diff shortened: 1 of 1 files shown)


build/patches/Add-webrtc-site-setting.patch
```diff
@@ -0,0 +1,381 @@
...
++
++    @Override
++    public int getCategorySummary(@Nullable @ContentSettingValues int value) {
++        switch (value) {
++            case ContentSettingValues.ALLOW:
++                return R.string.website_settings_category_webrtc_enabled;
++            case ContentSettingValues.BLOCK:
++                return R.string.website_settings_category_webrtc_disabled;
++            default:
++                return 0;
```
csagan5, line 120: When is this used?

csagan5:
Slightly related: https://github.com/GrapheneOS/Vanadium/issues/181

(Something similar could be added to the other WebRTC patch)

build/patches/Add-webrtc-site-setting.patch
uazo: here
https://github.com/bromite/bromite/blob/master/build/patches/Content-settings-infrastructure.patch#L641

image

it is the value that appears there, when opening the settings panel with the summary of all contentsettings, and it can be a different string from the other used by the interface as an option for the setting.

uazo:
> Something similar could be added to the other WebRTC patch

i.e.?
`ungoogled-chromium-Disable-webRTC-log-uploader.patch` and `webRTC-do-not-expose-local-IP-addresses.patch` it is correct that it is always active

build/patches/Add-webrtc-site-setting.patch
csagan5: Please add a comment since this `0` is a magic constant.

csagan5:
I am referring to `webRTC-do-not-expose-local-IP-addresses.patch`.

> it is correct that it is always active

What is always active?

build/patches/Add-webrtc-site-setting.patch
uazo: > Please add a comment since this 0 is a magic constant.

ok, the comment is "with zero a java exception is thrown". I add it

uazo:
> What is always active?

I meant by activate the patch, to apply it on the chromium code.
`webRTC-do-not-expose-local-IP-addresses.patch` patch and `ungoogled-chromium-Disable-webRTC-log-uploader.patch` are anyway needed (as they are) even after introducing this one (`Add webRTC site settings`, of this pull)

csagan5:
Ok. When I wrote "Something similar could be added to the other WebRTC patch" I was referring to the linked Vanadium issue and to the device identifier that webRTC uses.

build/patches/Add-webrtc-site-setting.patch
```diff
@@ -0,0 +1,381 @@
+From: uazo <uazo@users.noreply.github.com>
+Date: Fri, 6 May 2022 14:27:17 +0000
+Subject: Add webRTC site settings
+
+Require: Content-settings-infrastructure.patch
```
csagan5, line 5:
```suggestion
Requires patch: Content-settings-infrastructure.patch
```

build/patches/Add-webrtc-site-setting.patch
csagan5: I will add the comment.

PR #2075, opened 2022-05-23
Fix incognito screenshots flag name

ArchangeGabriel:
It was apparently renamed at some point but not reflected here.


csagan5 (approved):

csagan5:
Thanks!

PR #2092, opened 2022-05-31
Remove duplicate offline-indicator-v2 flag

Zelda189:
Removed duplicated `offline-indicator-v2` flag.


csagan5:
This PR is not removing anything?

Zelda189:
In Bromite version `102.0.5005.67`, offline-indicator-v2 accidentally removed and needs to fix next version.

csagan5:
I see what you mean, thanks!

PR #2120, opened 2022-06-07
Fix call to StartExpiringOldStuff

### LINKED ISSUE #2104: History threshold don't work
### Preliminary checklist

No

### Bromite version

102.0.5005.67

### Device architecture

arm

### Android version

10

### Device model

Galaxy J4

### Changed flags

move-top-toolbar-to-bottom

No

No

No

### Is this bug a crash?

No.

### Describe the bug

History don't get deleted after the specified interval on the "Expire history days threshold" setting, it just works if i change to "No history".

1. Go to "Settings"
2. Enter in "Privacy and Security"
3. Change the "Expire history days threshold" interval

### Expected behavior

The navigation history should be deleted after the specified interval, but it doesn't.

### Screenshots

_No response_

uazo:
I fixed the bug, caused by an incorrect order of the calls between assignment of the range and actual removal.
in fact it was possible that the call to [StartExpiringOldStuff()](https://source.chromium.org/chromium/chromium/src/+/main:components/history/core/browser/history_backend.cc;l=1011;drc=1946212ac0100668f14eb9e2843bdd846e510a1e;bpv=1;bpt=1) overwrites the value set by the retrieval of the threshold in the preferences, which could happen later.

fixes #2104

### DIFF
--- a/build/patches/History-number-of-days-privacy-setting.patch
+++ b/build/patches/History-number-of-days-privacy-setting.patch
@@ -17,15 +17,15 @@ flag and immediately deletes all the history.
  .../java/res/layout/preference_spinner.xml    |  7 +++
  .../layout/preference_spinner_single_line.xml |  7 +++
  .../settings/SpinnerPreference.java           |  7 +++
- .../core/browser/expire_history_backend.cc    | 21 +++++++
+ .../core/browser/expire_history_backend.cc    | 13 +++-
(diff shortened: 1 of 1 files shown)


csagan5:
Is there any race condition left?

uazo:
> Is there any race condition left?

no, I don't think so.

PR #2166, opened 2022-07-04
Patch for allowing to match only system dark mode

krlvm:
Unlike other apps, when Chromium-based Android browsers are configured to follow system dark mode preference, they ignore the actual system preference and respect battery saver mode first of all, while most users always have power saving mode enabled, making this setting useless.

This patch forces Chromium to ignore power saving mode when choosing a theme, making its behavior consistent with the rest of the apps, including the system and Google apps that follow actual system theme on Android 10 and above in the first place.


csagan5:
Are there any other side effects? I assume not.

krlvm:
I don't think so, I was trying to make this patch as less invasive as possible

csagan5:
Thanks, I will merge it.

krlvm:
Thank you! For a long time I hoped that the Chromium developers would finally fix this shortcoming. But unfortunately :(

PR #2171, opened 2022-07-06
Add kill switch for no-unqualified-std-cast-call build option

uazo:
I realize you don't need it, but compiling with clangd won't work unless I turn off the `no-unqualified-std-cast-call` option.
I need clangd to create the index from which I get all the flags for each version, and in the future I would like to activate the autocomplete in visual studio code.
it doesn't matter to you, since the default is false, but I can understand if you don't accept it


csagan5:
clangd does not accept `-Wno-unqualified-std-cast-call`, no plans to add support? this warning means that there are missing `std::` in the codebase.

csagan5:
> I didn't find anything, but it doesn't matter

It matters to me because this is a workaround for an external tool (`clangd`) so one would ideally check if the external tool has plans to support the feature.

uazo:
> It matters to me because this is a workaround for an external tool

ok.
https://releases.llvm.org/13.0.0/tools/clang/docs/DiagnosticsReference.html here you find the list of flags, but there isn't.
some reference https://reviews.llvm.org/rG70b1f6de539867353940d3dcb8b25786d5082d63

I say that it does not matter because, since the code is already compiled with chromium clang, there cannot be those warnings.

csagan5:
> I say that it does not matter because, since the code is already compiled with chromium clang, there cannot be those warnings.

I do not understand, does Chromium use a different version of Clang which has the `no-unqualified-std-cast-call`?

We could merge this, although not relevant for the project; you would have to add a comment there explaining what the new GN flag is about and maybe rename it? Perhaps `skip_clangd_unsupported_options`?

uazo:
> does Chromium use a different version of Clang 

Yes, it is

> Perhaps skip_clangd_unsupported_options?

Ok.

csagan5:
> Yes, it is

Ok, then please mention it in the patch description because once Chromium upgrades we will have to remove this.

PR #2172, opened 2022-07-06
Remove navigator.connection info

uazo:
since it is a fingerprinting surface, it makes sense to remove the result of `navigator.connection`

### DIFF
--- a/build/patches/Remove-navigator-connection-info.patch
+++ b/build/patches/Remove-navigator-connection-info.patch
@@ -1,21 +1,97 @@
 From: uazo <uazo@users.noreply.github.com>
-Date: Wed, 6 Jul 2022 10:18:16 +0000
+Date: Thu, 14 Jul 2022 09:48:45 +0000
 Subject: Remove navigator.connection info
 
+Change the result of navigator.connection to default values 
+and disable observers
 ---
- .../blink/renderer/modules/netinfo/network_information.cc        | 1 +
- 1 file changed, 1 insertion(+)
+ .../renderer/modules/netinfo/network_information.cc   |  7 ++++++-
+ .../renderer/modules/netinfo/network_information.h    |  1 +
+ .../platform/network/network_state_notifier.cc        | 11 +++++++++++
+ .../platform/network/network_state_notifier.h         |  2 +-
+ 4 files changed, 19 insertions(+), 2 deletions(-)
 
 diff --git a/third_party/blink/renderer/modules/netinfo/network_information.cc b/third_party/blink/renderer/modules/netinfo/network_information.cc
 --- a/third_party/blink/renderer/modules/netinfo/network_information.cc
 +++ b/third_party/blink/renderer/modules/netinfo/network_information.cc
-@@ -261,6 +261,7 @@ void NetworkInformation::StopObserving() {
- const char NetworkInformation::kSupplementName[] = "NetworkInformation";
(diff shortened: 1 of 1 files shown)


csagan5:
It returns always `4g` and 10Mbps; after an event listener is added it should start using real information. Rather than breaking the Javascript API we should make it never start observing.

uazo:
> Rather than breaking the Javascript API we should make it never start observing.

I thought about it too, but I thought it was a waste of time.
`navigator.connection` is an [optional feature](https://caniuse.com/netinfo) (in [draft](https://github.com/WICG/netinfo/)) and appears to be checked for null by [many scripts](https://sourcegraph.com/search?q=context:global+navigator.connection+AND+lang:javascript+AND+-repo:chromium&patternType=regexp)
do we do it anyway?

csagan5:
In this case I think it's better to not differ from Chromium, it's not worth it.

csagan5:
> I propose an alternative: wire the result with constants, so maintenance becomes easier.

Sure, but the logic "when there is an observer then return real data" must also be removed.

uazo:
alright, it was just to simplify the patch, but if you deem it important (I trust your choices), I remove it

build/patches/Remove-navigator-connection-info.patch
```diff
@@ -0,0 +1,97 @@
...
+
+diff --git a/third_party/blink/renderer/modules/netinfo/network_information.cc b/third_party/blink/renderer/modules/netinfo/network_information.cc
+--- a/third_party/blink/renderer/modules/netinfo/network_information.cc
++++ b/third_party/blink/renderer/modules/netinfo/network_information.cc
+@@ -61,7 +61,7 @@ NetworkInformation::~NetworkInformation() {
+ }
+ 
+ bool NetworkInformation::IsObserving() const {
+-  return !!connection_observer_handle_;
++  return !!connection_observer_handle_ || is_fake_observing_;
```
csagan5, line 22: Actually we just need to return always false here.

build/patches/Remove-navigator-connection-info.patch
```diff
@@ -0,0 +1,97 @@
...
+@@ -522,6 +522,17 @@ NetworkStateNotifier::GetWebHoldbackDownlinkThroughputMbps() const {
+   return absl::nullopt;
+ }
+ 
++NetworkStateNotifier::NetworkStateNotifier() : has_override_(false) {
++  // set default data
++  // see third_party/blink/renderer/platform/network/network_state_notifier_test.cc
++  SetNetworkConnectionInfoOverride(
++    /*on_line*/true,
++    /*type*/WebConnectionType::kWebConnectionTypeUnknown,
```
csagan5, line 75: The default when not observing is 4g, not unknown, so this would make it stand out.

csagan5:
I think it's still not right, but I'll merge it and change it, thanks!

build/patches/Remove-navigator-connection-info.patch
uazo: no, it is necessary, otherwise all the `dchecks` fails.
it's just to minimize the changes

build/patches/Remove-navigator-connection-info.patch
uazo: it already is, the javascript api pass that value if it is `kWebConnectionTypeUnknown`

uazo:
> I think it's still not right

I tried on a real device but always ready to modify if it doesn't work!

PR #2173, opened 2022-07-06
Partitioning Blink memory cache

uazo:
as documented in [1127971](https://bugs.chromium.org/p/chromium/issues/detail?id=1127971), blink's in-memory cache is not partitioned.
normally this is not a problem, but in the event that the rendering process is reused, the cache can be reused to transfer information between different contexts.
a similar thing is present in [brave](https://github.com/brave/brave-core/pull/11013), but the implementation I made is different.

### DIFF
--- a/build/patches/Partitioning-Blink-memory-cache.patch
+++ b/build/patches/Partitioning-Blink-memory-cache.patch
@@ -1,0 +1,194 @@
+From: uazo <uazo@users.noreply.github.com>
+Date: Wed, 6 Jul 2022 10:15:20 +0000
+Subject: Partitioning Blink memory cache
+
+---
+ .../core/html/parser/html_srcset_parser.cc    |  2 +-
+ .../core/inspector/inspector_network_agent.cc |  2 +-
+ .../core/inspector/inspector_page_agent.cc    |  5 +++--
+ .../renderer/core/loader/image_loader.cc      |  3 ++-
+ .../platform/loader/fetch/memory_cache.cc     |  7 ++-----
+ .../platform/loader/fetch/memory_cache.h      |  4 ++--
+ .../platform/loader/fetch/resource_fetcher.cc | 21 ++++++++++++-------
+ .../platform/loader/fetch/resource_fetcher.h  |  3 ++-
+ 8 files changed, 27 insertions(+), 20 deletions(-)
+
+diff --git a/third_party/blink/renderer/core/html/parser/html_srcset_parser.cc b/third_party/blink/renderer/core/html/parser/html_srcset_parser.cc
+--- a/third_party/blink/renderer/core/html/parser/html_srcset_parser.cc
++++ b/third_party/blink/renderer/core/html/parser/html_srcset_parser.cc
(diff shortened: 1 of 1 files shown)


build/patches/Partitioning-Blink-memory-cache.patch
```diff
@@ -0,0 +1,194 @@
...
+ 
+-String ResourceFetcher::GetCacheIdentifier(const KURL& url) const {
++String ResourceFetcher::GetCacheIdentifier(const KURL& url,
++              scoped_refptr<const blink::SecurityOrigin> origin) const {
++  String origin_url = origin ? origin->ToRawString() : "";
++
+   if (properties_->GetControllerServiceWorkerMode() !=
+       mojom::ControllerServiceWorkerMode::kNoController) {
+-    return String::Number(properties_->ServiceWorkerId());
++    return origin_url + "_" + String::Number(properties_->ServiceWorkerId());
```
csagan5, line 163: So if there is an URL:
```
https://www.example.com/mypage
```

and another one
```
https://www.example.com/mypage_1234
```

Then the first one with service worker id=1234 will collide?

uazo: > Then the first one with service worker id=1234 will collide?

no, that change acts in `ResourceFetcher::GetCacheIdentifier` and not on the url, there are no changes to the url

uazo: > will collide

but you are talking about the `CacheIdentifier` ...
that is, are you asking me if it is possible to bypass the identifier to reuse the cache?

uazo: > If it is the same TLD+1 could it cause bugs related to wrong cache being used?

no, at most there is cache-miss

> In any case: why not use a different character than _ (that cannot exist on URLs) so that the problem is avoided altogether?

sounds good, then let's use space

csagan5: Better to use `|`? In printed outputs spaces can be missed.

uazo: spaces, however, are already used to build the Network Isolation Key, perhaps it is better to follow that

uazo: fixed in https://github.com/bromite/bromite/pull/2173/commits/216f4ce5c0bdad54a6e26ae0ca0f2965b9bf35fa

csagan5:
As usual what is in this PR description should go in the patch description so that it can be read without going back to this PR; I will update it for this time.

csagan5:
@uazo have you looked into https://github.com/chromium/chromium/commit/db55141e336c0173c6c624c586f9203a440f6ff7 and what is mentioned there:
> Blink.MemoryCache.RevalidationPolicy with feature ScopeMemoryCachePerContext.

uazo:
yes, I had seen it but
```
This CL does not disallow the memory cache to be used for cross-site
tracking which will be fixed in the follow up CL that implements
a per-fetcher memory cache .
```
in fact it only adds a metric for subsequent checks, but since I don't [see it](https://chromestatus.com/metrics/feature/popularity) I think it has never been active

PR #2182, opened 2022-07-09
[very-wip][do not merge] Remove all site data after page exit

uazo:
I had been saying it for a long time, in the end I did. 
as you may have guessed, all these patches are part of a series of checks that I am doing on the way that are used by the fingerpriting libraries

the idea is to delete all the data after exiting a page (`Page` in the sense of blink, i.e. same domain topframe) by (better) simulating the action of the famous cookie-auto-delete extension.
testing it, neither creepjs nor fingerprintjs (which seem the best libraries to me) are able to link my activity during the same browsing session, a good result I think.

to do this I had to modify the `ConditionalCacheDeletionHelper` to allow selective deletion of all data for correct `NetworkIsolation`, which is currently not managed in chromium.
However, I would like to activate it only for always incognito and under a site setting: we would also have to separate the `LastTabStandingTracker` activation logic from `Add lifetime options for permissions` patch.

what do you think?

### DIFF
--- a/build/patches/wip-remove-site-data-after-page-exit.patch
+++ b/build/patches/wip-remove-site-data-after-page-exit.patch
@@ -1,0 +1,206 @@
+From: uazo <uazo@users.noreply.github.com>
+Date: Wed, 6 Jul 2022 10:38:09 +0000
+Subject: [wip] remove site data after page exit
+
+---
+ .../permissions/last_tab_standing_tracker.cc  | 48 ++++++++++++++++++-
+ .../permissions/last_tab_standing_tracker.h   |  3 +-
+ .../last_tab_standing_tracker_tab_helper.cc   |  4 +-
+ net/http/http_cache.cc                        | 21 +++++++-
+ net/http/http_cache.h                         |  2 +-
+ .../conditional_cache_deletion_helper.cc      |  6 ++-
+ 6 files changed, 77 insertions(+), 7 deletions(-)
+
+diff --git a/chrome/browser/permissions/last_tab_standing_tracker.cc b/chrome/browser/permissions/last_tab_standing_tracker.cc
+--- a/chrome/browser/permissions/last_tab_standing_tracker.cc
++++ b/chrome/browser/permissions/last_tab_standing_tracker.cc
+@@ -10,6 +10,11 @@
+ #include "components/content_settings/core/browser/host_content_settings_map.h"
+ #include "components/content_settings/core/common/content_settings_utils.h"
+ #include "components/permissions/permissions_client.h"
++#include "content/public/browser/storage_partition.h"
++#include "services/network/network_context.h"
++#include "net/base/registry_controlled_domains/registry_controlled_domain.h"
++#include "content/public/browser/clear_site_data_utils.h"
++#include "content/public/browser/browsing_data_filter_builder.h"
+ 
+ namespace {
+   // Remove all sessions content setting by origin and type
+@@ -32,6 +37,16 @@ namespace {
+       }
+     }
+   }
++
++  bool DoesOriginMatchPredicate(
++    base::OnceCallback<bool(const url::Origin&)> predicate,
++    const url::Origin& origin,
++    storage::SpecialStoragePolicy* policy) {
++    if (!std::move(predicate).Run(origin))
++      return false;
++
++    return true;
++  }
+ }
+ 
+ LastTabStandingTracker::LastTabStandingTracker(content::BrowserContext* context)
+@@ -69,7 +84,7 @@ void LastTabStandingTracker::WebContentsLoadedOrigin(
+ }
+ 
+ void LastTabStandingTracker::WebContentsUnloadedOrigin(
+-    const url::Origin& origin) {
++    const url::Origin& origin, content::WebContents* web_contents) {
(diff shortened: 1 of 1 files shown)


uazo:
a really strange thing happens: creepjs keeps the same fingerprinting hash between different sessions:

tryed with official bromite arm64 version: **happen**
tryed with the (my) arm64 no debug version: **happen**
tryed with the (my) arm64 debug version: **happen**

tryed with official bromite x64 version: **happen**

BUT

tryed with the (my) x64 debug version: _**does not happen**_

so i checked with browserstack and the same fingerprint is kept between equal devices, whatever version you use

but the question is why in _my debug version_ doesn't this happen?

all suggestions are welcome, I almost ask the author of the library directly.
in the meantime I compile my non-debug x64

uazo:
@csagan5 can you try with your debug debug version in incognito mode?

uazo:
the problem _was_ the random number generator, **but I couldn't reproduce it anymore**.

that is, in my case, `/dev/urandom` always returned zero, even if entropy was available (with `cat /proc/sys/kernel/random/entropy_avail`).
I checked the documentation and it seems not possible, but that was what happened.
modifying the code as per [1545096 ](https://chromium-review.googlesource.com/c/chromium/src/+/1545096) the bug disappeared, but unfortunately, the next day, it was no longer reproducible.

moreover, in that way I found the timezone bug and verified that, in my opinion, it is necessary to add some mitigation on the svg as well. on this last thing I already have a patch ready, later I make a pull request so we can talk about it.

csagan5:
> the problem _was_ the random number generator, **but I couldn't reproduce it anymore**.
> 
> that is, in my case, `/dev/urandom` always returned zero, even if entropy was available (with `cat /proc/sys/kernel/random/entropy_avail`).

This would be a **major** security problem. What device did you use? Configuration? The most recent changes on that area of the kernel are these: https://www.theregister.com/2022/03/21/new_linux_kernel_has_improved/

But if you can reproduce that, it's a big issue (regardless if it is a bug in kernel or Android).






> moreover, in that way I found the timezone bug and verified that, in my opinion, it is necessary to add some mitigation on the svg as well. on this last thing I already have a patch ready, later I make a pull request so we can talk about it.

I do not follow, what do SVGs and timezones have in common?

uazo:
> I do not follow, what do SVGs and timezones have in common?

I started from this:

image

and I noticed that, for intl and svg the values were **always** the same.
for svg it is normal because [this](https://github.com/uazo/bromite/blob/master-v103/build/patches/changed/Multiple-fingerprinting-mitigations.patch) is missing (I make you the pull so you understand the differences better, we talk about it later there) but I was not expecting the same value for the timezone, hence the bug.
but then I noticed that **always the same** value was also obtained for the canvas, and it shouldn't have happened, hence the search for why and understanding that random failed.

> What device did you use? Configuration? 

I'm currently testing it on a MI5 pro with lineageos A10:
```
gemini:/ $ uname -a
Linux localhost 3.18.124-gedc169a #1 SMP PREEMPT Sun Mar 14 06:14:59 UTC 2021 aarch64
```
I think the kernel/drivers is [this](https://github.com/LineageOS/android_device_xiaomi_gemini), I think it is normal that they are not updated, and probably, like my device, there will be many others out of date.

> This would be a major security problem.

I had seen that problem, basically `urandom` always returns a value (not blocking mode) while `random` blocks if it does not have the value available. I have also read that some kernel versions can empty the entropy value with successive burst calls, but I was unable to reproduce it.
in fact, I just couldn't reproduce that situation at all, which settled after restarting the device (which I do every night).

now, although it is not able to give you a logical and rational reason for the problem, we could probably consider adding [1545096](https://chromium-review.googlesource.com/c/chromium/src/+/1545096) but without actual testing I wouldn't know how effective it is.

uazo:
related: https://bugs.chromium.org/p/chromium/issues/detail?id=1360489

csagan5:
> now, although it is not able to give you a logical and rational reason for the problem, we could probably consider adding [1545096](https://chromium-review.googlesource.com/c/chromium/src/+/1545096) but without actual testing I wouldn't know how effective it is.

No, the random numbers source should not be changed without evidence of a problem. And if it is a problem as you describe it should have an upstream security issue.

> related: https://bugs.chromium.org/p/chromium/issues/detail?id=1360489

This is not related, bounce tracking would work even when removing all site data after page exit.

uazo:
> This is not related, bounce tracking would work even when removing all site data after page exit.

well, however, that's what major browsers basically [currently do](https://github.com/wanderview/bounce-tracking-mitigations/blob/main/explainer.md#references-and-acknowledgements), and it's probably the only thing that can be done without breaking user navigation.

uazo:
accidentally I found that the same functionality exists in the [desktop version](https://source.chromium.org/chromium/chromium/src/+/main:chrome/browser/browsing_data/chrome_browsing_data_lifetime_manager.cc;l=304;drc=727666becab9ab1b26080180df45f043cac5f0aa?q=ChromeBrowsingDataLifetimeManager::ClearBrowsingDataForOnExitPolicy&sq=&ss=chromium%2Fchromium%2Fsrc), one more reason to try to bring it to android as well.

PR #2215, opened 2022-07-25
Add svg fingerprinting mitigation

uazo:
## Description

Adds a mitigation to scripts that use svg to generate hashes for the device fingerprint

## All submissions

### Format


csagan5:
Thanks!

PR #2248, opened 2022-08-08
Update zh_CN translations

zhmars:
## Description

  - add/update translations for patches below
    - Experimental-user-scripts-support.patch
    - History-number-of-days-privacy-setting.patch
    - Add-an-always-incognito-mode.patch
    - Bookmarks-select-all-menu-entry.patch
    - JIT-site-settings.patch
    - Add-webGL-site-setting.patch
    - Add-webRTC-site-settings.patch

@csagan5  I dont have a good enough machine to build bromite. So these changes are untested and I am not 100% sure bromite can be built without issues. But the patch should be applied correctly. Please let me know if I'm missing something.

## All submissions

### Format


csagan5:
Thanks!

PR #2294, opened 2022-08-26
Viewport protection

uazo:
## Description

continue from https://github.com/bromite/bromite/pull/2175

I went ahead with the implementation, and this is the one I like the most of the versions I have developed.
basically it acts on the size of the screen and the view through the use of the emulator mode (which is normally active only in the desktop versions of chromium).
the use of those api allowed me to minimize the changes (because I let chromium behave as normal) and above all to have a greater guarantee about future changes (being an api internal to chromium).
I also added the function that modifies the rect of the dom regardless of how they are defined via css: substantially I have changed the zoom of the page since all the coordinates calculated from that, so:

> Does this mitigation supersede any pre-existing one?

now yes. By disabling the flags of ungoogled, the values do not remain constant, indeed the fingerprinting tools do not detect the anomaly.

the patch in fact reduces (or increases) the space available to the blink view with two factors:
- a value that is multiplied to the current standard size (in css pixels, which is calculated by dividing the effective space of the device by the device pixel ratio, which can be set from the android interface).
- a value that rescales the view (and therefore only keeps the screen at the value calculated above) by acting on the zoom of the page, so that all the rect of the dom are different.

in the code you will also find the management of remote frames (which I finally understood :) that they are the local same-domain IFRAMEs and remote cross-domain ones) because to they I pass the screen value of the top page and inhibit access to the browser controls size (innerWidth/Height == outerWidth/Height).

if you can try to give an eye...
to test it I used

https://canvasblocker.kkapsner.de/test/domRectTest.php
https://abrahamjuliot.github.io/creepjs/tests/domrect.html
https://abrahamjuliot.github.io/creepjs/tests/screen.html
https://dev-pages.brave.software/fingerprinting/farbling.html
https://browserleaks.com/rects
https://privacycheck.sec.lrz.de/active/fp_gcr/fp_getclientrects.html
https://arkenfox.github.io/TZP/tests/domrectspoof.html

but i'm thinking the next step will be to finally start building some bromite test

## All submissions

### Format

### DIFF
--- a/build/patches/Add-Viewport-Protection.patch
+++ b/build/patches/Add-Viewport-Protection.patch
@@ -22,6 +22,7 @@ Subject: Viewport Protection
  .../renderer/content_settings_agent_impl.h    |  1 +
  .../platform/web_content_settings_client.h    |  2 +
  .../blink/renderer/core/css/media_values.cc   | 10 +++
+ .../core/frame/dom_visual_viewport.cc         |  4 +-
  .../renderer/core/frame/local_frame_view.cc   |  3 +
  .../blink/renderer/core/frame/screen.cc       | 17 +++-
  .../renderer/core/frame/visual_viewport.cc    |  4 +
@@ -31,7 +32,7 @@ Subject: Viewport Protection
  third_party/blink/renderer/core/page/page.cc  |  8 ++
  third_party/blink/renderer/core/page/page.h   |  3 +
  .../screen_enumeration/screen_detailed.cc     | 15 ++++
- 28 files changed, 247 insertions(+), 12 deletions(-)
+ 29 files changed, 250 insertions(+), 13 deletions(-)
  create mode 100644 components/browser_ui/site_settings/android/java/src/org/chromium/components/browser_ui/site_settings/BromiteViewportContentSetting.java
  create mode 100644 components/browser_ui/strings/android/viewport.grdp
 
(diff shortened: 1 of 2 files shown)


uazo:
when you have time, can I ask you an opinion on this too?

csagan5:
There is also https://arthuredelstein.github.io/tracking_demos/screen.html (from https://github.com/brave/brave-browser/issues/23170).

So this would make the current fingerprinting patch obsolete, except for the `--fingerprinting-canvas-image-data-noise` part, correct?

Can you please remove the content setting and make it a feature flag for now? We might have a content setting for fingerprinting mitigations in future but I have not yet figured out what is best on that.

uazo:
> So this would make the current fingerprinting patch obsolete

yes, that is the goal.

> except for the --fingerprinting-canvas-image-data-noise part, correct?

is the next one I will make. I have already seen that it is probably possible to change the content of the canvas without the fingerprinting scripts noticing.

> Can you please remove the content setting and make it a feature flag for now? 

as you prefer. 
For the record, I had added it so I could check if it worked.

uazo:
> Can you please remove the content setting and make it a feature flag for now?

in https://github.com/bromite/bromite/pull/2294/commits/fd38a5ab78b0423ea519ab691fabe0a92d6dbbc8 I removed the site setting.
I also keep that version should be needed in the future.

PF4Public:
@uazo Does this PR apply over chromium 112?

uazo:
this one no, it is out of date
you will have to use https://github.com/uazo/bromite/blob/next-v112/build/patches/00Viewport-Protection-flag.patch and https://github.com/uazo/bromite/blob/next-v112/build/patches/00Viewport-Protection-Site-Setting.patch

PR #2334, opened 2022-09-19
Userscripts: fix injection with back-forward cache enabled

uazo:
## Description

fix injection with back-forward cache enabled
Related discussion: https://github.com/bromite/bromite/discussions/2137

## All submissions

### Format


build/patches/Experimental-user-scripts-support.patch
```diff
@@ -10405,15 +10405,15 @@ new file mode 100755
 +  enabled_ = params.allow_userscript;
 +  if (!enabled_) return;
 +
-+  if (loaded_ == false) {
-+    loaded_ = true;
```
csagan5, line 10409: Is `loaded_` going to be inconsistent without this? Do we use `loaded_` anywhere else or can it be removed?

uazo: > or can it be removed?

I have already removed it, look at the line 10486

PR #2335, opened 2022-09-19
Add missed changes for Remove-navigator.connection-info.patch

uazo:
## Description

we missed a change from https://github.com/bromite/bromite/pull/2172. without that the patch does nothing.
I propose it again here

## All submissions

### Format

### DIFF
--- a/build/patches/Remove-navigator.connection-info.patch
+++ b/build/patches/Remove-navigator.connection-info.patch
@@ -8,8 +8,11 @@ and disable observers
 Original License: GPL-2.0-or-later - https://spdx.org/licenses/GPL-2.0-or-later.html
 License: GPL-3.0-only - https://spdx.org/licenses/GPL-3.0-only.html
 ---
- .../blink/renderer/modules/netinfo/network_information.cc  | 7 ++++++-
- 1 file changed, 6 insertions(+), 1 deletion(-)
+ .../renderer/modules/netinfo/network_information.cc   |  7 ++++++-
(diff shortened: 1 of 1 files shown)


csagan5:
I changed it like this on purpose, as I mentioned [here](https://github.com/bromite/bromite/pull/2172#issuecomment-1191294273).

uazo:
> changed it like this on purpose, as I mentioned

I do not understand sorry.
have you changed a patch that works with one that is wrong? am I missing something?

csagan5:
> have you changed a patch that works with one that is wrong? am I missing something?

It is wrong from your perspective. From my perspective your patch is also wrong.

On desktop this is what I see for `navigator.connection`:
```json
{"effectiveType": "4g", "rtt": 0, "downlink": 10, "saveData": false}
```

On mobile:
```json
{
"downlink": 1.7,
"downlinkMax": Infinity,
"effectiveType": "4g",
"rtt": 0,
"saveData": false,
"type": "wifi"
}
```
On another mobile:
```json
{
"downlink": 10,
"downlinkMax": Infinity,
"effectiveType": "4g",
"rtt": 0,
"saveData": false,
"type": "wifi"
}
```

Those defaults in the patch are not correct, they would stand out.

I will add the default as per last example.

uazo:
> It is wrong from your perspective. From my perspective your patch is also wrong.

I honestly do not understand, but you must not be offended.

> Those defaults in the patch are not correct, they would stand out.
> I will add the defaults as per last example.

modify as you see fit, that's fine!

PR #2343, opened 2022-09-21
Remove http referrals in cross origin navigation

uazo:
## Description

the patch removes the referrals if the navigation is cross origin and occurs in the top frame.
the goal is to eliminate that header that is present, for example, in the navigation from the search engine to the destination, and that allows to know the origin of the user's navigation: I detected it with duckduckgo and bing but not with google which correctly uses all browser tools to modify the policy of that header.

I don't remove the value between iframes because the referrals are, let's say, statically defined by the html and the javascript of the page, and therefore they say nothing about the user. Also, some services may not work, such as video iframes.

I also added a flag that completely removes referrals management, for users who know what they are doing.

Also tested here with the excellent privacytests.org

## All submissions

### Format

### DIFF
--- a/build/patches/Remove-http-referrals-in-cross-origin-navigation.patch
+++ b/build/patches/Remove-http-referrals-in-cross-origin-navigation.patch
@@ -2,7 +2,15 @@ From: uazo <uazo@users.noreply.github.com>
 Date: Wed, 21 Sep 2022 12:28:17 +0000
 Subject: Remove http referrals in cross origin navigation
 
-and adds a flag to disable them completely
+The patch removes the referrals if the navigation is cross origin and occurs in the top frame.
+We do not remove the value between the iframes because the referrals are statically defined
+by the html and the javascript of the page, and therefore they do not say anything about the
+user. Also, some services may not work, such as video iframes.
+We also added a flag that completely removes referrals management, for users who know
+what they are doing.
+
+Original License: GPL-2.0-or-later - https://spdx.org/licenses/GPL-2.0-or-later.html
+License: GPL-3.0-only - https://spdx.org/licenses/GPL-3.0-only.html
 ---
  chrome/android/java/res/xml/privacy_preferences.xml       | 4 ++++
  .../chrome/browser/privacy/settings/PrivacySettings.java  | 8 ++++++++
--- a/build/patches/Remove-http-referrals-in-cross-origin-navigation.patch
+++ b/build/patches/Remove-http-referrals-in-cross-origin-navigation.patch
@@ -65,7 +65,7 @@ diff --git a/chrome/browser/ui/android/strings/android_chrome_strings.grd b/chro
  For example, some websites may respond to this request by showing you ads that aren’t based on other websites you’ve visited. Many websites will still collect and use your browsing data — for example to improve security, to provide content, ads and recommendations, and to generate reporting statistics.
        </message>
 +      <message name="IDS_ENABLE_REFERRERS_TITLE" desc="Title for 'Enable referrers' preference">
-+        Enable http referrer header
++        Enable HTTP Referer header
 +      </message>
        <message name="IDS_CAN_MAKE_PAYMENT_TITLE" desc="Title for preference to allow websites to know whether you have payment methods available through PaymentRequest.CanMakePayment interface">
          Access payment methods
--- a/build/patches/Remove-http-referrals-in-cross-origin-navigation.patch
+++ b/build/patches/Remove-http-referrals-in-cross-origin-navigation.patch
@@ -1,0 +1,103 @@
+From: uazo <uazo@users.noreply.github.com>
+Date: Wed, 21 Sep 2022 12:28:17 +0000
+Subject: Remove http referrals in cross origin navigation
+
+and adds a flag to disable them completely
+---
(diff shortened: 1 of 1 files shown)


build/patches/Remove-http-referrals-in-cross-origin-navigation.patch
```diff
@@ -0,0 +1,103 @@
...
+     url::Origin origin_header_value = initiator_origin.value_or(url::Origin());
+diff --git a/services/network/public/cpp/resource_request.h b/services/network/public/cpp/resource_request.h
+--- a/services/network/public/cpp/resource_request.h
++++ b/services/network/public/cpp/resource_request.h
+@@ -130,7 +130,7 @@ struct COMPONENT_EXPORT(NETWORK_CPP_BASE) ResourceRequest {
+   std::vector<GURL> navigation_redirect_chain;
+ 
+   GURL referrer;
+-  net::ReferrerPolicy referrer_policy = net::ReferrerPolicy::NEVER_CLEAR;
++  net::ReferrerPolicy referrer_policy = net::ReferrerPolicy::REDUCE_GRANULARITY_ON_TRANSITION_CROSS_ORIGIN;
```
uazo, line 98: this i think is a chromium bug, because that policy does not match the default and is used as is in service workers

build/patches/Remove-http-referrals-in-cross-origin-navigation.patch
```diff
@@ -0,0 +1,103 @@
...
+--- a/chrome/android/java/res/xml/privacy_preferences.xml
++++ b/chrome/android/java/res/xml/privacy_preferences.xml
+@@ -57,6 +57,10 @@
+         android:title="@string/close_tabs_on_exit_title"
+         android:summary="@string/close_tabs_on_exit_summary"
+         android:defaultValue="false" />
++    <org.chromium.components.browser_ui.settings.ChromeSwitchPreference
++        android:key="enable_referrers"
++        android:title="@string/enable_referrers_title"
++        android:defaultValue="false" />
```
csagan5, line 24: Shouldn't the default be `true`?

build/patches/Remove-http-referrals-in-cross-origin-navigation.patch
```diff
@@ -0,0 +1,103 @@
...
+             preloadPagesPreference.setSummary(
+diff --git a/chrome/browser/ui/android/strings/android_chrome_strings.grd b/chrome/browser/ui/android/strings/android_chrome_strings.grd
+--- a/chrome/browser/ui/android/strings/android_chrome_strings.grd
++++ b/chrome/browser/ui/android/strings/android_chrome_strings.grd
+@@ -873,6 +873,9 @@ CHAR_LIMIT guidelines:
+ 
+ For example, some websites may respond to this request by showing you ads that aren’t based on other websites you’ve visited. Many websites will still collect and use your browsing data — for example to improve security, to provide content, ads and recommendations, and to generate reporting statistics.
+       </message>
++      <message name="IDS_ENABLE_REFERRERS_TITLE" desc="Title for 'Enable referrers' preference">
++        Enable http referrer header
```
csagan5, line 68:
```suggestion
+        Enable HTTP Referer header
```

(Here the mispelling `Referer` is on purpose)

build/patches/Remove-http-referrals-in-cross-origin-navigation.patch
csagan5: It seems a quite big bug?

csagan5:
This is really nice work!

Can you please move more details in the patch description? The PR description is much more verbose.

build/patches/Remove-http-referrals-in-cross-origin-navigation.patch
uazo: I don't know what to tell you. I noticed it by logging in [URLRequestJob::ComputeReferrerForPolicy](https://source.chromium.org/chromium/chromium/src/+/main:net/url_request/url_request_job.cc;l=290;drc=f97e7e130b02a0fee5a06aa9cdf25d3a0a3715d0;bpv=1;bpt=1)

I noticed that the chromium team has to (always) follow the specs, maybe the specs say nothing about it (I did not check). in that case they often change the specifications first, then the code, maybe this is going on.

build/patches/Remove-http-referrals-in-cross-origin-navigation.patch
uazo: it's the same. the value is set from see line 54 while the default is that of prefs

uazo:
> This is really nice work!

thanks :) tomorrow I change the description.

build/patches/Remove-http-referrals-in-cross-origin-navigation.patch
csagan5: Perhaps there is something else going on that makes this bug less important; have you tried submitting it upstream as a security issue? Maybe they will mark yours as a duplicate of the main one.

uazo: > have you tried submitting it upstream as a security issue?

no, maybe I'll try, but I wouldn't classify it as a security bug, after all, sites can choose how to manage referrals

PR #2404, opened 2022-10-22
Patches for v107

uazo:
## Description

# List of changes:

* [x] Add a proxy configuration page:
	Upstream removed [PrefService::GetDictionary]( https://source.chromium.org/chromium/chromium/src/+/b7f43fd50ad1564fd1a32ebd57d230d2e9faff2a)
	
* [x] Add lifetime options for permissions
	Upstream changes: [Safety check: Refactor content_settings to pass bundled metadata](https://source.chromium.org/chromium/chromium/src/+/0643ed40a69f30ec346a347f15ea870439bc7c20)

* [x] Allow building without enable_reporting
* [x] Disable TLS resumption
	Upstream changes: [Partition network state by new partition key](https://bugs.chromium.org/p/chromium/issues/detail?id=1343856)
	
* [x] Client hints overrides
	Upstream changes: [Remove UserAgentClientHintFullVersionList feature](https://bugs.chromium.org/p/chromium/issues/detail?id=1332632)
	
* [x] Content settings infrastructure
	minor change: added a arg to ContentSettingsResources.getSiteSummary()
	
* [x] Disable conversion measurement api
	minor change: removed namespace
	
* [x] Disable FLoC and privacy sandbox
	minor change: rename IsConversionMeasurementAllowed() in IsAttributionReportingAllowed()
	
* [x] Disable text fragments by default
	readded kTextFragmentAnchor
	to check try: https://blog.chromium.org/2019/12/chrome-80-content-indexing-es-modules.html#:~:text=ECMAScript%20Modules%20in%20Web%20Workers

* [x] Enable native Android autofill
	Refactored due to [[Autofill] Eliminate ContentAutofillDriver::FooImpl() functions.](https://source.chromium.org/chromium/chromium/src/+/b1abfe126b0f21aff4a0cec06a3a4c812a906244)
	
* [x] Experimental user scripts support
	Upstream changes: [Remove PrefService::GetDictionary](https://source.chromium.org/chromium/chromium/src/+/b7f43fd50ad1564fd1a32ebd57d230d2e9faff2a)
Merged lost changes

* [x] Improve plain text rendering on mobile
	Upstream changes: [blink: makes use of enum for html tag names](https://source.chromium.org/chromium/chromium/src/+/ee80d5788e4075df94e7fa4ec0c1849e6cf5270c)
	
* [x] Increase number of autocomplete matches to 10
	removed "kDefaultMaxAutocompleteMatches = 10;" already in upstream

* [x] Move navigation bar to bottom
	no code changes, but I find a glitch that I have to fix

* [ ] Remove navigator.connection info
	fix DCHECK()
	
* [ ] Revert clipboard user gesture requirement removal
	we need to check https://source.chromium.org/chromium/chromium/src/+/10300ac7da93a7e322274f1ed89395a7c639986e
	but actually works as intended: check with https://output.jsbin.com/yuliwoz/quiet
	
* [x] Revert "Delete block-external-form-redirects"
	changed "if (DEBUG)" with "if (debug())"
	
* [ ] Revert removal of execution context address space
	work in progress changes from upstream
	see https://bugs.chromium.org/p/chromium/issues/detail?id=1130587

* [x] Timezone customization
	minor changes
	upstream make some modification like we done, so some code is dropped

* [x] ungoogled-chromium: no special hosts/domains
	~removed also "switches::kGoogleBaseURL"~

* [x] Welcome screen
	minor changes, some variable were renamed
		
# Patches dropped (already in upstream)
- 	Enable-user-agent-freeze-by-default.patch
- 	Partition-DNS-requests-by-top-frame-NIK.patch
- 	Enable-HEVC-by-default.patch
- 	Fix segmentation platform build error

# Additional notes
remember, there is no `Automated-domain-substitution.patch` and patches in https://github.com/bromite/bromite/commit/b6d2fee70924277fd079d9e17538f383ff31546e can be different from the way you make them

## All submissions

    it should, I launched a build as a verification
    https://github.com/uazo/bromite-buildtools/actions/runs/3302979999

### Format


uazo:
I apologise, I did delete `CommandLineGoogleBaseURL()` but I shouldn't have.
the reason for the confusion is that I also have [other changes](https://github.com/uazo/bromite/blob/next-v107/build/patches/00ungoogled-chr-no-special-hosts-domains--add-fixup-.patch) to `ungoogled-chr-no-special-hosts-domains` but which I did not want to introduce in this pull request but separately, so we can talk about the reasons.
with https://github.com/bromite/bromite/pull/2404/commits/dfdb8efa0263e8cfdc204cc8948b460eea343027 it build

uazo:
I fixed the glitch, which is the same as that described in #2407.
As I was editing the patch, I also [reversed the order in the tab list](https://github.com/bromite/bromite/discussions/2327#discussioncomment-3661593), which now goes from bottom to top.

uazo:
I have checked all the cve listed in the windows stable version, and unfortunately it is necessary to switch to the 107.0.5304.63, some bugs are in the common code and especially in blink.
it's easy because I already have the tool that does it automatically, but I don't know where you are with the porting, I don't want to mess you up. you tell me. 

then, if you have time, a huge doubt.
in https://chromereleases.googleblog.com/2022/10/stable-channel-update-for-desktop_25.html refer to 
[1377543](https://bugs.chromium.org/p/chromium/issues/detail?id=1377543) which in turn indicates [1361612](https://crbug.com/1361612) among the high bugs, but on checking it refers to https://webrtc.googlesource.com/src.git/+log/d1875a232c3b..b440c4579767 which, however, is in v108 and not v107.
We care little because webrtc is disabled by default, but what do you think is wrong with my reasoning anyway?

uazo:
> then, if you have time, a huge doubt.

nothing, they noticed it too!
came out with the [107.0.5304.88](https://chromium.googlesource.com/chromium/src/+/107.0.5304.88%5E%21/#F0) which has those fixes inside

csagan5:
Slightly related, see this comment from Brave authors about patches integration: https://github.com/brave/brave-browser/issues/6378#issuecomment-725773778

uazo:
>> see this comment from Brave authors
>
particularly this sentence:
>
> It would create far too much of a burden on our chromium update team who already have a difficult job keeping up with upstream changes.

their model is different from ours: for them, patches are used to wedge in the key points of chromium and replace the code with classes that inherit the code base, their changes must stay there.
for us it is easier to heavily modify chromium, but at the same time it is more complex to port in new versions.

but when I do porting it takes me a couple of days, not full-time of course.
I spend a lot of time in:
- adjusting the first patch to the new version, note: I don't care that the build works, but more that I don't miss pieces of patches.
- reassigning changes to the correct patches, once the build goes well
- checking if everything works!
- write down what I modified :)

the time for changes to make the build possible is a part that I think is small compared to the total (for me).
I need to automate those points

chirayudesai:
Do you cherry-pick the patches from a previous branch, or use `git am`?

I've found that `cherry pick` will deal with trivial conflicts very easily and that reduces a lot of the work. I'd like to somehow involve git in the workflow more in the future.

Another thing I'd like to bring up at some point is have enterprise features be considered? For example, there's a policy to disable safe browsing, and I wonder if we could simply turn that on in the code always to make that patch smaller.

Happy to open a separate discussion about these, or just send patch(es) whenever I get around to actually trying that approach.

uazo:
> Do you cherry-pick the patches from a previous branch, or use `git am`?

`git am`. and when it fails, I use `wiggle`, and when it fails I fix it by hand

> I've found that `cherry pick` will deal with trivial conflicts very easily and that reduces a lot of the work. I'd like to somehow involve git in the workflow more in the future.

I will try, do you have something automated?

> Another thing I'd like to bring up at some point is have enterprise features be considered? For example, there's a policy to disable safe browsing, and I wonder if we could simply turn that on in the code always to make that patch smaller.

Approaches other than "delete and chase" should be evaluated.
it would be interesting to test any patch that exploits policy, to see how far it goes.
another approach might be to leave it there and eliminate UI and all external calls (via net or intent).

> Happy to open a separate discussion about these

happy too :)

chirayudesai:
> > Do you cherry-pick the patches from a previous branch, or use `git am`?
> 
> `git am`. and when it fails, I use `wiggle`, and when it fails I fix it by hand
> 
> > I've found that `cherry pick` will deal with trivial conflicts very easily and that reduces a lot of the work. I'd like to somehow involve git in the workflow more in the future.
> 
> I will try, do you have something automated?

Yeah, I'll document at the discussion to avoid polluting this PR further.

https://github.com/bromite/bromite/discussions/2423

> 
> > Another thing I'd like to bring up at some point is have enterprise features be considered? For example, there's a policy to disable safe browsing, and I wonder if we could simply turn that on in the code always to make that patch smaller.
> 
> Approaches other than "delete and chase" should be evaluated. it would be interesting to test any patch that exploits policy, to see how far it goes. another approach might be to leave the function we would like to eliminate there (for example safe-browsing) and remove the UI and all external calls (via net or intent).

I'll dig up some info and post at https://github.com/bromite/bromite/discussions/2424

csagan5:
> * Allow building without enable_reporting

No changes here?

> * Disable text fragments by default
>   readded kTextFragmentAnchor
>   to check try: https://blog.chromium.org/2019/12/chrome-80-content-indexing-es-modules.html#:~:text=ECMAScript%20Modules%20in%20Web%20Workers

I am trying to find the upstream commit where this was removed; in `runtime_enabled_features.json5` I noticed you set `status` to `test`, but in the original patch I keep it removed. What is the reasoning behind that? AFAIK it has to be removed.



> * Experimental user scripts support
>   Upstream changes: [Remove PrefService::GetDictionary](https://source.chromium.org/chromium/chromium/src/+/b7f43fd50ad1564fd1a32ebd57d230d2e9faff2a)
>   Merged lost changes

Which lost changes? Haven't checked your patch yet.

> Enable native Android autofill
> Refactored due to [[Autofill] Eliminate ContentAutofillDriver::FooImpl() functions.](https://source.chromium.org/chromium/chromium/src/+/b1abfe126b0f21aff4a0cec06a3a4c812a906244)

Added.

> Increase number of autocomplete matches to 10
> removed "kDefaultMaxAutocompleteMatches = 10;" already in upstream

Funny coincidence! After several years, someone did it upstream :)



> * Move navigation bar to bottom
>   no code changes, but I find a glitch that I have to fix

Can you please open an issue for it? The patches, like the bugs, are all public.



> * Revert removal of execution context address space
>   work in progress changes from upstream
>   see https://bugs.chromium.org/p/chromium/issues/detail?id=1130587

I fixed a couple of issues with this, didn't check your patch yet.



> * ungoogled-chromium: no special hosts/domains
>   ~removed also "switches::kGoogleBaseURL"~

I am not changing anything in this patch. If you think there should be changes please let's discuss.



> * Timezone customization
>   minor changes
>   upstream make some modification like we done, so some code is dropped

I have adopted this change from yours.

I am not done yet, will go through the rest later.



> their model is different from ours: for them, patches are used to wedge in the key points of chromium and replace the code with classes that inherit the code base, their changes must stay there.

I think that my point stands: changes to the Chromium codebase are hard to maintain - no matter if it's for subclassing or for adjusting the existing code.

csagan5:
> * Move navigation bar to bottom
>   no code changes, but I find a glitch that I have to fix

I see changes related to the title provider: `invert the order if the toolbar is at bottom`

csagan5:
> * Revert clipboard user gesture requirement removal
>   we need to check https://source.chromium.org/chromium/chromium/src/+/10300ac7da93a7e322274f1ed89395a7c639986e
>   but actually works as intended: check with https://output.jsbin.com/yuliwoz/quiet

I dropped the revert patch since the upstream fix covers the original problem already, I also would have liked the alternative approach mentioned there instead:
> * A permission prompt could always be required (user gesture or not).
> This introduces permission fatigue and breaks legitimate use cases such as Gerrit (click one of the "copy to clipboard" buttons next to a commit hash or filename below).

There are probably some attacks that can be done this way (by silently overwriting the user clipboard).

csagan5:
> * Enable-user-agent-freeze-by-default.patch
> * Enable-HEVC-by-default.patch

Converted to guards.

> * Partition-DNS-requests-by-top-frame-NIK.patch

I see that your patch applied correctly though. Does it mean that it is redundant now and must be removed?

uazo:
> > * Allow building without enable_reporting
> 
> No changes here?

Yes, the same as `Disable TLS resumption` (Upstream changes: [Partition network state by new partition key](https://bugs.chromium.org/p/chromium/issues/detail?id=1343856))


> > * Disable text fragments by default
> >   readded kTextFragmentAnchor
> >   to check try: https://blog.chromium.org/2019/12/chrome-80-content-indexing-es-modules.html#:~:text=ECMAScript%20Modules%20in%20Web%20Workers
> 
> I am trying to find the upstream commit where this was removed; 

https://bugs.chromium.org/p/chromium/issues/detail?id=919204
https://chromium-review.googlesource.com/c/chromium/src/+/2606129

> in `runtime_enabled_features.json5` I noticed you set `status` to `test`, but in the original patch I keep it removed. What is the reasoning behind that? AFAIK it has to be removed.

in order for the setting to be generated in blink, otherwise the build will break. with `test` is deactivated by default.

uazo:
> Partition-DNS-requests-by-top-frame-NIK.patch
> I see that your patch applied correctly though. Does it mean that it is redundant now and must be removed?

yes, upstream developed it

uazo:
I also realised that it must be disabled `kStartSurfaceAndroid` in `Restore classic new tab page`.

csagan5:
> > in `runtime_enabled_features.json5` I noticed you set `status` to `test`, but in the original patch I keep it removed. What is the reasoning behind that? AFAIK it has to be removed.
> 
> in order for the setting to be generated in blink, otherwise the build will break. with `test` is deactivated by default.

:+1: 



> > Partition-DNS-requests-by-top-frame-NIK.patch
> > I see that your patch applied correctly though. Does it mean that it is redundant now and must be removed?
> 
> yes, upstream developed it

Removed it.



> I also realised that it must be disabled `kStartSurfaceAndroid` in `Restore classic new tab page`.

Why?

> * Increase number of autocomplete matches to 10
>   removed "kDefaultMaxAutocompleteMatches = 10;" already in upstream

I did not find this in upstream, it's 8 there.



> > > * Disable text fragments by default
> > >   readded kTextFragmentAnchor
> > >   to check try: https://blog.chromium.org/2019/12/chrome-80-content-indexing-es-modules.html#:~:text=ECMAScript%20Modules%20in%20Web%20Workers
> > 
> > 
> > I am trying to find the upstream commit where this was removed;
> 
> https://bugs.chromium.org/p/chromium/issues/detail?id=919204
> https://chromium-review.googlesource.com/c/chromium/src/+/2606129

Thanks; so the revert of that commit is now merged into the patch.

Can you please check if there is anything useful that I missed from your patches?

uazo:
>> I also realised that it must be disabled kStartSurfaceAndroid in Restore classic new tab page.
>
> Why?

I realised this afterwards.
with that function active, the empty page of the ui (i.e. the one without tabs) displays the google search box and any user action leads to crashes until a new tab is created.
Furthermore, the view source patch crashes the browser because the condition `tab == null` is not verified, and thus opening the appmenu with those premises leads to crashes. 
In v107 is enabled while in v106 is disabled, you can check it with https://github.com/uazo/bromite-flags-list.

> I did not find this in upstream, it's 8 there.

https://source.chromium.org/chromium/chromium/src/+/main:components/omnibox/browser/autocomplete_result.cc;l=81;bpv=1;bpt=1?q=AutocompleteResult::GetMaxMatches&ss=chromium%2Fchromium%2Fsrc

> Can you please check if there is anything useful that I missed from your patches?

sure, after I check

csagan5:
> with that function active, the empty page of the ui (i.e. the one without tabs) displays the google search box and any user action leads to crashes until a new tab is created.

Ok, thanks for the explanation; should the search box be removed from the "start surface" feature or is the whole feature just about showing a search box?
What are these other crashes due to (not the view-source one), some other Bromite patch?

> Furthermore, the view source patch crashes the browser because the condition tab == null is not verified, and thus opening the appmenu with those premises leads to crashes.

I will add a fix for this.

> > I did not find this in upstream, it's 8 there.
> 
> https://source.chromium.org/chromium/chromium/src/+/main:components/omnibox/browser/autocomplete_result.cc;l=81;bpv=1;bpt=1?q=AutocompleteResult::GetMaxMatches&ss=chromium%2Fchromium%2Fsrc

I see it now, thanks.

uazo:
> should the search box be removed from the "start surface" feature or is the whole feature just about showing a search box?
What are these other crashes due to (not the view-source one), some other Bromite patch?

I must tell you the truth, I did not investigate. I can look into it

csagan5:
> * Remove navigator.connection info
>   fix DCHECK()

I see some other differences there, we need to discuss this.

csagan5:
> I must tell you the truth, I did not investigate. I can look into it

Don't feel compelled to, I was just wondering. I will disable this feature for the time being.

uazo:
> I see some other differences there, we need to discuss this.

if you agree, after the release of 107, at least work on the same patches as you.

csagan5:
I will do the release unless there is something big that I missed.

uazo:
do you want me to check https://github.com/bromite/bromite/pull/2435?

csagan5:
> do you want me to check #2435?

Yes, that's what I meant in [this comment](https://github.com/bromite/bromite/pull/2404#issuecomment-1312809721).

But as usual, no hurry.

PR #2435, opened 2022-11-06
Patches for v107

csagan5:
## Description

Patches for v107; notable changes:
* moved big/often-breaking patches to the top; this required some re-ordering of the patches
* moved site-settings related patches to the top

## All submissions

### Format


csagan5:
Let's finish the discussions in #2404 before discussing this.

csagan5:
@uazo I will soon merge this PR; in future perhaps you could make a PR which updates only the patches you maintain? Leaving out all the rest; then I can merge that directly from here after review.

With this PR I am also moving some of the big patches to the top, so in theory I could release the fixes to those early so that others can use that as a starting point and save time.

uazo:
> in future perhaps you could make a PR which updates only the patches you maintain?

sure, let's try it. 
I also have the problem of changes related to the issues you assigned me. i have them ready but i don't know how to communicate them to you, if I give them to you as they are now, the patches might not apply since the line number might be different....
we have to find a way

> With this PR I am also moving some of the big patches to the top, so in theory I could release the fixes to those early so that others can use that as a starting point and save time.

that's fine, but it always depends on the timing, for now it's not useful for me.
for example for 108 as early as tomorrow I could work on it...

csagan5:
> > in future perhaps you could make a PR which updates only the patches you maintain?
> 
> sure, let's try it. I also have the problem of changes related to the issues you assigned me. i have them ready but i don't know how to communicate them to you, if I give them to you as they are now, the patches might not apply since the line number might be different.... we have to find a way

I think that's not a problem? Before merging the PR I will try to apply them; we can also automate this in future but I don't see it as a problem right now.

> > With this PR I am also moving some of the big patches to the top, so in theory I could release the fixes to those early so that others can use that as a starting point and save time.
> 
> that's fine, but it always depends on the timing, for now it's not useful for me. for example for 108 as early as tomorrow I could work on it...

I am thinking to parallelise the work, if we split it in patch ranges, and avoid double work.

uazo:
> I think that's not a problem?
> I am thinking to parallelise the work, if we split it in patch ranges, and avoid double work.

we should talk about it, I have some ideas too, we should start from the difficulties we encounter to find a solution.
I think it could also be a lack of tools, or maybe I don't use them as well as I could, non so.

build/patches/Client-hints-overrides.patch
```diff
@@ -160,23 +160,23 @@ diff --git a/third_party/blink/common/client_hints/enabled_client_hints.cc b/thi
...
- #if BUILDFLAG(IS_ANDROID)
+ // Enables concurrently marking Blink's heap.
+ const base::Feature kBlinkHeapConcurrentMarking{
+     "BlinkHeapConcurrentMarking", base::FEATURE_ENABLED_BY_DEFAULT};
+@@ -611,7 +611,7 @@ const base::Feature kDispatchBeforeUnloadOnFreeze{
+ const base::Feature kLowLatencyCanvas2dImageChromium {
+   "LowLatencyCanvas2dImageChromium",
+ #if BUILDFLAG(IS_CHROMEOS_ASH)
 -      base::FEATURE_ENABLED_BY_DEFAULT
 +      base::FEATURE_DISABLED_BY_DEFAULT
```
uazo: `LowLatencyCanvas2dImageChromium` is not the correct flag.
we are bringing us the change from 104, where it was `AllowClientHintsToThirdParty` that was removed.

build/patches/Disable-smart-selection-by-default.patch
```diff
@@ -98,9 +98,9 @@ diff --git a/testing/variations/fieldtrial_testing_config.json b/testing/variati
 --- a/testing/variations/fieldtrial_testing_config.json
 +++ b/testing/variations/fieldtrial_testing_config.json
 @@ -2705,6 +2705,21 @@
-             ]
-         }
-     ],
+                     "params": {
+                         "use_root_bookmark_as_default": "true"
+                     },
 +    "ChromeSmartSelection": [
```
uazo: surely you have noticed, probably the build does not fit, this is not the right position.

uazo (commented):
two little things, but maybe you had already noticed. the rest is ok for me!

build/patches/Client-hints-overrides.patch
csagan5: There were actually two here incorrect, I had only caught one. Removed both.

build/patches/Disable-smart-selection-by-default.patch
csagan5: No, I did not notice this and it does not fail the build; corrected.

PR #2447, opened 2022-11-17
v107 fix some build errors

uazo:
## Description

I found some errors in the build (maybe you didn't clean the out directory?)
small things, but I will make the changes available to you.

```
../../chrome/browser/commerce/price_tracking/android/price_tracking_notification_bridge.cc:9:10: fatal error: 'chrome/browser
/commerce/price_tracking/android/jni_headers/PriceTrackingNotificationBridge_jni.h' file not found                           #include "chrome/browser/commerce/price_tracking/android/jni_headers/PriceTrackingNotificationBridge_jni.h"
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```

```
In file included from ../../chrome/browser/file_system_access/chrome_file_system_access_permission_context.cc:53:
../../components/safe_browsing/content/common/file_type_policies.h:14:10: fatal error: 'components/safe_browsing/content/common/proto/download_file_types.pb.h' file not found
#include "components/safe_browsing/content/common/proto/download_file_types.pb.h"
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```

```
ld.lld: error: undefined symbol: PriceTrackingNotificationBridge::GetForBrowserContext(content::BrowserContext*)
>>> referenced by optimization_guide_keyed_service.cc:118 (../../chrome/browser/optimization_guide/optimization_guide_keyed_service.cc:118)
>>>               browser/optimization_guide_keyed_service.o:(OptimizationGuideKeyedService::MaybeCreatePushNotificationManager(Profile*)) in archive obj/chrome/browser/libbrowser.a

```

## All submissions

### Format

### DIFF
--- a/build/patches/Remove-signin-and-sync-integrations.patch
+++ b/build/patches/Remove-signin-and-sync-integrations.patch
@@ -87,7 +87,7 @@ License: GPL-3.0-only - https://spdx.org/licenses/GPL-3.0-only.html
  .../browser/firstrun/MobileFreProgress.java   |   4 +-
  .../router/discovery/access_code/BUILD.gn     |   1 -
  .../browser/password_manager/android/BUILD.gn |   4 -
- ...sswordManagerErrorMessageHelperBridge.java |  16 --
+ ...sswordManagerErrorMessageHelperBridge.java |  41 +---
  .../PasswordManagerHelper.java                | 121 +---------
  ...swordSyncControllerDelegateBridgeImpl.java |   7 -
  chrome/browser/privacy/BUILD.gn               |   1 -
@@ -127,7 +127,7 @@ License: GPL-3.0-only - https://spdx.org/licenses/GPL-3.0-only.html
  .../ProfileOAuth2TokenServiceDelegate.java    |  52 ----
  .../net/HttpNegotiateAuthenticator.java       |  88 +------
  .../chromoting/base/OAuthTokenFetcher.java    |   2 -
- 108 files changed, 101 insertions(+), 2645 deletions(-)
+ 108 files changed, 102 insertions(+), 2669 deletions(-)
 
 diff --git a/chrome/android/BUILD.gn b/chrome/android/BUILD.gn
 --- a/chrome/android/BUILD.gn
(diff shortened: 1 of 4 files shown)


build/patches/Remove-signin-and-sync-integrations.patch
```diff
@@ -3719,16 +3719,48 @@ diff --git a/chrome/browser/password_manager/android/java/src/org/chromium/chrom
...
+-        Profile profile = Profile.getLastUsedRegularProfile();
+-        final CoreAccountInfo primaryAccountInfo =
+-                IdentityServicesProvider.get().getIdentityManager(profile).getPrimaryAccountInfo(
+-                        ConsentLevel.SIGNIN);
+-        // It is possible that the account is removed from Chrome between the password manager
+-        // calling the Google Play Services backend and Chrome receiving the reply. In that
+-        // case, the error is no longer relevant/fixable.
+-        if (primaryAccountInfo == null) return false;
+-
+-        if (ChromeFeatureList.getFieldTrialParamByFeatureAsBoolean(
```
csagan5, line 3735: I was thinking to remove only the part above this line.

csagan5 (approved):
I confirm that they happen if not building Chromium first.

build/patches/Remove-signin-and-sync-integrations.patch
uazo: I think it is irrelevant, this method should never be called

csagan5: But the patch should remove only the parts which are harmful/unwanted, to be smaller.

uazo: sure, but the patch must also be useful, and it seems to me that it is not, today.

from what i'm understanding with the policy work, sync is disabled because the API key is missing, while signin is disabled if the user has not given consent and anyway it doesn't seem to me to be active without the MIRROR mode, that we have disabled.
the only case of calling google's servers, is related to the DICE mode which is only active in the non-android version, but I think this is more a bug than a will.
but I am still lacking certainty

csagan5: It is useful because I would get a build error due to the reference to `IdentityServicesProvider`, so I am removing only that part.

Issue #2387, opened 2022-10-10
Unable to access Zoom meeting


shouzhong:
### Preliminary checklist

- [X] I have read the [README](https://github.com/bromite/bromite/blob/master/README.md).
- [X] I have searched the existing issues for my problem. This is a new ticket, NOT a duplicate or related to another open issue.
- [X] I have read the [FAQs](https://github.com/bromite/bromite/blob/master/FAQ.md).
- [X] I have updated Bromite to the latest version. The bug is reproducible on this latest version.
- [X] This is a bug report about the Bromite browser; not the website nor F-Droid nor anything else.

### Can the bug be reproduced with corresponding Chromium version?

No

### Bromite version

105.0.5195.147

### Device architecture

armeabi-v7a

### Android version

9

### Device model

Tablet

### Changed flags

no flags changed.

### Is this bug about the SystemWebView?

No

### Is this bug happening in an incognito tab?

Yes

### Is this bug caused by the adblocker?

No

### Is this bug a crash?

No

### Describe the bug

When I try to start a Zoom meeting on Bromite,  it will keep stuck on the 'Joining Meeting..' page and can't enter the meeting.
But working fine on https://www.apkmirror.com/apk/bromite/chromium/chromium-105-0-5195-147-release/.
Can you give some advice on how to resolve this issue?

### Steps to reproduce the bug

1. Open https://zoom.us
2. Sign in
3. 
![2022 10 10_吴一丰 e0a4879a43206c230eae0fc768d77244](https://user-images.githubusercontent.com/27850869/194815454-9acd820d-d4f1-4cfd-a7fd-20df1099919a.png)
4. 
![2022 10 10_吴一丰 e56aeff9fb3d5394d78dcd10fd147c75](https://user-images.githubusercontent.com/27850869/194815513-510eb72f-422a-4b48-99de-bdf087b167ac.png)

### Expected behavior

Able to access the meeting.

### Screenshots

![2022 10 10_吴一丰 d1a051c177f027cb39b9fee49da675ea](https://user-images.githubusercontent.com/27850869/194815673-215e90ad-7a0d-41f4-a67d-80ba20be3794.png)

uazo:
at first glance, I would say that you have not enabled webrtc from the site settings.
could you also check and report the error log in chromium devtools?

shouzhong:
Great, I can enter the meeting after enabling webrtc and webgl.
[...]

PR #2450, opened 2022-11-17
Adds the 'audio-context-api' flag

uazo:
## Description

Adds:

- 'audio-context-api' flag (default true) which allows the activation of the webaudio api 
The api is not available to javascript (throws an error) by deactivating the flag.

- 'fingerprinting-audio-context-data-noise' flag (default true) that modify the values sent to javascript by a random factor.

The code is loosely inspired by brave's code, with the difference that the random factor is recalculated each time the buffer is requested, for each sample, since it seems very simple to derive that value and then acquire the original value (ref https://fingerprint.com/blog/audio-fingerprinting)

Fix for #2387

## All submissions

### Format


build/patches/Add-AudioContextAPI-flags.patch
```diff
@@ -0,0 +1,546 @@
...
+ 
+diff --git a/third_party/blink/common/features.cc b/third_party/blink/common/features.cc
+--- a/third_party/blink/common/features.cc
++++ b/third_party/blink/common/features.cc
+@@ -561,6 +561,11 @@ const base::Feature kWebRtcUseMinMaxVEADimensions {
+ #endif
+ };
+ 
++const base::Feature kAudioContextAPIEnabled {
++  "AudioContextAPIEnabled", base::FEATURE_ENABLED_BY_DEFAULT};
```
csagan5: Can you please remove this feature? It is sufficient to have the below one.

uazo: fixed in https://github.com/bromite/bromite/pull/2450/commits/1b8e66db42c5caf533f4a5d5fcc61ec5bde7e452

PR #2451, opened 2022-11-17
Fix clean builds

chirayudesai:
* Both of these caused errors when doing a clean Bromite-only build.

## Description

*Please explain here what feature or bugfix these changes are addressing and why they should be included*

## All submissions

* [x] I have tested that the new change works as intended (AVD or physical device will do)

### Format


chirayudesai:
Builds fine, yet to test built APK.

Without this, build failed, I remember reporting about it before and being told that building vanilla chromium beforehand in the same output dir worked too.

chirayudesai:
> Builds fine, yet to test built APK.

Did a quick smoke test of APK and it seems fine.

chirayudesai:
Ah nevermind, just noticed https://github.com/bromite/bromite/pull/2447

PR #2454, opened 2022-11-18
Disable all predictors code: remove another method

uazo:
## Description

found another method to remove

## All submissions

part of the patch code is also present in #2447

### Format


csagan5:
Is this code ever reached in Bromite?

uazo:
> Is this code ever reached in Bromite?

currently only from the shopping feature, which is, however, disabled by another patch, but could be in the future since the comment
```
// TODO(crbug/1275612): Check whether we have consent to fetch.
```

csagan5:
Cherry-picked in be2a93f8b3557d9ec41e4edbed4cd2d732165275

PR #2507, opened 2022-12-18
Use helper macros to define base::Features

uazo:
## Description

Followed upstream for:
- Use helper macros to define base::Features
see https://bugs.chromium.org/p/chromium/issues/detail?id=1364289

## All submissions

### Format


csagan5:
I'll manually fix the conflicts on my v109 branch.

csagan5:
I think there are still patches that need the fix.

PR #2508, opened 2022-12-18
Fix some lost changes after v108

uazo:
## Description

We missed some changes during the porting to 108.
I propose them to you again

## All submissions

### Format


csagan5:
Are these changes that should have been submitted as individual PRs?

PR #2524, opened 2022-12-29
Fixes the order of the tabs in the tablist

### LINKED ISSUE #2456: Tabs View Inverted
### Preliminary checklist

- [X] This is a bug report about the Bromite browser (not about the website, building Bromite, F-Droid or anything else).

No

### Bromite version

107.0.5304.96

### Device architecture

arm64

### Android version

12.0

### Device model

LG Wing (LM-100VM)

### Changed flags

Move top toolbar to bottom
Move the top toolbar to the bottom. – Android

#move-top-toolbar-to-bottom

No

Yes

No

### Is this bug a crash?

No

### Describe the bug

The tab view is now inverted. The previous public release did not have this issue. The newest tabs opened are in the top left when they used to be in the bottom right. This makes reaching the new tabs near impossible on larger devices. Please at least allow this to be set to the old normal if this is the new normal and not actually a bug. It genuinely makes using this app too unbearable to use daily. 

Just tap the tabs button and it is immediately noticeable. 

### Expected behavior

Newest tabs should be on the bottom right of the screen.

### Screenshots

_No response_

uazo:
## Description

I finally solved it by leaving the default order but aligning the tabs to the bottom, as shown in the picture
image

Fix for #2456

## All submissions

### Format

### DIFF
--- a/build/patches/Move-navigation-bar-to-bottom.patch
+++ b/build/patches/Move-navigation-bar-to-bottom.patch
@@ -12,81 +12,80 @@ Support for tablet mode is also included.
 Original License: GPL-2.0-or-later - https://spdx.org/licenses/GPL-2.0-or-later.html
 License: GPL-3.0-only - https://spdx.org/licenses/GPL-3.0-only.html
 ---
- cc/base/features.cc                           |  4 ++
+ cc/base/features.cc                           |  4 +
  cc/base/features.h                            |  1 +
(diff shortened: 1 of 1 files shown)


emvaized:
Thanks a lot! I think it's perfect solution

Universalizer:
**Really Thanks for permanent solution.**

csagan5:
This should be a fix for #2456

PR #2541, opened 2023-01-09
Partitioning all cookies by top frame domain

uazo:
## Description

let me start by saying that if I could, I would congratulate the developer of chromium because the changes he has made are among the most beautiful I have seen. the code is straightforward and well done, so much so that very few changes were necessary.

so, this patch activates CHIPS in bromite for all cookies, not just those decided by the websites, leaving the possibility for the user via the interface to decide whether to accept third-party cookies or not ('standard' chips does not allow this option).
by the way, but i think it is a bug (i opened a [report](https://bugs.chromium.org/p/chromium/issues/detail?id=1405772)), in chromium it is not even possible to delete partitioned cookies, but it is true that with partitioned cookies cross-site tracking is not possible, so I do not know whether this is really intended

Thus, with the flag active, first-party cookies (all types, not just those samesite=none) are no longer sent in third-party contexts, because they are partitioned by top frame domain.

I expect some malfunctions, especially with sso, so I'll test it in my version for now, but i'd like it to be thoroughly checked by others too.

## All submissions

### Format

### DIFF
--- a/build/patches/Partitioning-all-cookies-by-top-frame-domain.patch
+++ b/build/patches/Partitioning-all-cookies-by-top-frame-domain.patch
@@ -1,0 +1,266 @@
+From: uazo <uazo@users.noreply.github.com>
+Date: Mon, 9 Jan 2023 12:02:05 +0000
+Subject: Partitioning all cookies by top frame domain
+
+Enables cookie partitioning by top frame etld, respecting the
+user's possible wish to disable all third-party cookies.
+Disabling the flag via the ui restores the normal mode, where 
+samesite=none first-party cookies are sent in third-party contexts.
+---
+ components/browsing_data/core/features.cc        |  2 +-
+ .../browsing_data/same_site_data_remover_impl.cc |  3 ++-
+ net/base/features.cc                             | 16 ++++++++--------
+ net/cookies/canonical_cookie.cc                  |  4 +---
+ net/cookies/cookie_deletion_info.cc              |  3 ++-
+ net/cookies/parsed_cookie.h                      |  7 ++++++-
+ .../sqlite/sqlite_persistent_cookie_store.cc     | 10 ++++++++++
+ services/network/cookie_settings.cc              | 12 ++++++++++--
+ services/network/restricted_cookie_manager.cc    |  3 +++
+ .../modules/cookie_store/cookie_init.idl         |  2 +-
+ .../modules/cookie_store/cookie_store.cc         | 12 ++++++++++++
+ .../cookie_store/cookie_store_delete_options.idl |  2 +-
+ 12 files changed, 57 insertions(+), 19 deletions(-)
+
+diff --git a/components/browsing_data/core/features.cc b/components/browsing_data/core/features.cc
+--- a/components/browsing_data/core/features.cc
++++ b/components/browsing_data/core/features.cc
+@@ -11,7 +11,7 @@ namespace features {
+ 
+ BASE_FEATURE(kEnableRemovingAllThirdPartyCookies,
+              "EnableRemovingAllThirdPartyCookies",
+-             base::FEATURE_DISABLED_BY_DEFAULT);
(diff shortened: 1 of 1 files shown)


uazo:
the developer has fixed the [bug](https://bugs.chromium.org/p/chromium/issues/detail?id=1405772), but in one of their comments I read this:
```
However, you're fixing a bug which was only discovered in production because of missing test
coverage and a lack of manual testing of Clear Browsing Data. So if we don't add test coverage
for the "Important sites" scenario, did we actually test it manually?

Especially since I suspect we want a different behavior for "Important sites" (deletion blocklist) 
than the Clear-Site-Data header (deletion allowlist). Because Clear-Site-Data is triggered by 
the web, and we don't want the deletion to cross storage isolation boundaries; whereas 
"Important sites" is a browser UI feature, so we may want a different behavior. 
E.g. protecting a.com may mean protecting a.com with all of its subframes, or a.com in all 
storage contexts - I'm not sure if we ever discussed what the behavior should be here.
```
so [my change](https://github.com/bromite/bromite/pull/2541/commits/3e2b1d2f680d99661acc02d65b6d98bf0b417bf4#diff-a91fd552748514f6bf8cd567c6ce5f2d686042ecdbe2cdaf6a289edf951e548aR116) is correct that it remains, so that cookies from sites deemed "important" are also deleted.
[here](https://bugs.chromium.org/p/chromium/issues/detail?id=1406581) the related bug to be kept under review

build/patches/Partitioning-all-cookies-by-top-frame-domain.patch
```diff
@@ -0,0 +1,266 @@
...
+diff --git a/content/browser/browsing_data/same_site_data_remover_impl.cc b/content/browser/browsing_data/same_site_data_remover_impl.cc
+--- a/content/browser/browsing_data/same_site_data_remover_impl.cc
++++ b/content/browser/browsing_data/same_site_data_remover_impl.cc
+@@ -49,7 +49,8 @@ void OnGetAllCookiesWithAccessSemantics(
+     // that site's First-Party Set). Since partitioned cookies cannot be used as
+     // a cross-site tracking mechanism, we exclude them from this type of
+     // clearing.
+-    if (!cookie.IsPartitioned() &&
++    // in Bromite, we clear all cookies
++    if (((true)) && !cookie.IsPartitioned() &&
```
csagan5, line 45: This `true` here does nothing? What is the purpose?

csagan5:
> if I could, I would congratulate the developer of chromium because the changes he has made are among the most beautiful I have seen. the code is straightforward and well done, so much so that very few changes were necessary.

I think you can reach out to them via email.

> so [my change](https://github.com/bromite/bromite/pull/2541/commits/3e2b1d2f680d99661acc02d65b6d98bf0b417bf4#diff-a91fd552748514f6bf8cd567c6ce5f2d686042ecdbe2cdaf6a289edf951e548aR116) is correct that it remains, so that cookies from sites deemed "important" are also deleted.
[here](https://bugs.chromium.org/p/chromium/issues/detail?id=1406581) the related bug to be kept under review

Congrats for finding the bug! They have made a fix upstream to have it apply to all partitions?

build/patches/Partitioning-all-cookies-by-top-frame-domain.patch
uazo: none :( is an bug.
that code is not even used in Chromium, I only included it for future reference.
is to be removed, it is not necessary

uazo:
> They have made a fix upstream to have it apply to all partitions?

yes, it seems so.

csagan5:
Can you please update the patch since there is now an upstream improvement?

uazo:
> Can you please update the patch since there is now an upstream improvement?

no need, the changes made by the chromium team have no impact on my code.

> I think you can reach out to them via email.

Maybe next time :) maybe it's not fair to receive congratulations from those who found a bug in production due to a lack of testing..

PR #2545, opened 2023-01-11
Increase maximum adblock filter file size to 20MB

ac130kz:
## Description

I've made my own generated filter file, which doesn't fit into 10 MB. The author of [filtrite](https://github.com/xarantolus/filtrite#using-your-own-filter-lists) also mentions this limitation. Hopefully, 20 MB is more than enough for everyone (tm).

## All submissions

* [x] I have tested that the new change works as intended (AVD or physical device will do)

### Format

### DIFF
--- a/build/patches/Bromite-subresource-adblocker.patch
+++ b/build/patches/Bromite-subresource-adblocker.patch
@@ -1157,8 +1157,8 @@ new file mode 100644
 +
 +namespace adblock_updater {
 +
-+// maximum 10MB for the filters file
-+const int kMaxBodySize = 1024 * 1024 * 10;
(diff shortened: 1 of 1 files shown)


Poopooracoocoo:
> But there will be an integration of Eyeo AdBlock Plus later...

Unless you know when it'll be merged, I think this is a good idea.

csagan5:
> But there will be an integration of Eyeo AdBlock Plus later...

No, there is no such plan. That is an experiment and it is not guaranteed that it will be merged.

csagan5:
Thanks for the PR; it is a download file limit so it is fine to increase it.

PR #2548, opened 2023-01-12
Remove redundant patch hunk

chirayudesai:
* One patch removes it, the other adds it

## Description

*Please explain here what feature or bugfix these changes are addressing and why they should be included*

## All submissions

### Format


uazo:
pure curiosity: how did you notice this? do you have something automatic to propose? i am always looking for improvements in that area.

chirayudesai:
I wish I had a better answer, but in this our fork only has 1 of these patches so the other one conflicted, and I tried to see why that was.

csagan5:
Perhaps the code should be kept and not removed by the binary blobs patch?

uazo:
> Perhaps the code should be kept and not removed by the binary blobs patch?

correct.
That code alerts the ui that the download phase of the language files is finished. In bromite, the download is not done, since bromite already has all the files in the apk, but, by not communicating anything to the ui, it hangs and does not allow language selection.
I remember putting it back in Override-Navigator-Language.patch, but I honestly don't remember why I didn't take it out of Remove-binary-blob-integrations.patch

PR #2561, opened 2023-01-27
Patches for v109

csagan5:
## Description

Patches for v109.

## All submissions

### Format

* [ ] no unnecessary whitespace or unrelated changes


uazo:
there is a crash in the always incognito patch, if you close a group of tabs.
I fixed it with this https://github.com/uazo/bromite/blob/next-v109/build/patches/00109-Add-an-always-incognito-mode-fixup.patch
tell me whether to pass you the whole patch with a pull

uazo:
in addition, a new flag [OPAQUE_ORIGIN_FOR_INCOMING_INTENTS](https://source.chromium.org/chromium/chromium/src/+/main:chrome/android/java/src/org/chromium/chrome/browser/IntentHandler.java;l=1617) was added in v109, which is active by default but which, as we developed the "Enable share intent" and "force open external links in incognito flag" in "Add custom tab intents privacy option" patches, we do not take into account.

https://github.com/uazo/bromite/blob/next-v109/build/patches/00fix-InitiatorOrigin--add-fixup-.patch the fix, but the code must be separated in the two patches respectively

uazo:
in 109 I found

- activation of [MathMLCore](https://chromestatus.com/feature/5240822173794304)
which suffers from the same problem of fonts in canvas, see https://w3c.github.io/mathml-core/#privacy-considerations

- DIPS (Detect Incidental Party State) feature, actually in development (bugid not public!)
Incidental parties are sites without meaningful user interactions, such as bounce trackers: function appears to remove data if it detects a bounce tracker.

- SpeculationRulesPrefetchProxy and SpeculationRulesPrefetchWithSubresources, ~actually enabled by default~
this needs to be checked, it does not seem to me to be related to 'Allow preloading' in the privacy settings, but rather to a flag on the blink side which is always active.
EDIT: are not active, see https://bugs.chromium.org/p/chromium/issues/detail?id=1382315#c9
Preload are disabled with [IsSomePreloadingEnabledIgnoringFinch](https://source.chromium.org/chromium/chromium/src/+/main:chrome/browser/prefetch/prefetch_prefs.cc;drc=69d70eda471b06c74a6a05f41f60a632c96b508b;bpv=1;bpt=0;l=74?q=kPreloadingHoldback)

- kSupportPartitionedBlobUrl
disabled due to some bug, https://crbug.com/1407944 not public!

- kOutOfProcessSystemDnsResolution actually disabled
Servicify or sandbox getaddrinfo on Linux/ChromeOS: Enables out-of-process system DNS resolution so getaddrinfo() never runs in the network service sandbox. System DNS resolution will instead be brokered out over Mojo, likely to run in the browser process.
https://bugs.chromium.org/p/chromium/issues/detail?id=1312224#c5
(android) https://bugs.chromium.org/p/chromium/issues/detail?id=1320192 https://docs.google.com/document/d/18cVwhfOHVO2RFzBUpG9xmjmkci5J_rF0FagWe71uKBU/edit 

there are also others that I think should be verified, see https://github.com/uazo/bromite-flags-list/blob/main/flags-to-check#L520

csagan5:
@uazo I will pick up from this and release it; then I will check whether to skip some versions or not.

uazo:
I am happy to hear from you again. 

> I will check whether to skip some versions or not.

if you want to talk about it, I would be happy to

Abyssgrowth:
Hello there, may i ask about the status / future of bromite, it seems so the project is dead now.
Are there people releasing forks of bromite for the latest version with all security patches? Where can we find anything?

Spark4000:
https://github.com/uazo/bromite-buildtools

Abyssgrowth:
> https://github.com/uazo/bromite-buildtools

Oh it was right there, thank you!!

